PR #1, opened 2022-04-01
Initial adapted copy-paste from ECM [BW-1175]

cjllanwarne:
Status:
- Service starts up and stops
- Liquibase creates a database and table
- OpenAPI generates a swagger page, service responds with hard-coded response
- No connection yet between API and database tables (still todo)


client-resttemplate/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+	id 'com.jfrog.artifactory' version '4.18.2'
+	id 'org.hidetake.swagger.generator'
+}
+
+repositories { mavenCentral() }
+
+apply from: 'artifactory.gradle'
+apply from: 'swagger.gradle'
+
+sourceCompatibility = JavaLanguageVersion.of(11)
```
cjllanwarne, line 15: Should also be 17?

build.gradle
```diff
@@ -0,0 +1,53 @@
+plugins {
+	id 'idea'
```
pshapiro4broad, line 2: Is use of tabs here is a choice for a specific reason? If so I'd be interested to know the reasoning behind it. In my experience I try to avoid tabs as they have different settings in different systems so their display can't be controlled by the code author.

build.gradle
```diff
@@ -0,0 +1,53 @@
...
+
+allprojects {
+	dependencyLocking {
+		lockAllConfigurations()
+	}
+}
+
+subprojects {
+	group = 'bio.terra'
+	version = '0.76.0-SNAPSHOT'
```
pshapiro4broad, line 25: Since you're using a root `build.gradle`, it wouldn't be hard to edit this value in place, but I would recommend putting it in a separate root `gradle.properties` file and using a variable to refer to it here. Something to think about when you get to adding build and bumper GitHub workflows.

codecov.yml
```diff
@@ -0,0 +1 @@
+comment: false
```
pshapiro4broad, line 1: If you're using sonarqube then you shouldn't need to use codecov too, since sonar's scans can include coverage information.

common/spotbugsExclude.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FindBugsFilter>
+  <Match>
+    <Class name="~.*\.generated.*"/>
```
pshapiro4broad, line 4: Spotbugs is probably redundant with the use of sonarqube, but catalog is still using both too. Anyway catalog is using spotbugs but doesn't need to specify a filter like this; I think the generated code is annotated so it isn't scanned by spotbugs. Does the build fail if you remove this filter?

service/src/main/resources/static/openapi.yml
```diff
@@ -0,0 +1,51 @@
+openapi: 3.0.3
+info:
+  title: Terra Batch Analysis
+  description: Manage batch analyses, workflows, and submissions in userspace
+  version: 0.0.1
+paths:
```
pshapiro4broad, line 6: I'm sure you already know this, but in order to deploy in K8S you need at least `/status` as well.

service/src/main/resources/application.yml
```diff
@@ -0,0 +1,44 @@
+logging:
```
pshapiro4broad, line 1: I kind of like how @rtitle did this in the billing profile manager, where the environment variables are all at the top, so I copied that for catalog. Did you consider doing it that way and reject it? I don't really know the pros/cons of each way so I'd be interested to hear if there's a reason to prefer one way or the other.  https://github.com/DataBiosphere/terra-billing-profile-manager/blob/main/service/src/main/resources/application.yml

service/src/main/java/bio/terra/batchanalysis/controllers/BatchApiController.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Controller
+@Slf4j
+public class BatchApiController implements BatchApi {
+
+  @Override
+  public ResponseEntity<List<String>> listMethods() {
+    // TODO: Get an answer from the database?
+    return ResponseEntity.ok(
+        Collections.singletonList(
```
pshapiro4broad, line 18: With JDK 11+ I tend to prefer the new collection constructors over the old ones in `Collection`. In this case you can write
```suggestion
        List.of(
```

service/spring.gradle
```diff
@@ -0,0 +1,35 @@
...
+
+	// DB deps (still versioned by Spring):
+	implementation 'org.liquibase:liquibase-core'
+	runtimeOnly 'org.postgresql:postgresql'
+}
+
+test { useJUnitPlatform() }
+
+task bootRunDev {
+	bootRun.configure { systemProperty 'spring.profiles.active', 'human-readable-logging' }
```
pshapiro4broad, line 28: In order to enable human readable logs for all versions of running the app, and all tests, catalog uses `tasks.withType()`. With that form you don't need to create a custom task, although maybe you want to have a custom one here for other reasons. See https://github.com/DataBiosphere/terra-data-catalog/blob/main/buildSrc/src/main/groovy/bio.terra.catalog.java-common-conventions.gradle#L13

service/spring.gradle
```diff
@@ -0,0 +1,35 @@
...
+	finalizedBy(bootRun)
+}
+
+springBoot {
+	mainClass = 'bio.terra.batchanalysis.BatchAnalysisWebApplication'
```
pshapiro4broad, line 33: Unless your .jar has multiple classes with `public static void main(String[] args)` declared, this isn't necessary as Spring will search and find the main class on startup.

service/src/main/java/bio/terra/batchanalysis/controllers/BatchApiController.java
cjllanwarne: Funnily enough, IntelliJ corrected my original use of `List.of` to this `singletonList` construct, on account of having only one item

service/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# Set the base/parent image
+# see https://github.com/broadinstitute/dsp-appsec-blessed-images/tree/main/jre
+# Google's distroless images are openjdk, this is the simplest with adoptopenjdk
+FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-debian
```
pshapiro4broad, line 4: I'm wondering about this comment. Is it true that we shouldn't be using distroless since it uses openjdk? I don't think that's true for the JDK 17 images at least. Catalog and BPM are using distroless `us.gcr.io/broad-dsp-gcr-public/base/jre:17-distroless`

service/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+# download and untar the cloud profiler
+RUN mkdir -p /opt/cprof && \
+  curl -q O https://storage.googleapis.com/cloud-profiler/java/latest/profiler_java_agent.tar.gz \
```
pshapiro4broad, line 8: If all our java services include this agent, I wonder if we could have appsec maintain a custom base image that includes this code.

service/analysis.gradle
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	dependsOn test
+
+	// https://www.baeldung.com/jacoco-report-exclude#2-gradle-configuration
+	afterEvaluate {
+		classDirectories.setFrom(files(classDirectories.files.collect {
+			fileTree(dir: it, exclude: ['bio/terra/batchanalysis/generated/**'])
+		}))
+	}
```
pshapiro4broad, line 14: I wonder if this is necessary; the catalog and BPM project's don't have these lines. (Although I'm not sure if BPM's coverage reports are configured yet)

service/src/main/java/bio/terra/batchanalysis/controllers/BatchApiController.java
pshapiro4broad: Maybe the IDE's language level is incorrect? I have the opposite behavior in my setup. The intention I see is `Java > Collections > Replace with `Arrays.asList()`. (The name is misleading as it suggests the newer `List.of()` form if you have the IDE set to JDK 9 or later.)

cjllanwarne:
Closing for now. Reconsidering the copy/paste origin

PR #2, opened 2022-04-08
Initial copy/paste of data catalog (repo bootstrap attempt 2) [BW-1175]

cjllanwarne:
An alternative attempt (vs #1) to bootstrap the (now) CBAS service, this time using the slightly newer [data-catalog](https://github.com/DataBiosphere/terra-data-catalog) repo as the template.


gradlew
```diff
@@ -0,0 +1,234 @@
+#!/bin/sh
```
cjllanwarne, line 1: Note to reviewer: this is a standard copy/pasted gradle wrapper. Don't review!

gradlew.bat
```diff
@@ -0,0 +1,89 @@
+@rem
```
cjllanwarne, line 1: Note to reviewer: this is a standard copy/pasted gradle wrapper. Don't review!

service/src/main/resources/static/swagger-ui.html
```diff
@@ -0,0 +1,37 @@
+<!-- HTML for static distribution bundle build -->
```
cjllanwarne, line 1: Standard copy/pasted swagger ui page. Don't review.

client/.swagger-codegen-ignore
```diff
@@ -0,0 +1,2 @@
+**
```
cjllanwarne, line 1: Think about `// TODO! Maybe get started by copy/pasting from data-catalog?` until you know what this does

docs/getting-started.md
```diff
@@ -0,0 +1,3 @@
+# Getting Started
```
cjllanwarne, line 1: Might duplicate `DEVELOPMENT.md`?

service/sonar-project.properties
```diff
@@ -0,0 +1,3 @@
+# Inheritance tree of classes should not be too deep
```
cjllanwarne, line 1: Reminder to find out and tell people what sonar is

service/src/main/resources/static/openapi.yml
```diff
@@ -0,0 +1,51 @@
+openapi: 3.0.3
```
cjllanwarne, line 1: Should be a softlink to `common/openapi.yml`

client/.swagger-codegen-ignore
cjllanwarne: Update: tells the codegen not to override anything in these directories. Seems like a good idea to leave it.

docs/getting-started.md
cjllanwarne: Yup, this was the data-catalog equivalent of the external-credential-manager's `DEVELOPMENT.md`. Deleting this file.

service/sonar-project.properties
cjllanwarne: Update: It's a code quality checker

service/src/main/resources/static/openapi.yml
cjllanwarne: Hopefully updated now 🤞

service/src/main/resources/static/openapi.yml
```diff
@@ -0,0 +1 @@
+../../../../../common/openapi.yml
```
cjllanwarne, line 1: This is github's way of unpacking the softlink, I guess

common/openapi.yml
```diff
@@ -0,0 +1,51 @@
...
+    ServerError:
+      description: Server error
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/ErrorReport'
+
+  schemas:
+
+    ErrorReport:
```
pshapiro4broad, line 44: This (and catalog) follows the ECM's `ErrorReport` model. Other projects like BPM have a `causes` field that contains an array of all the exceptions up the chain. The TCL's global exception handler supports gathering up these exceptions so I've been thinking about changing catalog to use the BPM model instead.

common/postgres-init.sql
```diff
@@ -0,0 +1,3 @@
+CREATE ROLE cbas_user WITH LOGIN ENCRYPTED PASSWORD 'cbas_password';
+CREATE DATABASE cbas_db OWNER cbas_user;
+\c cbas_db;
```
pshapiro4broad, line 3: We this line because it was necessary on some systems to install `pgcrypto` in the correct database. If you're not using `pgcrypto` (and I'm not sure catalog should be, but the project we copied did...) then you shouldn't need the `\c cbas_db;` line.

common/openapi.yml
salonishah11: What is BPM?

service/src/main/resources/application.yml
```diff
@@ -0,0 +1,46 @@
...
+    mime-types: text/css,application/javascript
+  port: 8080
+
+spring:
+  # application name and version are used to populate the logging serviceContext
+  # https://github.com/DataBiosphere/terra-common-lib/blob/480ab3daae282ddff0fef8dc329494a4422e32f1/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java#L118
+  application.name: cbas
+  application.version: ${cbas.version.gitHash:unknown}
+
+  config.import: optional:classpath:rendered/providers.yaml;classpath:rendered/version.properties;optional:classpath:rendered/allowed_jwks.yaml;optional:classpath:rendered/allowed_jwt_issuers.yaml
```
pshapiro4broad, line 17: In catalog we aren't using this (although it looks like there's still some cruft in catalog from the copied config) so I would bet that you don't need it for this project.

service/sonar-project.properties
pshapiro4broad: This comment should probably have linked to the exact sonar rule, which is https://rules.sonarsource.com/java/RSPEC-110

common/postgres-init.sql
pshapiro4broad: As for why you might want `pgcrypto`, catalog uses it to generate UUIDs in SQL. The relevant code is 

https://github.com/DataBiosphere/terra-data-catalog/blob/main/service/src/main/resources/db/changelog.xml#L4
https://github.com/DataBiosphere/terra-data-catalog/blob/main/service/src/main/resources/db/changesets/20220301_datasettable.yaml#L12

aednichols (approved):

service/sonar-project.properties
cjllanwarne: Added this link as a comment in the file

common/postgres-init.sql
cjllanwarne: I suspect we will generate all IDs in the java service, so will remove this for now. Good clue if we need it later though!

service/src/main/resources/application.yml
cjllanwarne: I'm going to follow the "remove it as long as it doesn't break" rule. The less "cruft that we might need one day", the better

PR #3, opened 2022-04-26
CBAS Docker build and publish [BW-1178]

cjllanwarne:
- Adds docker build via jib
- Adds a tag action which updates the project and creates a github tag
- Adds docker publish to dockerhub via github action, triggered whenever a github tag is created

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,7 +10,9 @@ on:
 # For future functionality like 'publish to Artifactory', 'push to GCR', ..., see the data catalog version of this file.
 
 env:
-  SERVICE_NAME: ${{ github.event.repository.name }}
+  # SERVICE_NAME: ${{ github.event.repository.name }}
+  # NB: Unlike terra-data-catalog, this doesn't match the repo name
+  SERVICE_NAME: composite-batch-analysis-service
   DOCKERHUB_ORG: broadinstitute
 
 jobs:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,14 +2,13 @@ name: Publish and deploy
 on:
   create:
   workflow_dispatch:
+    inputs:
+      tagOverride:
+        description: 'Specify the tag to use for this publish'
+        required: true
   push:
     branches:
       - cjl_docker_publish
-
-inputs:
-  tagOverride:
-    description: 'Specify or override the tag used for this publish'
-    required: false
 
 # Note: Adapted from terra-data-catalog.
 # For future functionality like 'publish to Artifactory', 'push to GCR', ..., see the data catalog version of this file.
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,5 +1,10 @@
 name: Publish and deploy
-on: create
+on:
+  create:
+  workflow_dispatch:
+  push:
+    branches:
+      - cjl_docker_publish
 
 # Note: Adapted from terra-data-catalog.
 # For future functionality like 'publish to Artifactory', 'push to GCR', ..., see the data catalog version of this file.
@@ -10,7 +15,7 @@ env:
 
 jobs:
   publish-job:
-    if: startsWith(github.ref, 'refs/tags/')
+#    if: startsWith(github.ref, 'refs/tags/')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -25,7 +30,10 @@ jobs:
         description: |
           Sets up a step output called TAG with the appropriate tag
         id: tag
-        run: echo ::set-output name=TAG::$(git describe --tags)
+        run:
+          echo ::set-output name=TAG::cjl_test_tag
+          # TODO: Revert back to this after testing:
+          # echo ::set-output name=TAG::$(git describe --tags)
 
       - name: Construct docker image name and tag
         id: image-name
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,9 +10,6 @@ on:
       tagOverride:
         description: 'Specify the tag to use for this publish'
         required: true
-  push:
-    branches:
-      - cjl_docker_publish
 
 env:
   # SERVICE_NAME: ${{ github.event.repository.name }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,8 +27,7 @@ jobs:
           cache: 'gradle'
 
       - name: Parse tag
-        description: |
-          Sets up a step output called TAG with the appropriate tag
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/cbas/app/App.java
```diff
@@ -0,0 +1,35 @@
...
+      "bio.terra.common.retry.transaction",
+      // Scan for tracing-related components & configs
+      "bio.terra.common.tracing",
+      // Scan all service-specific packages beneath the current package
+      "bio.terra.cbas"
+    })
+@ConfigurationPropertiesScan("bio.terra.cbas")
+@EnableRetry
+@EnableTransactionManagement
+public class App {
```
cjllanwarne, line 31: This `App` class is here to let the jib publish to have a `main` class to point to.

service/publishing.gradle
```diff
@@ -1 +1,45 @@
...
+}
+task downloadProfilerAgent(type: Download) {
+    // where to download the Cloud Profiler agent https://cloud.google.com/profiler/docs/profiling-java
+    src "https://storage.googleapis.com/cloud-profiler/java/latest/profiler_java_agent.tar.gz"
+    dest "${buildDir}/cprof_java_agent_gce.tar.gz"
+}
+task extractProfilerAgent(dependsOn: downloadProfilerAgent, type: Copy) {
+    from tarTree(downloadProfilerAgent.dest)
+    into "${jibExtraDirectory}/${cloudProfilerLocation}"
+}
```
pshapiro4broad, line 19: The other approach here is to use a `Dockerfile` to install the profiler agent. I copied this approach from another project, and it works, but I've thought that the dockerfile version might be a bit cleaner. The main difference is that the agent archive is downloaded and installed by docker, not by gradle. An example is https://github.com/DataBiosphere/terra-external-credentials-manager/blob/dev/service/Dockerfile

service/publishing.gradle
```diff
@@ -1 +1,45 @@
...
+    }
+    extraDirectories {
+        paths = [file(jibExtraDirectory)]
+    }
+    container {
+        filesModificationTime = ZonedDateTime.now().toString() // to prevent ui caching
+        mainClass = 'bio.terra.cbas.app.App'
+        jvmFlags = [
+            "-agentpath:" + cloudProfilerLocation + "/profiler_java_agent.so=" +
+                "-cprof_service=bio.terra.catalog" +
```
pshapiro4broad, line 34:
```suggestion
                "-cprof_service=bio.terra.cbas" +
```

service/src/main/java/bio/terra/cbas/app/App.java
pshapiro4broad: It's also used by spring boot when launching your app via `gradle bootRun`

.github/workflows/tag.yml
```diff
@@ -0,0 +1,22 @@
...
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.BROADBOT_TOKEN }} # this allows the push to succeed later
+      - name: Bump the tag to a new version
+        # https://github.com/DataBiosphere/github-actions/tree/master/actions/bumper
+        uses: databiosphere/github-actions/actions/bumper@bumper-0.0.4
+        id: tag
+        env:
+          GITHUB_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
+          HOTFIX_BRANCHES: hotfix.*
```
aednichols, line 18: Are we still doing `hotfix` branches in the new continuous deployment world?

I would love to favor continuous + seamless deployment early on, and make accommodations for slower releases later if necessary. Because it's really hard to go in the other direction (making things more continuous).

aednichols (approved):

.github/workflows/tag.yml
pshapiro4broad: The terra service hotfix process is documented [here](https://github.com/broadinstitute/terra-helmfile/blob/master/docs/HOTFIX.md). I think the hotfix process is necessary until we break apart the monolith release process.

cjllanwarne: Great question. This looks optional in the action so I'll happily remove it.

cjllanwarne: @pshapiro4broad because Cromwell is an app I think our deployments will necessarily be separate from the monolith from the start anyway

aednichols: @pshapiro4broad you are right about the monolith of course, but as Chris says this service is intended as "never was in the monolith" versus "needs to break out of the monolith".

pshapiro4broad: Oh, I didn't know that about cbas. That's good!

evanbernstein: We should make sure to consult with Ambica & QA about these things. We still need to run through the new service checklist and if there are differences due to the app model, then we should document them there rather than us accidentally going rogue.

aednichols: The recommendation is to review [the checklist](https://docs.google.com/document/d/1deV-i3_oMGBwUreDUKEhovdLawBdf0feshcP1h11wR0/edit#heading=h.sogi2y5vnhy) at the start of development, fill in parts as possible during development (now), and have a review meeting before designating the service as in-production (long way off).

service/publishing.gradle
cjllanwarne: At first glance, I like the `jib` option more, but it feels like it would be relatively easy to switch between them if some compelling reason came up

cjllanwarne:
Now tested and validated using 'push to branch' testing. See action history and the [dockerhub repo](https://hub.docker.com/repository/docker/broadinstitute/composite-batch-analysis-service) for the results.

Final validation will occur when the PR merges and (hopefully) a new tag and a new release will be pushed automatically.

aednichols (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,63 @@
...
+          -Djib.console=plain
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: dsdejenkins
+          password: ${{ secrets.DSDEJENKINS_DOCKERHUB_PASSWORD }}
+
+      - name: Push dockerhub image
+        run: docker push ${{ steps.image-name.outputs.DOCKERHUB_NAME }}
```
salonishah11, line 62: I had recently learnt ([Slack conversation](https://broadinstitute.slack.com/archives/C0DSD41QT/p1651166000447249)) that GCR is the standard place to host images. Should we also consider pushing CBAS image to GCR?

cjllanwarne: Yes I think that's certainly going to be the long term plan. For now (and since we'll also want to use this in Azure), docker seemed like a good first step.

salonishah11 (approved):

PR #5, opened 2022-05-12
BW-1177 Add Swagger path, update link

aednichols:
There is no redirect yet and hitting the bare path shows an error.

Apologies for the trailing whitespace spam. IntelliJ absolutely insists on doing it, even when I disable in Preferences > Editor > General > On Save

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,4 +1,4 @@
-# Developing BatchAnalysis 
+# Developing BatchAnalysis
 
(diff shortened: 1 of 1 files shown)


cjllanwarne:
Not to worry:
image

Interesting that the redirect doesn't work, I was pretty sure that _used_ to work.... 🤔 . Maybe in a prior iteration...

cjllanwarne (approved):

salonishah11 (approved):

PR #6, opened 2022-05-19
BW-1180 Skeleton integration test

aednichols:
Run using
```
gw --build-cache runTest --args="suites/FullIntegration.json build/reports"
```
which is very similar to [the command used in Data Catalog's integration tests](https://github.com/DataBiosphere/terra-data-catalog/blob/main/.github/workflows/build-and-test.yml#L175-L176):
```
./gradlew runTest --args="suites/FullIntegration.json build/reports"
```

If `gw` is not available for you, install `gng` as described [here](https://github.com/DataBiosphere/terra-batch-analysis/blob/main/DEVELOPMENT.md#tips).

To make sure it's really doing something:
- Edit `integration/src/main/java/scripts/testscripts/GetStatus.java` and change `true` to `false`. The test fails.
- Run the test and search for `>>> Got here` in the console. It is sandwiched in the middle of a ton of other output.


integration/build.gradle
```diff
@@ -1 +1,36 @@
...
+    implementation 'org.slf4j:slf4j-api'
+    implementation 'org.glassfish.jersey.inject:jersey-hk2'
+    implementation 'org.junit.jupiter:junit-jupiter-api'
+    implementation 'org.hamcrest:hamcrest'
+
+    // Google Dependencies
+    implementation 'com.google.auth:google-auth-library-oauth2-http:1.4.0'
+
+    // Terra Test Runner Library
+    implementation 'bio.terra:terra-test-runner:0.1.4-SNAPSHOT'
```
pshapiro4broad, line 35: You should also update the dependency for this in https://github.com/DataBiosphere/terra-batch-analysis/blob/main/buildSrc/build.gradle#L22 so it uses the same version

integration/build.gradle
```diff
@@ -1 +1,36 @@
...
+
+dependencyManagement {
+    imports {
+        mavenBom(SpringBootPlugin.BOM_COORDINATES)
+    }
+}
+
+// In `terra-data-catalog` the only `repositories { }` stanzas are in
+// the top-level `buildSrc` directory. But for some reason it does not get
+// inherited (or whatever the mechanism is) in this project.
```
pshapiro4broad, line 16: This is because this subproject doesn't have a dependency on any of your plugin projects (which is what `buildSrc` is defining). Change `plugins` to be
```
plugins {
    id 'java-application-conventions'
    id 'io.spring.dependency-management'
    id 'bio.terra.test-runner-plugin'
 }
```
and you can delete this `repositories` block.

integration/build.gradle
pshapiro4broad: There is most likely a way to avoid the duplicate dependency but I don't know how to do it. This is case where plugin dependencies and library dependencies use different configurations, which is usually what you want.

integration/build.gradle
aednichols: Thanks... I knew I was off on something... not sure why I originally deleted `java-application-conventions`

salonishah11 (approved):
LGTM 👍

integration/src/main/resources/rendered/testrunner-perf.json
```diff
@@ -0,0 +1,4 @@
+{
+  "placeholder": "Not actually checking in a service account key",
+  "next steps": "Figure out if we need an SA; if so, create and securely store the key"
```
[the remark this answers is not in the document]
aednichols: The test framework exits with an error if this JSON file does not exist. In the minimal spirit of the ticket, I chose to implement the simplest "solution".

integration/src/main/java/scripts/testscripts/GetStatus.java
```diff
@@ -0,0 +1,14 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import bio.terra.testrunner.runner.TestScript;
+import bio.terra.testrunner.runner.config.TestUserSpecification;
+
+public class GetStatus extends TestScript {
+  @Override
+  public void userJourney(TestUserSpecification testUser) throws Exception {
+    System.out.println(">>> Got here");
```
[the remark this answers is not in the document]
aednichols: It is definitely removable, I wanted reviewers (and myself) to have some confidence the test is really running the code intended.

integration/src/main/resources/servers/cbas-local.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "",
+  "description": "",
+
+  "catalogUri": "http://localhost:8080",
```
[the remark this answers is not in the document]
aednichols: Yes, removed.

PR #14, opened 2022-05-24
BW-1244: Implement tests for getStatus

cjllanwarne:


integration/src/main/java/scripts/client/BatchClient.java
```diff
@@ -0,0 +1,21 @@
...
+
+  /**
+   * Build the no-auth API client object for the CBAS server. No access token is needed for this API
+   * client.
+   *
+   * @param server the server specification. Currently ignored and tests actually against localhost
+   *     (in github, a test service created in a previous action step).
+   */
+  public BatchClient(ServerSpecification server) throws IOException {
+    // TODO: Let this float? Set this up automatically? (Right now the github action which triggers
```
pshapiro4broad, line 17: You probably already know this, but the usual way to do this is to specify the path to your service is using a field in `ServerSpecification`. The catalog code for this is here: https://github.com/DataBiosphere/terra-test-runner/blob/main/src/main/java/bio/terra/testrunner/runner/config/ServerSpecification.java#L56

When I was first setting up catalog tests, I used another field in this object to store the catalog URI while I waited for my PR into `terra-testrunner` to merge up. The fields are only used by your test so you can use a field for any purpose you like, regardless of its name.

aednichols (approved):

salonishah11 (approved):

integration/src/main/java/scripts/client/BatchClient.java
cjllanwarne: Yeah... to be honest Cromwell/CBAS feels like a slightly awkward fit for these integration tests, since they seem to be "set up a control plane environment to test everything together" style, and our deployment medium is a k8s cluster inside an app... so the concept of "a Cromwell" doesn't really make sense in `ServerSpecification`... but if it allows us have different URLs in different config files... 🤷

pshapiro4broad: Maybe the test runner scripts aren't a great match for your use case. I think the main benefit outside your team is for QA to have a track of test runs, once you enable uploading test results to the QA result bucket (or big query table?). I also kind of like having to think about testing from scratch (vs a usual java integration tests) but for some use cases it's a lot of work to write the tests this way.

PR #15, opened 2022-05-25
BW-1246: Add sonarqube and srcclr

cjllanwarne:


aednichols (approved):

aednichols:
> 0 Bugs

Bold claim haha

pshapiro4broad:
> > 0 Bugs
> 
> Bold claim haha

One thing that wasn't immediately obvious from using sonar is that the report is only on lines in the current PR. So your code base could have 100s of "bugs", but if there's no bugs in the code in the current PR, the report will show `0`.

salonishah11 (approved):

PR #18, opened 2022-06-10
BW-1250 Basic status check against Cromwell URL

aednichols:
CBAS `/status` response when Cromwell is running:
```
{
  "ok": true,
  "systems": {
    "Cromwell": {
      "ok": true,
      "messages": [
        "{}"
      ]
    }
  }
}
```
When it's stopped:
```
{
  "ok": true,
  "systems": {
    "Cromwell": {
      "ok": false,
      "messages": [
        "I/O error on GET request for \"http://localhost:8000/engine/v1/status\": Connection refused; nested exception is java.net.ConnectException: Connection refused"
      ]
    }
  }
}
```

As per AC the CBAS global status stays `"ok": true` regardless of what Cromwell is doing.

What explicitly does not work yet (future tickets):
- Running anywhere other than locally, with a local Cromwell
- Using a generated API client instead of a generic HTTP library
  - As demonstrated by Terra Data Catalog calling Rawls [here](https://github.com/DataBiosphere/terra-data-catalog/blob/main/service/src/main/java/bio/terra/catalog/rawls/RawlsService.java#L43).

### DIFF
--- a/service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
@@ -4,11 +4,12 @@ import bio.terra.cbas.api.PublicApi;
 import bio.terra.cbas.config.CromwellServerConfiguration;
 import bio.terra.cbas.model.SystemStatus;
 import bio.terra.cbas.model.SystemStatusSystems;
+import java.net.URL;
+import java.net.URLConnection;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
-import org.springframework.web.client.RestTemplate;
 
 @Controller
 public class PublicApiController implements PublicApi {
@@ -23,11 +24,15 @@ public class PublicApiController implements PublicApi {
   @Override
   public ResponseEntity<SystemStatus> getStatus() {
 
-    RestTemplate restTemplate = new RestTemplate();
     String result;
-    Boolean isOk;
+    boolean isOk;
     try {
-      result = restTemplate.getForObject(this.cromwellConfig.healthUri(), String.class);
+      URL url = new URL(this.cromwellConfig.healthUri());
+      URLConnection connection = url.openConnection();
+      connection.setConnectTimeout(5000);
+      var stream = connection.getInputStream();
+
+      result = new String(stream.readAllBytes());
       isOk = true;
     } catch (Exception e) {
       result = e.getLocalizedMessage();
--- a/service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
@@ -1,8 +1,11 @@
 package bio.terra.cbas.controllers;
 
 import bio.terra.cbas.api.PublicApi;
+import bio.terra.cbas.config.CromwellServerConfiguration;
 import bio.terra.cbas.model.SystemStatus;
 import bio.terra.cbas.model.SystemStatusSystems;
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
```diff
@@ -1,16 +1,43 @@
...
+    this.cromwellConfig = cromwellConfig;
+  }
+
   @Override
   public ResponseEntity<SystemStatus> getStatus() {
-    return new ResponseEntity<>(new SystemStatus().ok(true), HttpStatus.OK);
+
+    RestTemplate restTemplate = new RestTemplate();
+    String result;
+    Boolean isOk;
```
pshapiro4broad, line 28: Two things; if you have a local var that will only be assigned once, you can make it clear to the reader (and compiler) by using `final`. Second, I would avoid using boxed types (like `Boolean`) unless you need to store `null` at some point. IMO this also makes the code's intent clearer.
```suggestion
    final String result;
    final boolean isOk;
```

service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
```diff
@@ -1,16 +1,43 @@
...
+    try {
+      result = restTemplate.getForObject(this.cromwellConfig.healthUri(), String.class);
+      isOk = true;
+    } catch (Exception e) {
```
pshapiro4broad, line 32: I've never used `RestTemplate` and this looks OK but there is a JDK builtin for reading from URLs, and I usually prefer JDK APIs when possible.
```suggestion
    try (var stream = new URL(cromwellConfig.healthUri()).openStream()) {
      result = new String(stream.readAllBytes());
      isOk = true;
    } catch (Exception e) {
```

service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
aednichols: I am in support of "assign exactly once" style types, thanks for the suggestion.

I think the compiler might not be smart enough to understand that only one branch of the `try` will execute, getting `variable isOk might already have been assigned` in the `catch`.

cjllanwarne (approved):
Very cool!

service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
aednichols: Thanks, I did not implement this exact suggestion, but I switched to a different JDK built-in that gave me the hook to add a timeout (from debugging various Cromwell IO things I have a paranoia of infinite timeouts)

salonishah11 (approved):

service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
```diff
@@ -1,16 +1,49 @@
...
+    String result;
+    boolean isOk;
+    try {
+      URL url = new URL(this.cromwellConfig.healthUri());
+      URLConnection connection = url.openConnection();
+      connection.setConnectTimeout(5000);
+      var stream = connection.getInputStream();
+
+      result = new String(stream.readAllBytes(), StandardCharsets.UTF_8);
+      isOk = true;
```
pshapiro4broad, line 37: In my code I don't use an explicit timeout but if you do this, maybe you want to set read timeout too? Both have infinite timeout by default.

Anyway there's a minor bug with this version; the opened stream is never closed (although likely it will eventually be closed when GC'd). Using try-with-resources is how I'd make sure the stream is closed:

```
            try (var stream = connection.getInputStream()) {
                result = new String(stream.readAllBytes());
                isOk = true;
            }
```

aednichols: Ooh, thanks! Fixing now.

PR #19, opened 2022-06-13
BW-1288 Auto close stream

aednichols:
@pshapiro4broad I think this should do the trick; though I'm admittedly not sure the nested `try` is a nice pattern.

If not I will just adopt your suggestion literally and forget about the timeout stuff for now, so I can move on to something more important.

Thank you for your reviews, they are super helpful!

### DIFF
--- a/service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
@@ -31,10 +31,13 @@ public class PublicApiController implements PublicApi {
       URL url = new URL(this.cromwellConfig.healthUri());
       URLConnection connection = url.openConnection();
       connection.setConnectTimeout(5000);
-      var stream = connection.getInputStream();
-
-      result = new String(stream.readAllBytes(), StandardCharsets.UTF_8);
-      isOk = true;
+      try (var stream = connection.getInputStream()) {
+        result = new String(stream.readAllBytes());
+        isOk = true;
+      } catch (Exception e) {
+        result = e.getLocalizedMessage();
+        isOk = false;
+      }
     } catch (Exception e) {
       result = e.getLocalizedMessage();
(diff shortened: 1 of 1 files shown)


service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
```diff
@@ -31,10 +31,14 @@ public ResponseEntity<SystemStatus> getStatus() {
...
-      result = new String(stream.readAllBytes(), StandardCharsets.UTF_8);
-      isOk = true;
+      connection.setReadTimeout(5000);
+      try (var stream = connection.getInputStream()) {
+        result = new String(stream.readAllBytes());
+        isOk = true;
+      } catch (Exception e) {
+        result = e.getLocalizedMessage();
+        isOk = false;
+      }
```
pshapiro4broad, line 41: You don't need to duplicate the `catch` here, you can have a `try` with no `catch` or `finally`
```suggestion
      }
```

pshapiro4broad (approved):
I have one change to eliminate the nested `catch`. I agree nested `try` is generally bad but when you write it without `catch` and `finally` it's pretty clear to the reader that it's only used to ensure that the stream is closed.

> Thank you for your reviews, they are super helpful!

You're welcome! Java is simpler than Scala (IMO) but there's still a lot of odd corners.

service/src/main/java/bio/terra/cbas/controllers/PublicApiController.java
aednichols: Oh, so the uncaught exception will bubble up to the outer `try`?

pshapiro4broad: Yes, only a matching `catch` clause will stop the exception propagation

cjllanwarne (approved):

PR #20, opened 2022-06-16
BW-1278: Add a WES-like `/runs` POST endpoint

cjllanwarne:
It's 'wes-like' in that the parameters and URL shape is the same. It's not WES-complete because only a handful of parameters are supported. TBD whether we want to go full-WES, but in the absence of reasons not to, I copy/pasted a few of the most relevant sections from the WES spec.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -17,6 +17,27 @@ paths:
           $ref: '#/components/responses/ServerError'
         '503':
           $ref: '#/components/responses/SystemStatusResponse'
+  "/api/batch/v1/runs":
+    post:
+      tags: [ runs ]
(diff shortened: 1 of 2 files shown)


aednichols (approved):

common/openapi.yml
```diff
@@ -17,6 +17,27 @@ paths:
           $ref: '#/components/responses/ServerError'
         '503':
           $ref: '#/components/responses/SystemStatusResponse'
+  "/api/batch/v1/runs":
```
salonishah11, line 20: Is `batch` needed in API path if we are only submitting one workflow and not batch? Or is this just WES convention?

salonishah11 (approved):

PR #23, opened 2022-07-11
BW-1253: Auto-update helm chart upon CBAS release

fredlas:
Upon publishing a new version to this repo, github will now automatically write the tag associated with that new version into the [broadinstitute/cromwhelm repo](https://github.com/broadinstitute/cromwhelm)'s CBAS Helm Chart.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -85,3 +85,21 @@ jobs:
 
       - name: Push GCR image
         run: docker push ${{ steps.image-name.outputs.GCR_NAME }}
+
+      - name: Update CBAS Helm Chart in cromwhelm
+        env:
+          BROADBOT_GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
+        run: |
+          set -e
+          cd cromwhelm
+          git checkout main
+          ls -la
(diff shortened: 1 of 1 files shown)


fredlas:
Not sure if this should be a separate "jobs" entry, as opposed to just a couple of "steps" tacked onto the end.

Also, just for my own understanding: the code I copied from includes the commands `ls -la` and `git diff`. I left them in, thinking they're meant to make the logs more helpful in case any debugging is needed around these actions. Is that correct?

fredlas:
Ok, separated it out (I think). This should serve to rerun the tests. For the future, if I were to want to rerun presubmits without having any changes to make, does an empty commit work for that?

aednichols (commented):
I doubt the failures are related to the content of the PR, you may simply have unfortunate timing with respect to the repo rename. I would check in with Chris who is working on that.
```
Execution failed for task ':service:sonarqube'.
> Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator
```
```
Execution failed for task ':service:srcclrInit'.
> Invalid configuration - apiToken must be set!
```

fredlas:
Moved this over to #25 - turns out the issue with the presubmits was that this PR was coming from a forked repo, rather than a branch in the main repo. Thanks @cjllanwarne !

PR #24, opened 2022-07-12
BW - 1304 get /runs endpoint

kpierre13:


service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -22,6 +27,25 @@ public RunsApiController(CromwellServerConfiguration cromwellConfig) {
...
 
+  @Override
+  public ResponseEntity<RunLogResponse> getRun() {
+    List<LogRunRequest> runs = new ArrayList<>();
+    String runId = UUID.randomUUID().toString();
+    String name = "CBAS";
+    Date now = new Date();
+
+    runs.add(
+        (new LogRunRequest()
```
fredlas, line 38: You can skip this paren pair (unless you are keeping consistent style with how other HTTP request handlers are written in this codebase, or something like that)

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -22,6 +27,25 @@ public RunsApiController(CromwellServerConfiguration cromwellConfig) {
...
+            .runId(runId)
+            .state(RunState.UNKNOWN)
+            .workflowUrl("urlHere")
```
fredlas, line 41: (Just to confirm) is this an intentional placeholder for something that can't be done yet? (Same question for "params" below)

fredlas (approved):
Looks good; all I had were a couple of style suggestions and a question/confirmation.

salonishah11 (approved):

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
kpierre13: Yes, definitely an intentional placeholder. The next ticket will be the actual wiring to Cromwell.

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -22,6 +27,25 @@ public RunsApiController(CromwellServerConfiguration cromwellConfig) {
     this.cromwellConfig = cromwellConfig;
   }
 
+  @Override
+  public ResponseEntity<RunLogResponse> getRun() {
+    List<LogRunRequest> runs = new ArrayList<>();
+    String runId = UUID.randomUUID().toString();
```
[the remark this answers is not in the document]
kpierre13: Do you mind explaining what you mean by, 'directly where they're used'? Do you mean in-line with where they appear later on in the code?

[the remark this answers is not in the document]
kpierre13: Maybe something like ```runs.add(
        new LogRunRequest()
            .runId(UUID.randomUUID().toString())```?

PR #25, opened 2022-07-13
BW-1253: Auto-update helm chart upon CBAS release

fredlas:
Upon publishing a new version to this repo, github will now automatically write the tag associated with that new version into the [broadinstitute/cromwhelm repo](https://github.com/broadinstitute/cromwhelm)'s CBAS Helm Chart.

(Continued from https://github.com/DataBiosphere/cbas/pull/23 , because the presubmits don't work if you try to do a PR from a forked repo)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,70 +18,103 @@ env:
   GOOGLE_PROJECT: broad-dsp-gcr-public
 
 jobs:
-  publish-job:
-    if: startsWith(github.ref, 'refs/tags/') || github.event.inputs.tagOverride != ''
-    runs-on: ubuntu-latest
-    permissions:
-      contents: 'read'
-      id-token: 'write'
+  - publish-job:
+      if: startsWith(github.ref, 'refs/tags/') || github.event.inputs.tagOverride != ''
+      runs-on: ubuntu-latest
+      permissions:
+        contents: 'read'
+        id-token: 'write'
 
-    steps:
-      - uses: actions/checkout@v2
-      - name: Set up JDK
-        uses: actions/setup-java@v2
-        with:
-          java-version: '17'
-          distribution: 'temurin'
-          cache: 'gradle'
+      steps:
+        - uses: actions/checkout@v2
+        - name: Set up JDK
+          uses: actions/setup-java@v2
+          with:
+            java-version: '17'
+            distribution: 'temurin'
+            cache: 'gradle'
 
-      - name: Parse tag
-        # Sets up a step output called TAG with the appropriate tag
-        id: tag
-        run: |
-          TAG_OVERRIDE=${{ github.event.inputs.tagOverride }}
-          if [[ ! -z "${TAG_OVERRIDE}" ]]
-          then
-            echo ::set-output name=TAG::${TAG_OVERRIDE}
-          else
-            echo ::set-output name=TAG::$(git describe --tags)
-          fi
+        - name: Parse tag
+          # Sets up a step output called TAG with the appropriate tag
+          id: tag
+          run: |
+            TAG_OVERRIDE=${{ github.event.inputs.tagOverride }}
+            if [[ ! -z "${TAG_OVERRIDE}" ]]
+            then
+              echo ::set-output name=TAG::${TAG_OVERRIDE}
+            else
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -18,70 +18,103 @@ env:
...
+        - name: Update CBAS Helm Chart in cromwhelm
+          env:
+            BROADBOT_GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
+          run: |
+            set -e
+            cd cromwhelm
+            git checkout main
+            ls -la
+            HELM_CUR_TAG=$(grep "/composite-batch-analysis-service:" cromwell-helm/templates/cbas-deployment.yaml | sed "s,.*/composite-batch-analysis-service:,,")
+            HELM_NEW_TAG=${{ steps.tag.outputs.TAG }}
```
cjllanwarne, line 113: I'm not 100% sure on the difference and scoping between/across "jobs" and "steps"... did you confirm that this correctly references the `tag` step from the `publish-job` job?

.github/workflows/publish.yml
```diff
@@ -18,70 +18,103 @@ env:
...
+
+        - name: Update CBAS Helm Chart in cromwhelm
+          env:
+            BROADBOT_GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
+          run: |
+            set -e
+            cd cromwhelm
+            git checkout main
+            ls -la
+            HELM_CUR_TAG=$(grep "/composite-batch-analysis-service:" cromwell-helm/templates/cbas-deployment.yaml | sed "s,.*/composite-batch-analysis-service:,,")
```
salonishah11, line 112: This might need to be updated to look for version tag in file `cromwhelm/main/terra-batch-libchart/templates/_cbas-api.tpl` because of [this PR in cromwhelm](https://github.com/broadinstitute/cromwhelm/pull/48/files) that merges cromwell-helm with terra-batch-libchart

.github/workflows/publish.yml
```diff
@@ -18,70 +18,103 @@ env:
...
+            HELM_NEW_TAG=${{ steps.tag.outputs.TAG }}
+            [[ -n "$HELM_CUR_TAG" && -n "$HELM_NEW_TAG" ]]
+            sed -i "s/$HELM_CUR_TAG/$HELM_NEW_TAG/" cromwell-helm/templates/cbas-deployment.yaml
```
salonishah11, line 115: Same comment as above

.github/workflows/publish.yml
fredlas: Thanks for the heads-up. I could make it accommodate both the old and new location, but probably makes sense to just block this on that one merging.

...but really this exposes the hackiness of the whole "bash commands to edit a version number in a text file" approach. It will probably break like this again eventually! I wonder if we can just leave off the version to have it default to "most recent". Not familiar with the syntax, whether that's a thing. I'll ask the team and investigate on my own if nobody knows.

fredlas: Ok, a fruitful discussion with Janet indicates no, that is not a good approach. (It may or may not "work" mechanically speaking but is not a good idea and not how we do things).

aednichols (approved):

.github/workflows/publish.yml
```diff
@@ -85,3 +85,28 @@ jobs:
...
+
+      - name: Update CBAS Helm Chart in cromwhelm
+        env:
+          BROADBOT_GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
+        run: |
+          set -e
+          cd cromwhelm
+          git checkout main
+          ls -la
+          HELM_CUR_TAG=$(grep "/composite-batch-analysis-service:" cromwell-helm/templates/cbas-deployment.yaml | sed "s,.*/composite-batch-analysis-service:,,")
```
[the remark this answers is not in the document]
fredlas: Do you mean to write "cbas" rather than "composite-batch-analysis-service"? What is the different file you're referring to?

cjllanwarne (approved):
LGTM. Might be good to test-run with a test tag before merging finally

PR #29, opened 2022-08-05
BW-1305 Adds GET /runs endpoint

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/BW-1305

### DIFF
--- a/service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
@@ -9,7 +9,6 @@ import bio.terra.cbas.model.RunStateResponse;
 import cromwell.client.ApiClient;
 import cromwell.client.api.WorkflowsApi;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 import java.util.UUID;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -22,28 +24,65 @@ public class RunsApiController implements RunsApi {
...
+
+  private Date convertToDate(OffsetDateTime submissionTimestamp) {
+    if (submissionTimestamp != null) {
+      return new Date(submissionTimestamp.toInstant().toEpochMilli());
+    }
+
+    return null;
+  }
+
+  private LogRunRequest convertToLogRunRequest(WorkflowQueryResult queryResult) {
```
cjllanwarne, line 48: This feels more like a `LogRunResponse` than a request?

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -75,17 +112,11 @@ public ResponseEntity<RunStateResponse> postRun(String workflowUrl, Object workf
...
-              new RunStateResponse().runId(runId).state(RunState.QUEUED), HttpStatus.CREATED);
+      return new ResponseEntity<>(
+          new RunStateResponse().runId(runId).state(RunState.QUEUED), HttpStatus.CREATED);
     } catch (cromwell.client.ApiException e) {
-      System.out.println(e);
-      result =
-          new ResponseEntity<>(
-              new RunStateResponse().runId(runId).state(RunState.SYSTEM_ERROR),
-              HttpStatus.INTERNAL_SERVER_ERROR);
+      System.out.println(e.getMessage());
```
cjllanwarne, line 118: We should log the full error even if we only return the short message to the user

cjllanwarne (approved):
LGTM, a few TOL level comments

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
salonishah11: Good point. I will rename it.

kpierre13 (approved):

PR #33, opened 2022-08-16
BW-1341: Add POST run_sets API

cjllanwarne:
## Notes:

- The OpenAPI definition has a little bit of circularity with `ParameterDefinition` declaring it must be `oneOf` its subtypes and its subtypes declaring that they must have `allOf` the `ParameterDefinition` fields. That's not an openapi requirement but it is required to make our specific codegen play nicely. It took a little juggling but we got there in the end.
- You can add `log.info(request.toString());` to the java handler class to see the objects that the codegen creates for us (that's how I produced the examples below). Because of the `oneOf` juggling mentioned above there's a tiny bit more nested than I'd ideally have hand-crafted... but hopefully not too bad, considering.
- In order to avoid breaking changes while the UI catches up, I haven't _removed_ the old `runs` API in this PR. We can do that in a follow up.

## Java Interpretation Examples:

Some examples of how the codegen'd java classes interpret requests

#### JSON value submitted over the API:
```json
{
  "workflow_url": "http://www.example.com/example.wdl",
  "workflow_param_definitions": [
    {
      "parameter_name": "workflow_input_foo_rating",
      "parameter_type": "String",
      "source": {
        "definitionType": "literal",
        "entity_attribute": "hello world"
      }
    }
  ],
  "wds_entities": {
    "entity_type": "FOO",
    "entity_ids": [
      "F0011111-1111-1111-1111-111111111111"
    ]
  }
}
```

#### Java Interpretation:
```
class RunSetRequest {
    workflowUrl: http://www.example.com/example.wdl
    workflowParamDefinitions: [class WorkflowParamDefinition {
        parameterName: workflow_input_foo_rating
        parameterType: String
        source: class ParameterDefinitionLiteralValue {
            class ParameterDefinition {
                definitionType: literal
            }
            parameterValue: hello world
        }
    }]
    wdsEntities: class WdsEntitySet {
        entityType: FOO
        entityIds: [F0011111-1111-1111-1111-111111111111]
    }
}
```

#### JSON value submitted over the API:
```json
{
  "workflow_url": "http://www.example.com/example.wdl",
  "workflow_param_definitions": [
    {
      "parameter_name": "workflow_input_foo_rating",
      "parameter_type": "String",
      "source": {
        "definitionType": "entity_lookup",
        "entity_attribute": "entity_field_foo_rating"
      }
    }
  ],
  "wds_entities": {
    "entity_type": "FOO",
    "entity_ids": [
      "F0011111-1111-1111-1111-111111111111"
    ]
  }
}
```

#### Java Interpretation:
```
class RunSetRequest {
    workflowUrl: http://www.example.com/example.wdl
    workflowParamDefinitions: [class WorkflowParamDefinition {
        parameterName: workflow_input_foo_rating
        parameterType: String
        source: class ParameterDefinitionEntityLookup {
            class ParameterDefinition {
                definitionType: entity_lookup
            }
            entityAttribute: entity_field_foo_rating
        }
    }]
    wdsEntities: class WdsEntitySet {
        entityType: FOO
        entityIds: [F0011111-1111-1111-1111-111111111111]
    }
}
```

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -19,7 +19,7 @@ dependencies {
     implementation 'de.undercouch.download:de.undercouch.download.gradle.plugin:5.0.1'
     implementation 'gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.7.2'
     implementation 'io.spring.dependency-management:io.spring.dependency-management.gradle.plugin:1.0.11.RELEASE'
-    implementation 'org.hidetake.swagger.generator:org.hidetake.swagger.generator.gradle.plugin:2.19.1'
+    implementation 'org.hidetake.swagger.generator:org.hidetake.swagger.generator.gradle.plugin:2.19.2'
     implementation 'org.sonarqube:org.sonarqube.gradle.plugin:3.3'
     implementation 'org.springframework.boot:spring-boot-gradle-plugin:2.6.6'
     implementation 'bio.terra:terra-test-runner:0.1.4-SNAPSHOT'
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -211,6 +211,13 @@ components:
 
     ParameterDefinition:
       title: ParameterDefinition
+      properties:
+        definitionType:
+          type: string
+          description: Indicates that this is a literal value.
+          required: true
+          enum: [ literal, entity_lookup ]
+          example: entity_lookup
       oneOf:
         - $ref: '#/components/schemas/ParameterDefinitionEntityLookup'
         - $ref: '#/components/schemas/ParameterDefinitionLiteralValue'
@@ -225,14 +232,8 @@ components:
       allOf:
         - $ref: '#/components/schemas/ParameterDefinition'
       properties:
-        definitionType:
-          type: string
-          description: Indicates that this is a literal value.
-          required: true
-          enum: [ 'literal' ]
-          example: literal
-        param_value:
-          required: true
+        parameter_value:
+          required: True
           oneOf:
             - type: string
             - type: integer
@@ -247,12 +248,6 @@ components:
       allOf:
         - $ref: '#/components/schemas/ParameterDefinition'
       properties:
-        definitionType:
-          type: string
-          description: Indicates that this is an entity lookup.
-          required: true
-          enum: [ 'entity_lookup' ]
-          example: 'entity_lookup'
         entity_attribute:
           type: string
           description: The attribute name on the entity to use.
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -7,21 +7,6 @@ servers:
   - url: ./
     description: Relative to the current swagger page
 paths:
-  /pets:
-    patch:
-      tags: [ public ]
-      operationId: getPets
-      requestBody:
-        content:
-          application/json:
-            schema:
-              oneOf:
-                - $ref: '#/components/schemas/Cat'
-                - $ref: '#/components/schemas/Dog'
-      responses:
-        '200':
-          description: Updated
-
   /status:
     get:
       summary: Check status of the service
@@ -364,19 +349,3 @@ components:
         - SET_COMPLETE
         - SET_ERROR
       type: string
-
-    Dog:
-      type: object
-      properties:
-        bark:
-          type: boolean
-        breed:
-          type: string
-          enum: [Dingo, Husky, Retriever, Shepherd]
-    Cat:
-      type: object
-      properties:
-        hunts:
-          type: boolean
-        age:
-          type: integer
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -210,16 +210,17 @@ components:
       type: string
 
     # This ParameterDefinition holds the 'oneOf' declaration for all the various types that can
-    # be used as parameter definitions. In addition it acts as the superclass for the various
-    # ParameterDefinitionX classes below. Because the codegen is imperfect, we have to explicitly
+    # be used as parameter definitions.
+    # In addition, in Java it acts as the superclass for the various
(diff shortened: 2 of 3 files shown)


common/openapi.yml
```diff
@@ -1,12 +1,27 @@
...
+  description: Manage batch analyses, run configurations, and run sets in userspace
   version: 0.0.1
 servers:
   - url: ./
     description: Relative to the current swagger page
 paths:
+  /pets:
+    patch:
+      tags: [ public ]
+      operationId: getPets
```
salonishah11, line 13: Is this intentional? 🤔

common/openapi.yml
```diff
@@ -114,6 +158,115 @@ components:
...
+    # This ParameterDefinition holds the 'oneOf' declaration for all the various types that can
+    # be used as parameter definitions.
+    # In addition, in Java it acts as the superclass for the various
+    # ParameterDefinitionXYZ classes below. Because the codegen is imperfect, we have to explicitly
+    # add 'allOf: ParameterDefinition' declarations to those classes to force the codegen to make
+    # them subclasses... but in terms of how the swagger UI works, they seem to make no difference
+    ParameterDefinition:
+      title: ParameterDefinition
+      properties:
+        definitionType:
```
salonishah11, line 221: Should this be `definition_type` to match other fields?

common/openapi.yml
```diff
@@ -114,6 +158,115 @@ components:
...
+          type: string
+          description: Indicates what type of parameter declaration this is.
+          required: true
+          enum: [ literal, entity_lookup ]
+          example: entity_lookup
+      oneOf:
+        - $ref: '#/components/schemas/ParameterDefinitionEntityLookup'
+        - $ref: '#/components/schemas/ParameterDefinitionLiteralValue'
+      discriminator:
+        propertyName: definitionType
```
salonishah11, line 231: Same comment as above

common/openapi.yml
```diff
@@ -202,3 +355,28 @@ components:
...
+      enum:
+        - SET_UNKNOWN_STATE
+        - SET_RUNNING
+        - SET_COMPLETE
+        - SET_ERROR
+      type: string
+
+    Dog:
+      type: object
+      properties:
```
salonishah11, line 370: This and `Cat` object can be removed?

common/openapi.yml
```diff
@@ -202,3 +355,28 @@ components:
...
         - CANCELING
       type: string
+    RunSetState:
+      title: RunSetState
+      example: RUNNING
+      enum:
+        - SET_UNKNOWN_STATE
+        - SET_RUNNING
+        - SET_COMPLETE
+        - SET_ERROR
```
salonishah11, line 365: Should this list be same as [WES responses as we had in RunState](https://github.com/DataBiosphere/cbas/blob/dc134dfc4793a3840903f74cd39cee39fad497ab/common/openapi.yml#L193-L203)?

common/openapi.yml
cjllanwarne: Oops, no, this was left over testing the `oneOf` examples from their docs 🤦

common/openapi.yml
cjllanwarne: I'll try. I changed this because I was testing whether the problem codegen had was to do with camel vs underscore casing. Assuming it works with an underscore that would indeed be preferable... 🤞

common/openapi.yml
cjllanwarne: Happy to if you feel strongly... I thought it would help to make them an obviously different set, to help distinguish between the status types?

common/openapi.yml
cjllanwarne: Indeed, and alas, the discriminator relies on using camel case for this particular field. I can add a comment to that effect.

common/openapi.yml
salonishah11: Would it make sense to add `QUEUED`, `CANCELING` `CANCELLED` statuses? This way when a request is accepted it is put to Queued state and then when picked up to Running state? Also adding the other 2 statuses will support aborting the run set?

common/openapi.yml
cjllanwarne: Resolved by renaming to `type` which doesn't need to worry about casing

common/openapi.yml
cjllanwarne: I'm not sure... I was going to leave those out until we implemented the functionality behind them

buildSrc/build.gradle
```diff
@@ -19,7 +19,7 @@ dependencies {
     implementation 'de.undercouch.download:de.undercouch.download.gradle.plugin:5.0.1'
     implementation 'gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.7.2'
     implementation 'io.spring.dependency-management:io.spring.dependency-management.gradle.plugin:1.0.11.RELEASE'
-    implementation 'org.hidetake.swagger.generator:org.hidetake.swagger.generator.gradle.plugin:2.19.1'
+    implementation 'org.hidetake.swagger.generator:org.hidetake.swagger.generator.gradle.plugin:2.19.2'
```
kpierre13, line 22: This may be a very trivial question, but how do you know when to upgrade a dependency, and how do you determine how far to upgrade it?

salonishah11 (approved):

kpierre13 (approved):

buildSrc/build.gradle
cjllanwarne: I guess that's up for debate... in theory it's better to always be using the latest version. In practice, updating often leads to extra fixup or problem solving if the library has changed since the version we started with.

In this case, I tried updating to see if it helped address the `oneOf` weirdness. I'm not sure it did anything, but since it's newer than the previous version, I'm inclined to keep it anyway

PR #35, opened 2022-08-24
BW-1390 Porting postRun endpoint

kpierre13:
…point

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -123,6 +123,14 @@ components:
         workflow_params:
           type: object
           description: Input set
+#        workflow_attachments:
+#          type: array
+#          description: Files for workflow to run
+#          required: false
+#          items:
+#            type: string
+#            format: binary
+
     RunLog:
       type: object
       properties:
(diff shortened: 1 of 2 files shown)


common/openapi.yml
```diff
@@ -123,6 +123,14 @@ components:
         workflow_params:
           type: object
           description: Input set
+#        workflow_attachments:
```
mspector, line 126: are these commented lines included here on purpose? (perhaps as a reminder of some kind?)

mspector (approved):
Now I see the commit message that says the comments were intentional!

LGTM. If you wanted, you could add a one-liner comment that explains why you're leaving those commented configs in there. But you can judge the importance of that better than me.

cjllanwarne (commented):
Current changes LGTM, but I think we also need to update the `RunSetsApiController`

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -90,31 +92,16 @@ public ResponseEntity<RunStateResponse> postRun(String workflowUrl, Object workf
...
-          null,
-          null,
-          null,
-          null,
-          null,
-          null,
-          null,
-          null,
-          null);
+      wesApi.runWorkflow(workflowParams.toString(), null, null, null, null, workflowUrl, null);
```
salonishah11, line 100: Would you need to update the Cromwell client with changes from [PR 6837](https://github.com/broadinstitute/cromwell/pull/6837) before merging this PR?

kpierre13: I don't think so. The problem was the way that Wes was submitting the workflows on the cromwell end, so nothing on the client side would need to be updated that I'm aware of.

cjllanwarne (approved):

salonishah11 (approved):

PR #38, opened 2022-09-14
BW-1375 - GET /runs Retrieves From Database

kpierre13:


service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -20,4 +30,37 @@ public int createRun(Run run) {
...
+
+  public List<Run> retrieve() {
+    String sql =
+        "SELECT * FROM run INNER JOIN run_set ON run.run_set_id = run_set.id"
+            + " INNER JOIN method ON run_set.method_id = method.id";
+
+    try {
+      return jdbcTemplate.query(sql, new RunMapper());
+    } catch (EmptyResultDataAccessException er) {
+      throw new RunQueryNotFoundException("Runs not found.");
```
cjllanwarne, line 42: This doesn't feel quite right... if there are no runs then don't we get an empty result back? So I wonder what would cause this exception to happen?

cjllanwarne (approved):
LGTM, but I'm surprised that status is being excluded?

common/openapi.yml
```diff
@@ -278,12 +278,14 @@ components:
...
-        state:
-          $ref: '#/components/schemas/RunState'
-        workflow_params:
-          type: string
+# The below information is not currently returned in our current model on the API.
+# May be returned in later work.
+#        name:
+#          type: string
+#        state:
+#          $ref: '#/components/schemas/RunState'
```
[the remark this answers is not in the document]
kpierre13: Oversight by myself. Will add back on next commit.

service/src/main/java/bio/terra/cbas/dao/RunDao.java
kpierre13: Right, I think I was thinking in the mindset of entering an id, and having that id match or not match data in the db, but we're returning everything, so this isn't really necessary. I'll remove on next commit.

cjllanwarne (approved):

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -55,15 +55,15 @@ private RunLog runToRunLog(Run run) {
     return new RunLog()
         .runId(run.id().toString())
         .workflowUrl(run.runSet().method().methodUrl())
-        .submissionDate(convertToDate(run.submissionTimestamp()));
+        .submissionDate(convertToDate(run.submissionTimestamp()))
```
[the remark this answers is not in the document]
cjllanwarne: Spoke in person. Not this line, has been spun out into https://broadworkbench.atlassian.net/browse/BW-1411

service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -20,4 +28,33 @@ public int createRun(Run run) {
             + " values (:id, :engineId, :runSetId, :entityId, :submissionTimestamp, :status)",
         new BeanPropertySqlParameterSource(run));
   }
+
+  public List<Run> retrieve() {
+    String sql =
+        "SELECT * FROM run INNER JOIN run_set ON run.run_set_id = run_set.id"
+            + " INNER JOIN method ON run_set.method_id = method.id";
```
aednichols, line 35: I am a bit surprised we are not using some kind of abstraction layer, but selecting such would probably be beyond the scope of this story

aednichols (approved):

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -47,44 +50,21 @@ private Date convertToDate(OffsetDateTime submissionTimestamp) {
...
+  private RunLog runToRunLog(Run run) {
 
-  @Autowired
-  public RunsApiController(CromwellServerConfiguration cromwellConfig) {
-    this.cromwellConfig = cromwellConfig;
+    return new RunLog()
+        .runId(run.id().toString())
+        .workflowUrl(run.runSet().method().methodUrl())
+        .submissionDate(convertToDate(run.submissionTimestamp()))
+        .state(convertToRunState(run.status()));
```
salonishah11, line 59: This should also return workflow name (`name`) and input params definition ( `workflow_params`)

salonishah11 (changes requested):

service/src/main/java/bio/terra/cbas/dao/RunDao.java
cjllanwarne: This is also what happens in "the other Terra Tech Stack projects we copied from", FWIW

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -47,44 +52,25 @@ private Date convertToDate(OffsetDateTime submissionTimestamp) {
...
     return new RunLog()
-        .runId(queryResult.getId())
-        .state(convertToRunState(queryResult.getStatus()))
-        .workflowUrl(null)
-        .name(queryResult.getName())
+        .runId(run.id().toString())
+        .workflowUrl(run.runSet().method().methodUrl())
+        .name(run.entityId())
+        .state(convertToRunState(run.status()))
         .workflowParams(null)
```
salonishah11, line 62: I think this can now be fetched from the database and returned. We might be storing it in `methods` table

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -47,44 +52,25 @@ private Date convertToDate(OffsetDateTime submissionTimestamp) {
...
+  private RunLog runToRunLog(Run run) {
+
     return new RunLog()
-        .runId(queryResult.getId())
-        .state(convertToRunState(queryResult.getStatus()))
-        .workflowUrl(null)
-        .name(queryResult.getName())
+        .runId(run.id().toString())
+        .workflowUrl(run.runSet().method().methodUrl())
+        .name(run.entityId())
```
salonishah11, line 60: Do we store workflow/run name in database?

kpierre13: Entity_id was the closest thing I found to storing a name. I didn't find anything specifically identified as 'name' in the database.

salonishah11: I guess then for now it can return `null` and we can populate it later on when we have the name. What do you think?

salonishah11 (approved):
For now it might be better to return `null` for Workflow Name instead of entity ID. Apart from this nit, LGTM 👍

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
kpierre13: This makes sense to me

PR #41, opened 2022-09-20
BW-1385: Smart Polling on run queries

cjllanwarne:


service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -15,23 +16,12 @@
...
-    return switch (workflowStatus) {
-      case "On Hold" -> RunState.PAUSED;
-      case "Submitted" -> RunState.QUEUED;
-      case "Running" -> RunState.RUNNING;
-      case "Aborting" -> RunState.CANCELING;
-      case "Aborted" -> RunState.CANCELED;
-      case "Succeeded" -> RunState.COMPLETE;
-      case "Failed" -> RunState.EXECUTOR_ERROR;
-      default -> RunState.UNKNOWN;
-    };
```
cjllanwarne, line 34: These were Cromwell statuses, never returned by WES. WES statuses are an exact match for CBAS statuses... though it might be nice to separate out "API types" from "internal types" in case that statement slowly drifts away from truth

service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -29,14 +31,19 @@ public int createRun(Run run) {
         new BeanPropertySqlParameterSource(run));
   }
 
-  public List<Run> retrieve() {
+  public List<Run> getRuns() {
```
cjllanwarne, line 34: This was a personal preference choice... happy to revert if people don't like it!

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -0,0 +1,78 @@
...
+    // Filter only updatable runs:
+    Stream<Run> updatableRuns = runs.stream().filter(this::nonTerminal);
+
+    // Group by current (engine) status:
+    Map<State, List<Run>> engineStatus =
+        updatableRuns.collect(
+            groupingBy(
+                r -> {
+                  try {
+                    return cromwellService.runStatus(r.engineId()).getState();
```
pshapiro4broad, line 42: If there are a lot of statuses to check this could possibly be improved by calling `runStatus()` in parallel. replacing `groupingBy()` with `groupingByConcurrent()` should do this.

cjllanwarne: For context - it's known and accepted that this v0.2 is not going to scale well... I agree that polling in parallel is probably in our future, but perhaps by batching bigger questions to the engine rather than asking all the individual questions in parallel. In other words... tbd!

(Very glad to find out about `groupingByConcurrent()` though!)

service/src/main/java/bio/terra/cbas/dao/RunDao.java
mspector: I like it

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -0,0 +1,85 @@
...
+  private final RunDao runDao;
+
+  private static final org.slf4j.Logger logger = LoggerFactory.getLogger(SmartRunsPoller.class);
+
+  public SmartRunsPoller(CromwellService cromwellService, RunDao runDao) {
+    this.cromwellService = cromwellService;
+    this.runDao = runDao;
+  }
+
+  public List<Run> updateRuns(List<Run> runs) {
```
mspector, line 32: my first reaction here was to want a top-level comment briefly describing what this function does (and perhaps what makes it "smart" polling).

that being said, I basically understand what's happening here, and think it's fine. so I won't let this suggestion block approval

mspector (approved):

service/src/main/java/bio/terra/cbas/models/Run.java
```diff
@@ -14,4 +15,20 @@ public record Run(
...
+  public Run withStatus(String newStatus) {
+    return new Run(id, engineId, runSet, entityId, submissionTimestamp, newStatus);
+  }
+
+  public boolean isTerminal() {
+    RunState state = RunState.fromValue(this.status());
+    return RunState.CANCELED.equals(state)
+        || RunState.COMPLETE.equals(state)
+        || RunState.EXECUTOR_ERROR.equals(state)
+        || RunState.SYSTEM_ERROR.equals(state);
```
pshapiro4broad, line 28: If `RunState` is an enum, and `state` is non-`null`, I think it's more idiomatic to compare enums using `==` instead of `equals()`. Java enums are singleton objects to `==` is an OK way to compare them.

Or maybe a better way overall is to use an `EnumSet<>` to represent the known terminal states:

```
  private static final Set<RunState> TERMINAL_STATES = EnumSet.of(RunState.CANCELED, RunState.COMPLETE,
        RunState.EXECUTOR_ERROR, RunState.SYSTEM_ERROR);
  ...
  public boolean isTerminal() {
    return TERMINAL_STATES.contains(RunState.fromValue(status));
  }
```
This doesn't mind `null` (it returns `false`) but if `null` is possible maybe you'd want to handle that explicitly.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
salonishah11: I agree. A high level comment would be nice to have.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -0,0 +1,85 @@
...
+
+    // For metrics:
+    OffsetDateTime startTime = OffsetDateTime.now();
+
+    // Filter only updatable runs:
+    List<Run> updatableRuns = runs.stream().filter(Run::nonTerminal).toList();
+
+    // This has the nice outcome of counting even if the size is 0, which means this metric
+    // is created and stays current even if nothing is being updated:
+    logger.debug("METRIC: COUNT runs needing status update: {}", updatableRuns.size());
```
salonishah11, line 42: TOL: Do we want to add a TODO here saying something like "convert this and below log messages into metrics when metrics framework has been added"?

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -0,0 +1,85 @@
...
+    Set<Run> updatedRuns = new HashSet<>(runs);
+
+    for (Map.Entry<State, List<Run>> engineStateEntry : engineStatuses.entrySet()) {
+      for (Run r : engineStateEntry.getValue()) {
+        var currentState = engineStateEntry.getKey().toString();
+        if (!r.status().equals(currentState)) {
+          logger.debug("Updating status of Run {} (engine ID {})", r.id(), r.engineId());
+          logger.debug("METRIC: INCREMENT runs transitioned to final status");
+          var changes = runDao.updateRunStatus(r, currentState);
+          if (changes == 1) {
```
salonishah11, line 72: Do we want to add an `else` case in the situation it was not able to update the record? Or something like stop trying to update the record after X attempts? Here it seems it might keep trying until it was able to update the status?

service/src/main/java/bio/terra/cbas/models/Run.java
cjllanwarne: Oh, I like the set contains a lot. 

FWIW, I'm curious what your take on this thought is: I think long term this mixing of WES and CBAS API enums within business logic (especially writing to and reading from the database) is going to cause headaches. Is it normal to want to create yet-another-almost-parallel enum for business logic and database usages, with converters into and out of potentially different sets of API enums only on the interfaces?

pshapiro4broad: I've thought about this but don't have a great answer. The main problem I see is with auto generated code like `RunState`. For this code, I think what you really want is to make `isTerminal()` a property of `RunState` directly but there's no easy way that I know of to add code to an openapi generated enum.

Using parallel objects, and only using the openapi generated classes in the controller layer, seems like a good approach. At the controller layer you convert to your "real" objects, and convert "real" objects back to openapi objects. So any code outside the controller doesn't need to know about these objects. You can make this conversion easier by adding code to your "real" objects that convert from/to the generated objects as necessary. For example: https://github.com/DataBiosphere/terra-data-catalog/blob/main/service/src/main/java/bio/terra/catalog/common/StorageSystem.java

I like this approach since it isolates logic from the openapi layer. But it's a lot of work to do for all your input/output objects, and even in a very small service like catalog we haven't been consistent in doing it.

If you can use the same object for both database and business logic then it seems worth it to create a parallel object.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
cjllanwarne: I was hoping the big `METRIC` tag would be sufficient?

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
cjllanwarne: So the `else` case is to "do nothing and try again next time", which of course already happens... but logging and metrics for that seems like a good idea, so I'll add those in 👍

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
salonishah11: That seems fair 👍

PR #43, opened 2022-09-26
BW-1348 Support Updating WDS Rows 

kpierre13:


salonishah11 (approved):

common/openapi.yml
```diff
@@ -180,36 +198,36 @@ components:
...
           description: Indicates what type of parameter declaration this is.
           required: true
-          enum: [ literal, entity_lookup ]
-          example: entity_lookup
+          enum: [ literal, record_lookup ]
+          example: record_lookup
       oneOf:
-        - $ref: '#/components/schemas/ParameterDefinitionEntityLookup'
+        - $ref: '#/components/schemas/ParameterDefinitionRecordLookup'
         - $ref: '#/components/schemas/ParameterDefinitionLiteralValue'
```
salonishah11, line 205: Just a heads up, this and above conversion of `entity_lookup` to `record_lookup` is also being done in PR #44 . Depending who merges first there might be merge conflicts.

salonishah11 (approved):

common/openapi.yml
```diff
@@ -122,13 +122,17 @@ components:
...
           items:
-            $ref: '#/components/schemas/WorkflowParamDefinition'
+            $ref: '#/components/schemas/WorkflowInputDefinition'
           description: Declarative input definition for combination with WDS entities
         wds_entities:
           $ref: '#/components/schemas/WdsEntitySet'
+        workflow_output_definitions:
+          type: array
+          items:
+            $ref: '#/components/schemas/WorkflowOutputDefinition'
```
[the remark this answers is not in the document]
kpierre13: Yeah, I can add that in as well.

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -82,19 +82,19 @@ void chainCallsTogether() throws Exception {
         """
         {
           "workflow_url" : "%s",
-          "workflow_param_definitions" : [ {
-            "parameter_name" : "myworkflow.mycall.inputname1",
-            "parameter_type" : "String",
+          "workflow_input_definitions" : [ {
```
cjllanwarne, line 85: You probably want to add an output definition here too, to prove that the API definition (and thus JSON parsing) is all working properly.

cjllanwarne (approved):
LGTM, one small suggestion for improving an existing test case

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
kpierre13: These should be all set! Checks were failing as a result of these being mismatched, but they're passing now.

PR #44, opened 2022-09-26
[BW-1386] replace term 'entity' with 'record'

mspector:
...to accommodate new WDS client terminology

### DIFF
--- a/service/src/main/java/bio/terra/cbas/config/WdsServerConfiguration.java
+++ b/service/src/main/java/bio/terra/cbas/config/WdsServerConfiguration.java
@@ -1,6 +1,6 @@
 package bio.terra.cbas.config;
 
-import org.databiosphere.workspacedata.api.EntitiesApi;
+import org.databiosphere.workspacedata.api.RecordsApi;
 import org.databiosphere.workspacedata.client.ApiClient;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
@@ -11,10 +11,10 @@ public record WdsServerConfiguration(
     return "%s/%s".formatted(baseUri, healthcheckEndpoint);
   }
 
-  public EntitiesApi entitiesApi() {
+  public RecordsApi recordsApi() {
     ApiClient wdsApiClient = new ApiClient();
     wdsApiClient.setBasePath(baseUri);
 
-    return new EntitiesApi(wdsApiClient);
+    return new RecordsApi(wdsApiClient);
   }
 }
--- a/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
@@ -23,7 +23,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import org.databiosphere.workspacedata.client.ApiException;
-import org.databiosphere.workspacedata.model.EntityResponse;
+import org.databiosphere.workspacedata.model.RecordResponse;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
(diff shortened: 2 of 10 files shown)


service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -85,11 +85,11 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
     runSetDao.createRunSet(runSet);
 
     // Fetch the entity from WDS:
-    EntityResponse entityResponse;
+    RecordResponse recordResponse;
```
salonishah11, line 88: Same comment about `records` here from above. Should this be names `WdsRecordResponse`?

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
```diff
@@ -16,8 +16,8 @@ public MethodDao(NamedParameterJdbcTemplate jdbcTemplate) {
 
   public int createMethod(Method method) {
     return jdbcTemplate.update(
-        "insert into method (id, method_url, input_definition, entity_type)"
-            + " values (:id, :methodUrl, :inputDefinition, :entityType)",
+        "insert into method (id, method_url, input_definition, record_type)"
+            + " values (:id, :methodUrl, :inputDefinition, :recordType)",
```
salonishah11, line 20: I think you also need to update `method` table definition in database by adding a new changeset in `resources/changelog/changesets/`.

service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -25,7 +25,7 @@ public RunDao(NamedParameterJdbcTemplate jdbcTemplate) {
   public int createRun(Run run) {
     return jdbcTemplate.update(
         "insert into run (id, engine_id, run_set_id, entity_id, submission_timestamp, status)"
-            + " values (:id, :engineId, :runSetId, :entityId, :submissionTimestamp, :status)",
+            + " values (:id, :engineId, :runSetId, :recordId, :submissionTimestamp, :status)",
```
salonishah11, line 28: I think you also need to update `run` table definition in database by adding a new changeset in `resources/changelog/changesets/`.

service/src/main/java/bio/terra/cbas/config/WdsServerConfiguration.java
```diff
@@ -11,10 +11,10 @@ public String healthcheckUri() {
     return "%s/%s".formatted(baseUri, healthcheckEndpoint);
   }
 
-  public EntitiesApi entitiesApi() {
+  public RecordsApi recordsApi() {
```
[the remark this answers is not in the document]
mspector: no objections here!

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
mspector: is this for database migration? if so, are we at the point now where we want to be accumulating these migrations? (genuine question; I'm not sure how this works / what it's for)

service/src/main/java/bio/terra/cbas/config/WdsServerConfiguration.java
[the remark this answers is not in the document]
mspector: wait, just to clarify: `RecordsApi`, the class, is imported from WDS (so to change that we'd have to ask the WDS team)

should we change the function name to `wdsRecordsApi`?

(perhaps that's what you meant, and I'm focusing too much on the capitalization)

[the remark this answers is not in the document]
salonishah11: Ah I see. I didn't realize `RecordsApi` was being imported from WDS. If that's the case then let's keep the function name also `recordsApi()` just to be consistent.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Same thing here. I didn't realize `RecordResponse` is class from WDS library. It makes sense for the variable name to be `recordResponse` to be consistent.

salonishah11 (approved):

PR #45, opened 2022-09-27
[WM-1413] Store output definition in database

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/WM-1413


service/src/main/resources/changelog/changesets/20220823_method.yaml
```diff
@@ -22,6 +22,11 @@ databaseChangeLog:
                 type: text
                 constraints:
                   nullable: false
+            - column:
+                name: output_definition
+                type: text
+                constraints:
+                  nullable: true
```
cjllanwarne, line 29: This `nullable: true` is unlike `input_definition`. Is that deliberate?

cjllanwarne (approved):

service/src/main/resources/changelog/changesets/20220823_method.yaml
salonishah11: Yes it is intentional. This is to support the case where user doesn't want to write the outputs back to data table (similar to Terra UI / Rawls). Is my assumption correct?

kpierre13 (approved):

service/src/main/resources/changelog/changesets/20220823_method.yaml
cjllanwarne: Is that different from providing an empty list? I would almost prefer an empty list because (1) the API would need to provide an empty list anyway, and (2) we don't have to have a special case handler for nulls coming from the database. But up to you!

(I suspect "not attached to data table at all" is also a pretty solid "it doesn't make sense to have output definitions" use case, but the inputs currently assume a data table record type, so we might as well do the same in outputs... for now?)

salonishah11: I think I like the idea of UI sending empty list to CBAS because I don't like nulls from database. So that would mean we set `nullable` to `false` and have UI send `[]` when user doesn't enter any data? Right?

cjllanwarne: Yeah that sounds right

PR #48, opened 2022-10-03
[BW-1426] Add `last_modified_timestamp` and `last_polled_timestamp` to runs

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/WM-1426

### DIFF
--- a/service/src/main/java/bio/terra/cbas/dao/RunDao.java
+++ b/service/src/main/java/bio/terra/cbas/dao/RunDao.java
@@ -26,8 +26,8 @@ public class RunDao {
 
   public int createRun(Run run) {
     return jdbcTemplate.update(
-        "insert into run (id, engine_id, run_set_id, record_id, submission_timestamp, status)"
-            + " values (:id, :engineId, :runSetId, :recordId, :submissionTimestamp, :status)",
+        "insert into run (id, engine_id, run_set_id, record_id, submission_timestamp, status, last_modified_timestamp, last_polled_timestamp)"
+            + " values (:id, :engineId, :runSetId, :recordId, :submissionTimestamp, :status, :lastModifiedTimestamp, :lastPolledTimestamp)",
         new EnumAwareBeanPropertySqlParameterSource(run));
   }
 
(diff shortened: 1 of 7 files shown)


salonishah11:
I have updated the code to also return `last_modified_timestamp` in GET /runs API response.

kpierre13 (approved):

service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -39,9 +39,29 @@ public List<Run> getRuns() {
...
+                run.id(),
+                "status",
+                newStatus.toString(),
+                "lastModifiedTimestamp",
+                currentTimestamp,
+                "lastPolledTimestamp",
+                currentTimestamp)));
+  }
+
+  public int updateLastPolledTimestamp(UUID runID) {
```
cjllanwarne, line 59: TOL: Since these will _always_ be updated for every run, is it possible to update the whole set of runs together in bulk rather than one at a time?

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -89,6 +89,15 @@ public List<Run> updateRuns(List<Run> runs) {
...
                 currentState);
           }
+        } else {
+          // if run status hasn't changed, only update last polled timestamp
+          var changes = runDao.updateLastPolledTimestamp(r.id());
+          if (changes != 1) {
+            logger.warn(
+                "Something went wrong while updating last_polled_timestamp for Run {} which is in status {}.",
+                r.id(),
+                r.status());
```
cjllanwarne, line 99:
```suggestion
                "Expected 1 row change updating last_polled_timestamp for Run {} in status {}, but got {}.",
                r.id(),
                r.status(),
                changes);
```

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
```diff
@@ -69,10 +74,13 @@ void pollRunningRuns() throws ApiException {
     when(cromwellService.runStatus(eq(runningRunEngineId)))
         .thenReturn(new RunStatus().runId(runningRunEngineId).state(State.RUNNING));
 
+    when(runsDao.updateLastPolledTimestamp(eq(runToUpdate.id()))).thenReturn(1);
+
     var actual = smartRunsPoller.updateRuns(List.of(runToUpdate, runAlreadyCompleted));
 
     verify(cromwellService).runStatus(eq(runningRunEngineId));
     verify(cromwellService, never()).runStatus(eq(completedRunEngineId));
+    verify(runsDao).updateLastPolledTimestamp(eq(runToUpdate.id()));
```
cjllanwarne, line 83: Do we also want to verify that this is _never_ called for `runAlreadyCompleted`?
```suggestion
    verify(runsDao).updateLastPolledTimestamp(eq(runToUpdate.id()));
    verify(runsDao, never()).updateLastPolledTimestamp(eq(runAlreadyCompleted.id()));
```

cjllanwarne (approved):
LGTM, just a few minor comments

service/src/main/java/bio/terra/cbas/dao/RunDao.java
salonishah11: So when we call `updateRunStatus` we only update the `last_modified_timestamp`. And at the end in SmartPoller, before we return the list of Runs, we do a bulk update for `last_polled_timestamp`?

cjllanwarne: Something like that. Not necessary for this PR, we could keep it in our back pockets in case things are scaling badly in the future...

PR #50, opened 2022-10-05
WM-1414 Attach Workflow Outputs

kpierre13:


service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
```diff
@@ -36,11 +46,29 @@ public class TestSmartRunsPoller {
   private static final String completedRunEntityId = UUID.randomUUID().toString();
   private static final OffsetDateTime completedRunSubmittedtime = OffsetDateTime.now();
 
+  public ObjectMapper objectMapper =
+      new ObjectMapper()
+          .registerModule(new ParameterNamesModule())
+          .registerModule(new Jdk8Module())
+          .registerModule(new JavaTimeModule())
+          .setDateFormat(new StdDateFormat())
+          .setDefaultPropertyInclusion(JsonInclude.Include.NON_ABSENT);
```
cjllanwarne, line 55: Because you ended up using Gson, I suspect we no longer need this `objectMapper`?

kpierre13: Got rid of it here, but had to pass in `null` as an arg for the two tests, as `SmartRunsPoller` has `objectMapper` as part of its constructor.

common/openapi.yml
```diff
@@ -219,6 +219,11 @@ components:
           description: The literal value to use. Can be any type, but must match the declared input_type or output_type.
           required: True
           example: hello world
+        record_attribute:
```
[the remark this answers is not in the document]
kpierre13: Hm, seems like I added this accidentally. Will undo.

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
kpierre13: So, I forgot that the test is also directly testing `SmartRunsPoller` which does in fact use `objectMapper`, so I had to put it back in an pass it as in arg in the test to get it to pass.

service/src/main/java/bio/terra/cbas/runsets/outputs/OutputGenerator.java
```diff
@@ -0,0 +1,22 @@
+package bio.terra.cbas.runsets.outputs;
+
+import bio.terra.cbas.model.WorkflowOutputDefinition;
+import java.util.List;
+import java.util.Map;
+import org.databiosphere.workspacedata.model.RecordAttributes;
+
+public class OutputGenerator {
+  public static RecordAttributes buildOutputs(
```
salonishah11, line 9: It would be great to add tests for OutputGenerator similar to ones we have for [InputGenerator](https://github.com/DataBiosphere/cbas/tree/main/service/src/test/java/bio/terra/cbas/runsets/inputs).

service/src/main/java/bio/terra/cbas/runsets/outputs/OutputGenerator.java
```diff
@@ -0,0 +1,22 @@
...
+      List<WorkflowOutputDefinition> outputDefinitions, Object cromwellJsonObject) {
+    RecordAttributes outputParams = new RecordAttributes();
+    for (WorkflowOutputDefinition outputParam : outputDefinitions) {
+      String parameterName = outputParam.getOutputName();
+      Object outputValue;
+      outputValue = ((Map<String, Object>) cromwellJsonObject).get(parameterName);
```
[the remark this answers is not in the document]
cjllanwarne: Handling this is a really good idea. Ideally we'd be validating methods before submitting, but even then it's possible for things to go wrong and we should be graceful to that possibility here.

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -47,8 +47,13 @@ private RunLog runToRunLog(Run run) {
   public ResponseEntity<RunLogResponse> getRuns() {
 
     List<Run> queryResults = runDao.getRuns();
-    List<Run> updatedRunResults = smartPoller.updateRuns(queryResults);
-
+    List<Run> updatedRunResults;
+    try {
+      updatedRunResults = smartPoller.updateRuns(queryResults);
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
cjllanwarne: I like the idea of runs being units which we can try to update individually. We should probably also refactor this to not be a single giant method... (but I was holding this back because it's not really Katrina's fault and might happen anyway as part of a larger refactor)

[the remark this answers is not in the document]
salonishah11: Exceptions are now handled inside `updateRuns`.

service/src/main/java/bio/terra/cbas/runsets/outputs/OutputGenerator.java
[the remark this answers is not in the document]
salonishah11: The case where output name is not found is handled in this method.

service/src/main/java/bio/terra/cbas/runsets/outputs/OutputGenerator.java
salonishah11: Added tests for OutputGenerator.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -89,34 +99,32 @@ public List<Run> updateRuns(List<Run> runs) throws Exception {
...
+        var updatedRunState = engineStateEntry.getKey();
+        if (r.status() != updatedRunState) {
+          if (updatedRunState == CbasRunStatus.COMPLETE) {
+            try {
+              updateOutputAttributes(r);
+            } catch (Exception e) {
+              // log error and mark Run as Failed
+              // TODO: When epic WM-1433 is being worked on, add error message in database stating
+              //  updating output attributes failed for this particular Run.
+              logger.error("Error while updating output attributed for run {}.", r.id(), e);
```
cjllanwarne, line 111: Small typo. Also can we include the record we're failing to update? (Warning: I'd double check this suggestion in case I got the name wrong!)
```suggestion
              logger.error("Error while updating attributes for record {} from run {}.", r.recordId() r.id(), e);
```

service/src/main/java/bio/terra/cbas/runsets/outputs/OutputGenerator.java
```diff
@@ -6,17 +6,21 @@
...
+
   public static RecordAttributes buildOutputs(
-      List<WorkflowOutputDefinition> outputDefinitions, Object cromwellJsonObject) {
-    RecordAttributes outputParams = new RecordAttributes();
-    for (WorkflowOutputDefinition outputParam : outputDefinitions) {
-      String parameterName = outputParam.getOutputName();
-      Object outputValue;
-      outputValue = ((Map<String, Object>) cromwellJsonObject).get(parameterName);
-      String attributeName = outputParam.getRecordAttribute();
+      List<WorkflowOutputDefinition> outputDefinitions, Object cromwellOutputs) throws Exception {
```
cjllanwarne, line 11: TOL/Nit: Can we be more specific than "Exception"?

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
```diff
@@ -198,4 +199,58 @@ void updateNewlyCompletedRuns() throws Exception {
         COMPLETE,
         actual.stream().filter(r -> r.id().equals(completedRunId)).toList().get(0).status());
   }
+
+  @Test
+  void updatingOutputFails() throws Exception {
```
cjllanwarne, line 204: TOL/Nit: Do you also want to check that any other workflows around the failing update still manage to succeed?

cjllanwarne, line 204: TOL/Nit 2: Is it possible to restructure the flow of the updater method so that we can test various cases of the "record updates can succeed or fail independently of each other" behavior without having to spin up all this infrastructure of an API test each time?

cjllanwarne (approved):
Looks good to me. All comments are suggestions not demands, and can be punted on if you just want to merge the changes! 😆

service/src/main/java/bio/terra/cbas/runsets/outputs/OutputGenerator.java
salonishah11: Yeah I can add a new Exception type.

salonishah11: Added new Exception type called `WorkflowOutputNotFoundException`

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
salonishah11: I have updated this test to test this scenario.

salonishah11: I like the idea of updating this logic so we can test various scenarios easily. But I think refactoring this code would also require refactoring `updateRuns` which might be non-trivial amount of work and can be done in another PR/ticket.

cjllanwarne: I agree!

service/src/main/java/bio/terra/cbas/common/exceptions/WorkflowOutputNotFoundException.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.cbas.common.exceptions;
+
+public class WorkflowOutputNotFoundException extends Exception {
```
cjllanwarne, line 3: I think this is fine for now. If we end up with lots of individual exceptions for every possible error case, we might want to consolidate under a "workflow processing exception" with individual error messages, or have these all extend from a common super class. But as I said, I think it's fine for now.

cjllanwarne (approved):

mspector (approved):

PR #52, opened 2022-10-11
[WM-1443] multiple records run sets api

mspector:


service/src/main/java/bio/terra/cbas/common/MetricsUtil.java
```diff
@@ -161,6 +173,10 @@ public static void registerAllViews() {
                     25000.0,
                     50000.0)));
 
+    Aggregation recordsPerRequestDistribution =
+        Aggregation.Distribution.create(
+            BucketBoundaries.create(Arrays.asList(0.0, 10.0, 20.0, 30.0, 40.0, 50.0, 100.0)));
```
cjllanwarne, line 178: I was going to say "we can probably drop the 0"... but maybe it's worth leaving, just in case we have some bug that means we're submitting 0 workflows to CBAS from time to time...

cjllanwarne, line 178: Can/should these be integers instead of `.0`s?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -118,6 +124,8 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
     // Store the run:
     UUID runId = UUID.randomUUID();
 
+    // is there a reason that this expression is here again? why not put it in a variable?
+    // e.g. String dataTableRowId = recordIds.get(0);
```
cjllanwarne, line 128: Probably better as a tech talk or jira ticket rather than a comment in the code... I imagine finding this comment here in 10 years and wondering what to make of it.

(though FWIW... I believe Saloni's changes to iterate over the list rather than pick only the first will probably mean it's only like this temporarily)

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -34,14 +35,42 @@
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.http.MediaType;
 import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.MvcResult;
 
 @WebMvcTest
-@ContextConfiguration(classes = RunSetsApiController.class)
+@TestPropertySource(properties = "cbas.cbas-api.runSetsMaximumRecordIds=2")
```
cjllanwarne, line 43: Oh this is cool! Does this let you use `records` again (since you don't need to mock it, presumably it being final is no longer a problem...? 🤞 )

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -148,4 +153,32 @@ void chainCallsTogether() throws Exception {
         newRunCaptor.getValue().submissionTimestamp(),
         greaterThan(OffsetDateTime.now().minus(Duration.ofSeconds(60))));
   }
+
+  @Test
+  void tooManyRecordIds() throws Exception {
```
cjllanwarne, line 158: Also worth checking "exactly the right number of record IDs"?

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -148,4 +153,32 @@ void chainCallsTogether() throws Exception {
...
+        } ]""";
+
+    String request =
+        requestTemplate.formatted(workflowUrl, outputDefinitionAsString, recordType, recordIds);
+
+    MvcResult result =
+        mockMvc
+            .perform(post(API).content(request).contentType(MediaType.APPLICATION_JSON))
+            .andExpect(status().is4xxClientError())
+            .andReturn();
```
cjllanwarne, line 182: Interesting..... presumably this works because all of the mocks are created outside of the test method? But because all of the "when(x) / do(y)" setup is done in the `chainCallsTogether` test, is this now brittle to running these tests in the opposite order? Or running this test but not the first one?

service/src/main/java/bio/terra/cbas/app/App.java
```diff
@@ -28,7 +28,7 @@
       // Metrics exporting components & configs
       "bio.terra.common.prometheus"
     })
-@ConfigurationPropertiesScan("bio.terra.cbas")
+@ConfigurationPropertiesScan("bio.terra.cbas.config")
```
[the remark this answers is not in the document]
mspector: I think you're right that it's not, actually - will revert this

service/src/main/java/bio/terra/cbas/common/MetricsUtil.java
mspector: it seems like `BucketBoundaries` wants a list of Doubles: https://javadoc.io/doc/io.opencensus/opencensus-api/0.21.0/io/opencensus/stats/BucketBoundaries.html

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
mspector: this was mostly a reminder to myself to raise it at tech talk -- which, actually, I failed to do

I'll remove it

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
mspector: it doesn't seem to let me use `record`, sadly... there must be some piece that we're still missing to make that work.

this was Phil's suggestion -- he argued that configurations shouldn't be "mocked" as much as there should be a (real) configuration specifically for testing

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
mspector: easy enough! though I won't check all the stuff that happens after the request is successfully returned -- that should be in the ticket that follows this one

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -148,4 +154,109 @@ void chainCallsTogether() throws Exception {
...
+            .perform(post(API).content(request).contentType(MediaType.APPLICATION_JSON))
+            .andExpect(status().isOk())
+            .andReturn();
+
+    // Validate that the response can be parsed as a valid RunSetStateResponse:
+    RunSetStateResponse response =
+        objectMapper.readValue(
+            result.getResponse().getContentAsString(), RunSetStateResponse.class);
+
+    // Verify database storage:
```
mspector, line 207: the below is mostly copy/pasted from `chainCallsTogether` - there's a lot about the code below that I don't understand.

this is intended to be essentially the same test as `chainCallsTogether`, except with the maximum number of record IDs.

Until we implement the rest of this, it should have the same result. I suppose this test would be updated in the next ticket.

salonishah11 (approved):

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
mspector: no need to apologize -- honestly, a (less well-articulated) version of this thought occurred to me, but I found myself not wanting to engage in a back-and-forth about it, which was a mistake on my part. I should have raised this instead of blindly marching forward.

I think your suggestion for a unit test on the inner function is the best solution. Saloni's upcoming PR should include changes to the original `chainCallsTogether` test such that it takes more than one record ID. It doesn't seem worthwhile to have multiple copies of `chainCallsTogether` for different numbers of inputs.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -177,16 +174,15 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
 
-  private static Optional<ResponseEntity<RunSetStateResponse>> checkInvalidRequest(
-      RunSetRequest request, int runSetsMaximumRecordIds) {
+  public static boolean requestIsValid(RunSetRequest request, CbasApiConfiguration config) {
     int recordIdsSize = request.getWdsRecords().getRecordIds().size();
-    if (recordIdsSize > runSetsMaximumRecordIds) {
+    int recordIdsMax = config.getRunSetsMaximumRecordIds();
+    if (recordIdsSize > recordIdsMax) {
       log.warn(
           "Bad user request: %s record IDs submitted exceeds the maximum value of %s"
```
cjllanwarne, line 182: **Not for this PR**, just something to think about for the future... once we are able to return error messages back to the user in "failure" responses, how might we bring this message up so we can include it in the response we send back?

cjllanwarne (approved):
LGTM, one minor comment about splitting a composite test out into sub- cases and you should be good to go 👍

salonishah11 (approved):

PR #55, opened 2022-10-11
BW-1459: End at timestamp workflow

cjllanwarne:


useful workflows/end_at_timestamp.wdl
```diff
@@ -0,0 +1,77 @@
+version 1.0
+
+workflow end_at_timestamp {
+  input {
+    Int final_timestamp
```
salonishah11, line 5: I think it might be a good idea to add a comment here stating the timestamp is in seconds?

salonishah11 (approved):

useful workflows/end_at_timestamp.wdl
```diff
@@ -0,0 +1,77 @@
...
+    String input3
+    String input4
+    String input5
+    String input6
+    String input7
+    String input8
+
+    Int input9
+    Int input10
+    Int input11
```
kpierre13, line 18: Not a blocker.

Does it mean anything that the double digits' second digit is being displayed as a different color? I'm assuming it's fine.

kpierre13 (approved):

useful workflows/end_at_timestamp.wdl
cjllanwarne: No, we have a custom highlighter for WDL, which apparently is missing a regex that says "the second number is still part of the variable name". But nothing wrong with the file itself, luckily

PR #57, opened 2022-10-13
Adding error_messages to db and limiting char count to 1000

kpierre13:


common/openapi.yml
```diff
@@ -264,6 +264,8 @@ components:
         last_modified_timestamp:
           type: string
           format: date-time
+        error_messages:
+          type: string
```
cjllanwarne, line 268: Since we're returning nulls, you might want/need to mark this as either `optional: true` or `required: false` (I'm not sure which!)... or optional might be the default if we don't specify anything?

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -41,7 +41,8 @@ private RunLog runToRunLog(Run run) {
         .state(CbasRunStatus.toCbasApiState(run.status()))
         .workflowParams(run.runSet().method().inputDefinition())
         .submissionDate(convertToDate(run.submissionTimestamp()))
-        .lastModifiedTimestamp(convertToDate(run.lastModifiedTimestamp()));
+        .lastModifiedTimestamp(convertToDate(run.lastModifiedTimestamp()))
+        .errorMessages(null);
```
cjllanwarne, line 45: You're already adding this to the `Run` class and fetching from the database in `RunDao`... why not pass it through here (eg `.errorMessages(run.errorMessages())`?)

service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -64,6 +64,12 @@ public int updateLastPolledTimestamp(UUID runID) {
             Map.of("id", runID, "lastPolledTimestamp", OffsetDateTime.now())));
   }
 
+  //  public int updateErrorMessages(UUID runId, String error) {
+  //    String sql = "UPDATE run SET error_messages = :errorMessages WHERE id = :id";
+  //    return jdbcTemplate.update(
+  //        sql, new MapSqlParameterSource(Map.of("id", runId, "errorMessages", error)));
+  //  }
+
```
cjllanwarne, line 72: Remove commented code

cjllanwarne: (Or if it's good, uncomment it!)

service/src/main/resources/changelog/changesets/20220823_run.yaml
```diff
@@ -49,3 +49,9 @@ databaseChangeLog:
                   type: timestamp
                   constraints:
                     nullable: false
+              - column:
+                  name: error_messages
+                  type: varchar(1000)
+                  constraints:
+                    nullable: true
```
cjllanwarne, line 56: Fine for now... at some point we will need to get into the habit of making changesets instead of "changing history" for these database updates... but it's fine for now.

kpierre13: Does that mean, automatic vs manual?

common/openapi.yml
kpierre13: It didn't like/recognize `optional: true` so I opted for `required: false`

cjllanwarne (approved):

salonishah11 (approved):

PR #58, opened 2022-10-13
WM-1434 Store CBAS Processing Errors

kpierre13:


kpierre13:
This PR is branched from PR [57](https://github.com/DataBiosphere/cbas/pull/57), so the only changes that are part of this PR are changes in `RunSetsApiController.java`

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -102,12 +103,30 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
     Map<String, Object> params =
         InputGenerator.buildInputs(request.getWorkflowInputDefinitions(), recordResponse);
 
+    UUID runId = UUID.randomUUID();
+    String dataTableRowId = request.getWdsRecords().getRecordIds().get(0);
+
     // Submit the workflow and get its ID:
     RunId workflowResponse;
+    String errorMessage;
```
cjllanwarne, line 111: Since it's only used inside the `catch` I think you could move the variable declaration down to line 116, ie at the same point it gets a value set.

cjllanwarne (approved):
LGTM.

Quite Warning/Note: I've changed the PR base to your other branch so that the diff is reduced, and when the other one merges this will automatically get set back to be against `main`

salonishah11 (approved):

PR #60, opened 2022-10-18
[WM-1448] Submit N Runs to Cromwell

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/WM-1448


service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -63,13 +73,47 @@ public RunSetsApiController(
...
+        recordIdsWithError.put(recordId, errorMsg);
+      }
+    }
+
+    if (recordIdsWithError.size() > 0) {
+      String errorMsg = "Error while fetching WDS Records for Record ID(s): " + recordIdsWithError;
+      log.warn(errorMsg);
+      return new ResponseEntity<>(
+          new RunSetStateResponse().errors(errorMsg), HttpStatus.BAD_REQUEST);
+    }
```
cjllanwarne, line 113: This feels like a large enough, self-contained, chunk that it could be extracted into a separate method?

cjllanwarne: (selfishly, it would make understanding and therefore reviewing it easier if it were a self contained function, and it would probably make unit testing easier too)

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -63,13 +73,47 @@ public RunSetsApiController(
...
 
-    // Store the method
-    UUID methodId = UUID.randomUUID();
+    // Fetch WDS Records and keep track of errors while retrieving records
+    String recordType = request.getWdsRecords().getRecordType();
+    List<String> recordIds = request.getWdsRecords().getRecordIds();
+    recordRecordsInRequest(recordIds.size());
 
+    ArrayList<RecordResponse> recordResponses = new ArrayList<>();
+    HashMap<String, String> recordIdsWithError = new HashMap<>();
```
cjllanwarne, line 91: Nit: This is more a list of pairs than a map. Or... I also wonder whether reversing the key/value order would let us combine similarly-erroring runs together?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -63,13 +73,47 @@ public RunSetsApiController(
...
+    for (String recordId : recordIds) {
+      try {
+        recordResponses.add(wdsService.getRecord(recordType, recordId));
```
cjllanwarne, line 94: TOL: I suspect doing this as a stream would allow us to potentially do things in parallel rather than iterating one by one. A thought for "if we need to optimize", probably...

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -63,13 +73,47 @@ public RunSetsApiController(
...
+      } catch (ApiException e) {
+        log.warn("Record lookup for Record ID {} failed.", recordId, e);
+        ErrorResponse error = gson.fromJson(e.getResponseBody(), ErrorResponse.class);
```
cjllanwarne, line 97: What happens if this parse from json throws its own exception?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -63,13 +73,47 @@ public RunSetsApiController(
...
+        String errorMsg;
+        if (error == null) {
+          errorMsg = e.getMessage();
+        } else {
+          errorMsg = error.getMessage();
+        }
```
cjllanwarne, line 103: Perhaps this could be extracted into its own "find useful error message" utility function?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -82,107 +126,147 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
-    }
+    // For each Record ID, build workflow inputs and submit the workflow to Cromwell
+    List<RunStateResponse> runStateResponseList =
+        buildInputsAndSubmitRun(request, runSet, recordResponses);
 
-    // Build the inputs set from workflow parameter definitions and the fetched record:
-    Map<String, Object> params =
-        InputGenerator.buildInputs(request.getWorkflowInputDefinitions(), recordResponse);
+    // Figure out how many runs are in Failed state. If all Runs are in an Error state then mark the
+    // Run Set as Failed
```
cjllanwarne, line 145: TOL: This feels more like an activity for a future RunSetSmartPoller

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -82,107 +126,147 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+    RunSetState runSetState;
+    List<RunStateResponse> runsInErrorState =
+        runStateResponseList.stream()
+            .filter(run -> CbasRunStatus.fromValue(run.getState()).equals(SYSTEM_ERROR))
```
cjllanwarne, line 149: There's also the other ERROR type we should probably include too

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -82,107 +126,147 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
-      log.error(
-          "New workflow for record {} submitted to engine as {} but CBAS database INSERT returned '{} rows created'",
-          dataTableRowId,
-          workflowResponse.getRunId(),
-          created);
+      additionalErrorMsg =
+          String.format(
+              "CBAS failed to create new row for Record ID %s in %s state in database. INSERT returned '%s rows created'",
+              recordId, runState, created);
+      log.error(additionalErrorMsg);
```
cjllanwarne, line 223: We probably want to escalate this error condition instead of logging and forgetting?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -82,107 +126,147 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+    ArrayList<RunStateResponse> runStateResponseList = new ArrayList<>();
+
+    for (RecordResponse record : recordResponses) {
+      // Build the inputs set from workflow parameter definitions and the fetched record
+      Map<String, Object> workflowInputs =
+          InputGenerator.buildInputs(request.getWorkflowInputDefinitions(), record);
+
+      // Submit the workflow, get its ID and store the Run to database
+      RunId workflowResponse;
+      UUID runId = UUID.randomUUID();
```
cjllanwarne, line 243: TOL: In some future version, we'll probably want to store this first, separately from trying to actually submit (in case we get restarted an at inopportune time, for example)

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -177,30 +235,80 @@ void tooManyRecordIds() throws Exception {
...
-
-class UnitTestRunSetsApiController {
-  private CbasApiConfiguration config = new CbasApiConfiguration();
-  private RunSetRequest request = new RunSetRequest();
 
   @Test
-  void testRequestValidityFewerThanMax() {
-    config.setRunSetsMaximumRecordIds(2);
-    request.setWdsRecords(new WdsRecordSet().recordIds(Arrays.asList("r1")));
-    assertTrue(RunSetsApiController.requestIsValid(request, config));
```
cjllanwarne, line 190: This seems like maybe another case of an overly zealous rebase? It looks like this test has disappeared?

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -105,24 +114,57 @@ class TestRunSetsApiController {
   @Autowired private ObjectMapper objectMapper;
 
   @Test
-  void chainCallsTogether() throws Exception {
```
cjllanwarne, line 108: There's probably a case to be made for a super simple version of this test (chain calls together) as well as all of these different "should continue even with 1 failed run" variations. Maybe they can all call some "run a runset" base case, with special behavior expectations per test case?

service/src/main/java/bio/terra/cbas/common/MetricsUtil.java
```diff
@@ -26,6 +26,8 @@ private MetricsUtil() {}
   public static final String METRICS_PREFIX = "terra/cbas/";
 
   public static final String OUTBOUND_REQUEST_METRICS = METRICS_PREFIX + "request/outbound";
+  public static final String RUNS_SUBMITTED_SUCCESSFULLY_PER_RUN_SET_METRICS =
```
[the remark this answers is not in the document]
salonishah11: I had meant to put this question for reviewers and forgot about it. I think it would be great to record both success and failure Run submissions. How do I use the tag feature in metrics?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Good point. I think I had trouble separating it out initially but I will try again and see how it goes 👍

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Yeah it can be a list of pairs. I guess I found Map fitting my mental model as each error would be associated with 1 Record Id. 
>I also wonder whether reversing the key/value order would let us combine similarly-erroring runs together

WDS returns an `ErrorResponse` containing the below fields and hence putting that as key to combine similarly-erroring runs might not work because of timestamp field 🤔 
```
 {"timestamp":"2022-10-20T19:45:01.210+00:00","status":404,"error":"Not Found","message":"Instance does not exist","path":"/15f36863-30a5-4cab-91f7-52be439f1175/records/v0.2/FOO/FOO4"}
```

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: I had initially tried to do this as a stream, but since `wdsService.getRecord` throws an exception it was becoming hard to figure it all out inside the map function because it didn't like checked exceptions being thrown inside its body. I am happy to mob on this though.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: That's a good point. I could wrap this in a try and catch. I am doing this because `wdsService.getRecord` throws an `ErrorResponse` with below fields inside the ApiException. So I thought maybe it might be better to only extract the error message and send it to user. Do you think this is a good idea?
Example of exception thrown by WDS:
```
org.databiosphere.workspacedata.client.ApiException: {"timestamp":"2022-10-20T19:45:01.210+00:00","status":404,"error":"Not Found","message":"Instance does not exist","path":"/15f36863-30a5-4cab-91f7-52be439f1175/records/v0.2/FOO/FOO4"}
```

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Yeah good point. That way adding a try catch inside method would be cleaner than adding it here.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: I guess the idea behind this here is that if CBAS was unable to / failed to even submit Runs to Cromwell, we mark the RunSet as Errored instead of Running. My understanding was that the future RunSetSmartPoller would do this while monitoring Runs in the set, after the Runs were submitted to Cromwell. I guess I found it weird in the API response of POST /run_sets that the Runs had failed but Run Set state was Running and hence I added this. If you think it is overkill to add it now I can remove it.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Good point. I can add that 👍

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Yes. Hence I am storing it in [this variable here](https://github.com/DataBiosphere/cbas/blob/1a4fc18d77fc47a34cf5b66f6993f5c90e09bc18/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java#L219) and adding it to the [Run's error field here](https://github.com/DataBiosphere/cbas/blob/1a4fc18d77fc47a34cf5b66f6993f5c90e09bc18/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java#L229).

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
salonishah11: Since I added back return error from `RunSetsApiController.requestIsValid` I had to refactor these tests too. They are kind of merged into these [2 tests here](https://github.com/DataBiosphere/cbas/blob/1a4fc18d77fc47a34cf5b66f6993f5c90e09bc18/service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java#L279-L313). If the outcome of [this comment discussion](https://github.com/DataBiosphere/cbas/pull/60#discussion_r999949144) is to not return the error message to user yet I will revert my changes for the test as well.

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
salonishah11: I am not sure how to actually have a "run a runset" base case without putting it in each test 🤔 Should I refactor the mock code out of each tests?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -82,107 +126,147 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+      errorMsg +=
+          String.format("Duplicate Record ID(s) %s present in request.", duplicateRecordIds);
+    }
+
+    if (!errorMsg.isEmpty()) {
+      String finalErrorMsg = "Bad user request. Error(s): " + errorMsg;
+      log.warn(finalErrorMsg);
+      return Optional.of(
+          new ResponseEntity<>(
+              new RunSetStateResponse().errors(finalErrorMsg), HttpStatus.BAD_REQUEST));
```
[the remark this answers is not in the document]
salonishah11: After talking with Chris and Michael, we decided to return a `List<String>` to return the errors instead of `Optional<ResponseEntity<RunSetStateResponse>>` so that it is an addition of previous method rather than removal.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Added the other error state in this check.

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
salonishah11: These tests have now been added back.

kpierre13 (approved):

cjllanwarne (approved):

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -73,40 +76,21 @@ public RunSetsApiController(
...
-    Gson gson = new Gson();
-
     // request validation
-    Optional<ResponseEntity<RunSetStateResponse>> validateRequestResponse =
-        validateRequest(request, this.cbasApiConfiguration.getRunSetsMaximumRecordIds());
-    if (validateRequestResponse.isPresent()) {
-      return validateRequestResponse.get();
+    List<String> requestErrors = validateRequest(request, this.cbasApiConfiguration);
+    if (!requestErrors.isEmpty()) {
+      String errorMsg = "Bad user request. Error(s): " + requestErrors;
```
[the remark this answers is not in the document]
salonishah11: So the `errors` field even though plural is of type `String` which now I realize could be a misleading name. Do you suggest changing data type to List or change the name from `errors` to `errorMessage`?

cjllanwarne (approved):

PR #61, opened 2022-10-24
WM-1439 WDS Errors-out

kpierre13:


cjllanwarne:
sonarcloud has found 2 “code smells”, which do look like they're worth doing in this case

service/src/test/java/bio/terra/cbas/runsets/inputs/TestInputGeneratorBuildInputs.java
```diff
@@ -17,68 +21,68 @@ class TestInputGeneratorBuildInputs {
       new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 
   @Test
-  void stringLiteral() throws JsonProcessingException {
+  void stringLiteral() throws Exception {
```
cjllanwarne, line 24: It's a test, so this isn't the end of the world, but if we **_could_** be more specific about what type(s) of exception this might throw, that would still be nice, IMO

service/src/test/java/bio/terra/cbas/runsets/inputs/TestInputGeneratorBuildInputs.java
```diff
@@ -169,4 +173,56 @@ private static RecordResponse fooRatingRecord(String rawAttributeJson)
...
+
+    try {
+      InputGenerator.buildInputs(inputDefinitions, recordResponse);
+    } catch (Exception e) {
+      exception = e;
+    }
+
+    assertNotNull(exception);
+    assertTrue(exception instanceof WorkflowAttributesNotFoundException);
+    assertEquals(exception.getMessage(), "Attribute not found in WDS record.");
```
cjllanwarne: This might look nicer as an `assertThrows` (something like in [this answer](https://stackoverflow.com/questions/40268446/junit-5-how-to-assert-an-exception-is-thrown))

service/src/test/java/bio/terra/cbas/runsets/inputs/TestInputGeneratorBuildInputs.java
```diff
@@ -169,4 +173,56 @@ private static RecordResponse fooRatingRecord(String rawAttributeJson)
...
+            """
+        {
+          "id": "FOO1",
+          "type": "FOO",
+          "attributes": {
+            "MY_RECORD_ATTRIBUTE": "Hello, world"
+          },
+          "metadata": {
+            "provenance": "TODO: RECORDMETADATA"
+          }
```
cjllanwarne, line 206: Try rebasing again, because I think this is now going to need to be removed, due to WDS changing.

service/src/test/java/bio/terra/cbas/runsets/inputs/TestInputGeneratorBuildInputs.java
kpierre13: I agree. When the exception thrown is `JsonProcessingException` I get an error `Unhandled exception: bio.terra.cbas.common.exceptions.WorkflowAttributesNotFoundException`. Which is why I then changed it to the more general `Exception`. Is there a way to incorporate both?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -267,6 +270,11 @@ private List<RunStateResponse> buildInputsAndSubmitRun(
         log.warn(errorMsg, e);
         runStateResponseList.add(
             storeRun(runId, null, runSet, record.getId(), SYSTEM_ERROR, errorMsg + e.getMessage()));
+      } catch (WorkflowAttributesNotFoundException e) {
+        String errorMsg = "Attribute was not found in WDS record";
+        log.warn(errorMsg, e);
+        runStateResponseList.add(
+            storeRun(runId, null, runSet, record.getId(), SYSTEM_ERROR, e.getMessage()));
```
mspector, line 277: do we want to prepend `errorMsg` to `e.getMessage()` like we do in other catch blocks?

service/src/test/java/bio/terra/cbas/runsets/inputs/TestInputGeneratorBuildInputs.java
mspector: a quick search online suggests you could do something like:

```
void stringLiteral() throws JsonProcessingException, WorkflowAttributesNotFoundException {
    ...
}
```

https://stackoverflow.com/questions/2912565/throwing-multiple-exceptions-in-java

service/src/test/java/bio/terra/cbas/runsets/inputs/TestInputGeneratorBuildInputs.java
mspector: does that make this a good candidate for contract tests in the future?

I'm not suggesting Katrina actually try to do it in this review cycle, just thinking ahead

service/src/main/java/bio/terra/cbas/runsets/inputs/InputGenerator.java
```diff
@@ -34,7 +36,12 @@ public static Map<String, Object> buildInputs(
       } else {
         String attributeName =
             ((ParameterDefinitionRecordLookup) param.getSource()).getRecordAttribute();
-        parameterValue = record.getAttributes().get(attributeName);
+
+        if (!((Map<String, Object>) recordResponse.getAttributes()).containsKey(attributeName)) {
```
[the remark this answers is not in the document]
kpierre13: The short answer: A HashMap.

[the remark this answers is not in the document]
kpierre13: The short answer: A HashMap.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
kpierre13: This what I had initially, but that caused the two of the same error message to be printed in the DB. So in this case, I just opted for `e.getMessage`. If you think I should re-add `errorMsg` I'd be happy to.

service/src/test/java/bio/terra/cbas/runsets/inputs/TestInputGeneratorBuildInputs.java
kpierre13: I did this as well, but that caused errors in github tests, causing all of these to fail. Maybe I can add something like an `assert` to the end of the test to work around this?

service/src/test/java/bio/terra/cbas/runsets/inputs/TestInputGeneratorBuildInputs.java
cjllanwarne: it absolutely does!

service/src/main/java/bio/terra/cbas/runsets/inputs/InputGenerator.java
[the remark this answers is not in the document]
cjllanwarne: It's casting a lowest-common-denominator type (`recordResponse.getAttributes()` is just an `Object`) to something more specific, so that it has a `containsKey` method to call

service/src/main/java/bio/terra/cbas/common/exceptions/WorkflowAttributesNotFoundException.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.cbas.common.exceptions;
+
+public class WorkflowAttributesNotFoundException extends Exception {
+
+  public WorkflowAttributesNotFoundException(String message) {
+    super(message);
+  }
```
cjllanwarne, line 7: Today this message is always the same, and kind of just duplicates the name of the exception. Maybe we could template the message inside the class, and the caller could give us some useful debugging information on what was missing? eg something like
 ```suggestion
   public WorkflowAttributesNotFoundException(String attribute, String recordId, String inputName) {
     super("Attribute %s not found in WDS record %s (to populate workflow input %s).".formatted(attribute, recordId, inputName));
   }
```

cjllanwarne (approved):
Looks good as is, one optional improvement to think about

service/src/test/java/bio/terra/cbas/runsets/inputs/TestInputGeneratorBuildInputs.java
cjllanwarne: So
1. This was a low priority comment from me, feel free to ignore it if it's the only thing stopping you from merging.
2. It might be a useful learning experience to work out what the problem is, since exception wrangling is (unfortunately) quite a large part of Java programming... so we can mob if you're curious, or you can ignore this if you just want it to be done (and we can always revisit it later!)

cjllanwarne:
I think my changes in #66 may have broken this PR's assumption that "missing values are always bad", but I have a pretty good idea how to resolve that so we can pair on getting that part working again if you start getting weird errors.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
mspector: ah, got it. no need in that case

cjllanwarne (approved):

mspector (approved):
nice :)

PR #66, opened 2022-10-27
WM-1473 Implement optional input and output behavior

cjllanwarne:
Adds wiring and logic for optional types.

Note: Relies on behavior that is fixed in #61 (which fixes the bad assumption that _every_ input is optional 🤦). Whichever PR merges second will need to restore "optional inputs really are optional after all" behavior.


kpierre13 (approved):

service/src/test/java/bio/terra/cbas/runsets/outputs/TestOutputGeneratorPrimitives.java
```diff
@@ -154,6 +110,6 @@ void invalidOutputName() throws Exception {
 
     assertNotNull(exception);
     assertTrue(exception instanceof WorkflowOutputNotFoundException);
-    assertEquals(exception.getMessage(), "Output myWorkflow.naem not found in workflow outputs.");
+    assertEquals("Output myWorkflow.naem not found in workflow outputs.", exception.getMessage());
```
mspector, line 113: also here

mspector (approved):
found some small typos in the test cases but they don't look like they actually matter, from a correctness point of view

PR #67, opened 2022-10-31
[WM-1489] Use PATCH endpoint to update attributes in WDS

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/WM-1489

### DIFF
--- a/service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
+++ b/service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
@@ -47,12 +47,14 @@ public class SmartRunsPoller {
     this.objectMapper = objectMapper;
   }
 
-  public RecordResponse updateOutputAttributes(Run run) throws Exception {
-    List<WorkflowOutputDefinition> outputDefinitionList =
-        objectMapper.readValue(run.runSet().method().outputDefinition(), new TypeReference<>() {});
+  public RecordResponse updateOutputAttributes(
+      Run run, List<WorkflowOutputDefinition> outputDefinitionList) throws Exception {
     Object outputs = cromwellService.getOutputs(run.engineId());
     RecordAttributes outputParamDef = OutputGenerator.buildOutputs(outputDefinitionList, outputs);
     RecordRequest request = new RecordRequest().attributes(outputParamDef);
+
+    logger.info("Updating output attributes for Record ID {} from Run {}", run.recordId(), run.engineId());
+
     return wdsService.updateRecord(request, run.runSet().method().recordType(), run.recordId());
   }
 
@@ -122,7 +124,12 @@ public class SmartRunsPoller {
       if (r.status() != updatedRunState) {
         if (updatedRunState == CbasRunStatus.COMPLETE) {
           try {
-            updateOutputAttributes(r);
+            List<WorkflowOutputDefinition> outputDefinitionList =
+                objectMapper.readValue(
+                    r.runSet().method().outputDefinition(), new TypeReference<>() {});
+            if (!outputDefinitionList.isEmpty()) {
+              updateOutputAttributes(r, outputDefinitionList);
+            }
           } catch (Exception e) {
             // log error and mark Run as Failed
             // TODO: When epic WM-1433 is being worked on, add error message in database stating
--- a/service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
+++ b/service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
@@ -301,4 +301,54 @@ public class TestSmartRunsPoller {
         COMPLETE,
         actual.stream().filter(r -> r.id().equals(completedRunId)).toList().get(0).status());
   }
+
+  @Test
+  void shouldNotCallWDSIfOutputsEmpty() throws Exception {
+    CromwellService cromwellService = mock(CromwellService.class);
+    RunDao runsDao = mock(RunDao.class);
+    WdsService wdsService = mock(WdsService.class);
+    SmartRunsPoller smartRunsPoller =
+        new SmartRunsPoller(cromwellService, runsDao, wdsService, objectMapper);
+
+    String outputDefinition = "[]";
+    RunSet runSet =
+        new RunSet(
+            UUID.randomUUID(),
+            new Method(
+                UUID.randomUUID(), "methodurl", "inputdefinition", outputDefinition, "entitytype"));
+    Run runToUpdate1 =
+        new Run(
+            runningRunId1,
+            runningRunEngineId1,
+            runSet,
+            runningRunEntityId1,
+            runSubmittedTime,
+            RUNNING,
+            runningRunStatusUpdateTime,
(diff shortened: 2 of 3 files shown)


service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -47,12 +47,15 @@ public SmartRunsPoller(
...
-    List<WorkflowOutputDefinition> outputDefinitionList =
-        objectMapper.readValue(run.runSet().method().outputDefinition(), new TypeReference<>() {});
+  public RecordResponse updateOutputAttributes(
+      Run run, List<WorkflowOutputDefinition> outputDefinitionList) throws Exception {
     Object outputs = cromwellService.getOutputs(run.engineId());
     RecordAttributes outputParamDef = OutputGenerator.buildOutputs(outputDefinitionList, outputs);
     RecordRequest request = new RecordRequest().attributes(outputParamDef);
+
+    logger.info(
+        "Updating output attributes for Record ID {} from Run {}", run.recordId(), run.engineId());
```
cjllanwarne, line 57: This feels more like debug than info. And TBH I'd probably avoid putting confidential user values in logs at all, if we can avoid it.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -122,7 +125,12 @@ private void updateDatabaseRunStatus(
       if (r.status() != updatedRunState) {
         if (updatedRunState == CbasRunStatus.COMPLETE) {
           try {
-            updateOutputAttributes(r);
+            List<WorkflowOutputDefinition> outputDefinitionList =
+                objectMapper.readValue(
+                    r.runSet().method().outputDefinition(), new TypeReference<>() {});
+            if (!outputDefinitionList.isEmpty()) {
+              updateOutputAttributes(r, outputDefinitionList);
+            }
```
cjllanwarne, line 133: This presumably is here to avoid empty PATCHes if there were no output definitions? Maybe we could add a comment for that?

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
```diff
@@ -301,4 +301,54 @@ void updatingOutputFails() throws Exception {
         COMPLETE,
         actual.stream().filter(r -> r.id().equals(completedRunId)).toList().get(0).status());
   }
+
+  @Test
+  void shouldNotCallWDSIfOutputsEmpty() throws Exception {
```
cjllanwarne, line 306: TOL, but I wonder whether we could extract some smaller "update WDS" method, so that this test could just target that one piece of functionality, without having to set up a whole complex API test just to check a boolean decision half way through?

cjllanwarne (commented):
Looks good. I had two extra thoughts -

1. I was a surprised there were no tests already asserting how the WDS update method got used, and needed changing.
2. The two major sonarcloud "code smells" seem legit. The minor ones... tbh I kind of disagree with so feel free to ignore those AFAIC!

mspector (approved):
I don't have more to add beyond what Chris has said. LGTM once those get addressed (approving, but if you want me to look again after you've made changes, let me know)

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
salonishah11: I can change this to debug level. As we discussed in standup, this is logging only the Record ID and Run ID hence I will leave it as it is.

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
salonishah11: I can try extracting out the function 👍

kpierre13 (approved):

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
salonishah11: I extracted the "should it call WDS to update attributes" logic into a separate function that simplified the tests.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -47,16 +48,31 @@ public SmartRunsPoller(
...
     RecordRequest request = new RecordRequest().attributes(outputParamDef);
 
-    logger.info(
-        "Updating output attributes for Record ID {} from Run {}", run.recordId(), run.engineId());
+    if (logger.isDebugEnabled()) {
+      logger.debug(
+          "Updating output attributes for Record ID {} from Run {}.",
+          run.recordId(),
+          run.engineId());
+    }
```
cjllanwarne, line 73: Sorry for flip flopping! Since these are just IDs, I think it's fine at `info` level. Also, the main reason for the `isDebugEnabled` is to avoid formatting strings that never get printed... but this debug method probably doesn't format the string if that isn't needed anyway, so it might be less necessary here...

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
```diff
@@ -60,6 +63,12 @@ public class TestSmartRunsPoller {
           .setDateFormat(new StdDateFormat())
           .setDefaultPropertyInclusion(JsonInclude.Include.NON_ABSENT);
 
+  private CromwellService cromwellService = mock(CromwellService.class);
+  private RunDao runsDao = mock(RunDao.class);
+  private WdsService wdsService = mock(WdsService.class);
+  private SmartRunsPoller smartRunsPoller =
+      new SmartRunsPoller(cromwellService, runsDao, wdsService, objectMapper);
+
```
cjllanwarne, line 71: The problem with doing this is that now the exact same mock objects are shared between all the tests, meaning that any mock configuration is shared, and any "did x get called" checks might be picking up method calls from previously run tests.

My suggestion would be to make these global variables with default values (ie, set them all to `null`, and then use a `before` hook to create the instances for each of them afresh for each test case.

cjllanwarne (approved):
LGTM but with a couple of TOLs

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
salonishah11: I will change it to info level. The reason I added `logger.isDebugEnabled()` was because it was giving me a code smell saying it will affect performance. Maybe there is another way to avoid code smell 🤔

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPoller.java
salonishah11: Ah I see. Yeah I can implement your suggestion 👍

PR #68, opened 2022-11-01
BW-1474: Add File types as inputs

cjllanwarne:
To aid validation, includes type and value coercion from Json/WDS types to Cbas recognized types, and on to Cromwell inputs.

From the ticket.

AC:

- File inputs must be declarable via our input definitions.
- Input files are correctly passed along to Cromwell if used in input definitions
- Basic string validation of the file path to make sure it looks like a file
  - (put pieces in place in case we need more validation later)

Tests:

- Unit tests against the inputs generator

Metrics:

- Count of files that fail validation and those that don’t
- Counts of GCS, Azure, local, DRS file paths

### DIFF
--- a/service/src/main/java/bio/terra/cbas/runsets/types/CbasBoolean.java
+++ b/service/src/main/java/bio/terra/cbas/runsets/types/CbasBoolean.java
@@ -1,0 +1,25 @@
+package bio.terra.cbas.runsets.types;
+
+public class CbasBoolean implements CbasValue {
+  private final Boolean value;
+
+  /*
+  Private constructor. Use 'parse' instead.
+   */
+  private CbasBoolean(Boolean value) {
+    this.value = value;
+  }
+
+  @Override
+  public Object asCromwellInput() {
+    return value;
+  }
+
+  public static CbasBoolean parse(Object value) throws CoercionException {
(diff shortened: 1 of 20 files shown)


service/src/main/java/bio/terra/cbas/runsets/types/CbasBoolean.java
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  public static CbasBoolean parse(Object value) throws CoercionException {
+    if (value instanceof Boolean b) {
+      return new CbasBoolean(b);
+    } else {
+      throw new TypeCoercionException(value, "Int");
+    }
+  }
+}
```
kpierre13, line 25: [Not a Blocker] Why do we need different files for all of these different input types?

kpierre13 (approved):

service/src/main/java/bio/terra/cbas/runsets/types/CbasBoolean.java
cjllanwarne: You can thank Java... it needs every class to be in its own appropriately-named file. Or are you asking why we want separate java classes for every type?

PR #71, opened 2022-11-02
[WM-1498] limit workflow inputs outputs

mspector:


service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -77,6 +79,9 @@ public RunSetsApiController(
 
   @Override
   public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
+    // request metrics (before validation)
+    captureRequestMetrics(request);
```
mspector, line 83: what do we think about this being (1) it's own function, and (2) before any validation happens?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -147,8 +152,21 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+  }
+
   public static List<String> validateRequest(RunSetRequest request, CbasApiConfiguration config) {
     List<String> errorList = new ArrayList<>();
+    errorList.addAll(validateRequestRecordIds(request, config));
+    errorList.addAll(validateRequestInputsAndOutputs(request, config));
+    return errorList;
+  }
+
+  public static List<String> validateRequestRecordIds(
```
mspector, line 167: splitting off different functions for testing different parts of the request seemed like a good way of making the unit tests simpler

service/src/main/resources/application.yml
```diff
@@ -41,6 +41,8 @@ cbas:
   cbas-api:
     # for POST/run_sets, the number of record IDs must be less than or equal to this value.
     runSetsMaximumRecordIds: 100
+    maxWorkflowInputs: 25
+    maxWorkflowOutputs: 25
```
mspector, line 45: I haven't done the research yet on what these values should be

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: I think it makes sense to keep it as a separate function. For (2) I guess the benefit of capturing here is that we will be able to see how many requests are there with number of inputs and/or outputs higher than our max allowed. So it does make sense to have it here.

salonishah11 (approved):

cjllanwarne (approved):

PR #72, opened 2022-11-04
WM-1511, WM-1512: Array types in inputs

cjllanwarne:
AC in ticket:

> - It is possible to provide a literal array value, or a WDS lookup array value in the input definition.
> - It should be possible to mark arrays as non-empty, in which case CBAS would validate that the array is not empty before sending it.
> 
> Test:
> 
> - Unit tests validating that input definition JSON can be correctly interpreted for array inputs.
> - Unit tests of coercion from WDS values into arrays
> - Unit tests of writing CbasArrays into cromwell inputs
> - Unit tests validating non-emptiness checks.

### DIFF
--- a/service/src/main/java/bio/terra/cbas/runsets/types/CbasArray.java
+++ b/service/src/main/java/bio/terra/cbas/runsets/types/CbasArray.java
@@ -22,13 +22,20 @@ public class CbasArray implements CbasOptional {
     return values.stream().map(CbasValue::countFiles).reduce(0L, Long::sum);
   }
 
-  public static CbasArray parseValue(ParameterTypeDefinition innerType, Object values)
-      throws CoercionException {
+  public static CbasArray parseValue(
+      ParameterTypeDefinition innerType, Object values, boolean nonEmpty) throws CoercionException {
     if (values instanceof List valueList) {
       List<CbasValue> coercedValues = new ArrayList<>();
       for (Object value : valueList) {
         coercedValues.add(CbasValue.parseValue(innerType, value));
       }
+      if (nonEmpty && coercedValues.isEmpty()) {
+        throw new ValueCoercionException(
+            values,
+            "Array[%s]".formatted(innerType),
+            "Non-empty array must have at least one value.");
+      }
+
       return new CbasArray(coercedValues);
     } else {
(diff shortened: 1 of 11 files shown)


service/src/test/java/bio/terra/cbas/runsets/outputs/StockOutputDefinitions.java
```diff
@@ -45,4 +45,21 @@ public static WorkflowOutputDefinition optionalOutputDefinition(
 
     return objectMapper.readValue(rawOutputDefinition, new TypeReference<>() {});
   }
+
+  public static WorkflowOutputDefinition arrayOutputDefinition(
```
salonishah11, line 49: Would this belong in https://broadworkbench.atlassian.net/browse/WM-1512?

service/src/main/java/bio/terra/cbas/runsets/types/CbasArray.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @Override
+  public Object asSerializableValue() {
+    return values.stream().map(CbasValue::asSerializableValue).toList();
+  }
+
+  @Override
+  public long countFiles() {
+    return values.stream().map(CbasValue::countFiles).reduce(0L, Long::sum);
```
pshapiro4broad, line 22: Java's stream APIs have support for numeric types that provides unboxed access. It's more efficient and natively supports things like `sum`.
```suggestion
    return values.stream().mapToLong(CbasValue::countFiles).sum();
```

service/src/test/java/bio/terra/cbas/runsets/outputs/StockOutputDefinitions.java
cjllanwarne: Yeah, I ended up bundling them both into this PR since there was no additional coding needed (just test writing)

service/src/main/java/bio/terra/cbas/runsets/types/CbasArray.java
cjllanwarne: Very cool!

salonishah11 (approved):

kpierre13 (approved):

PR #73, opened 2022-11-07
[WM-1502] test target workflow 2

mspector:


useful workflows/target_workflow_2/target_workflow_2-test.wdl
```diff
@@ -0,0 +1,413 @@
...
+        Int                         output_int_1            = input_int_default
+        Int                         output_int_2            = input_int_default
+        Int                         output_int_3            = input_int_default
+        Float                       output_float_1          = input_float_default
+        Int                         output_int_4            = input_int_default
+        Int                         output_int_5            = input_int_default
+        
+        Array[Int]                  output_int_array_1      = [input_int_default]
+        Array[Float]                output_float_array      = [input_float_default]
+        Array[Map[String,String]]   array_of_maps_of_str    = [{input_file: input_string_optional}]
```
cjllanwarne, line 368: For the sake of testing our handling, it might be nice to have some variety in size for these maps and arrays (but think 0, 1, 2 rather than 100, 10000, 1000000...)

useful workflows/target_workflow_2/target_workflow_2-test.wdl
```diff
@@ -0,0 +1,413 @@
...
+    Int    output_int_1   = 1
+    Int    output_int_2   = 2
+    Int    output_int_3   = 3
+    Float  output_float_1 = 4.0
+    Float  output_float_2 = 5.0
+    String output_string  = "hello world"
+  }
+}
+
+task refine_assembly_with_aligned_reads_mock {
```
cjllanwarne, line 219: TOL it might be better to replace the names with something generic too

useful workflows/target_workflow_2/target_workflow_2-test.wdl
```diff
@@ -0,0 +1,413 @@
...
+    File     input_file_1
+    File     input_file_2
+    String   input_string
+
+    Boolean  input_boolean_default = false
+    Float    input_float_default = 0.5
+    Int      input_int_default = 3
+
+    Int?     input_int_optional
+    String   docker = "quay.io/broadinstitute/viral-assemble:2.1.16.1"
```
cjllanwarne, line 231: I was about to say "let's use tiny docker images rather than full ones... but use of `quay` might be a nice test case once this hits the "can Cromwell run it" stage...

useful workflows/target_workflow_2/target_workflow_2-test.wdl
```diff
@@ -0,0 +1,413 @@
...
+    Int?   input_int_optional_3
+    
+    Int?   input_int_optional_4
+    String input_string_default = "ivar trim mock string"
+  }
+
+  Float variable_float = 0.82
+
+  command {
+    echo "ivar_trim_mock"
```
cjllanwarne, line 71: It might be nice to produce a "digest" of the inputs as an output, to prove that the files are indeed actually making it to the command. I think target workflow 1's command block does something like that

useful workflows/target_workflow_2/target_workflow_2-test.wdl
```diff
@@ -0,0 +1,413 @@
...
+  }
+
+  output {
+    File   output_file   = input_file
```
cjllanwarne, line 75: It might be best to use files produced IN the task rather than mirroring inputs. 

1. Tests that Cromwell is localizing and delocalizing properly
2. Cromwell is a little strange about mirrored input files. I have a feeling it treats them differently than generated files (eg it just reuses the old path that came in as an input rather than generating a new path). Not sure I remember the exact problem but I do have a vague sense of "here be dragons"

useful workflows/target_workflow_2/target_workflow_2-test.wdl
mspector: interestingly, this was leftover, and isn't being used anywhere. I've deleted all the `runtime` blocks. Is that a mistake? Should I put them back in?

cjllanwarne: Yes, we'll need to have at least a docker image declared in every task for it to be able to run in CaaA

cjllanwarne (approved):

cjllanwarne (approved):

salonishah11 (commented):
Would it also make sense to add input and output definitions as part of this ticket?

cjllanwarne (approved):

cjllanwarne:
LGTM, +1 to Saloni's suggestion as either part of this PR, as a follow up, or as part of "test run target workflow 2"

useful workflows/target_workflow_2/target_workflow_2-test.wdl
```diff
@@ -0,0 +1,614 @@
...
+    String   input_string_default_1 = "this is a default string"
+    String?  input_string_optional
+    Boolean? input_bool_default = false
+    Int?     input_int_optional
+    String   input_string_default_2 = "my-docker-image.123"
+    String   input_string_default_3 = "this is a second default string"
+  }
+
+  Float declared_float = 0.67
+  Int declared_int = 5
```
kpierre13, line 21: [not a blocker] is this `declared_int` something like disk space size? Is it necessary for the workflow since it's used in the outputs?

kpierre13 (approved):

salonishah11 (approved):

useful workflows/target_workflow_2/target_workflow_2-test.wdl
mspector: ah, if it's not being used in the outputs, it could have been something like "disk space size" in the original task, that just didn't get mapped to anything when I cut away the unnecessary parts...

PR #78, opened 2022-11-15
[WM-1550] womtool to cbas script

mspector:


kpierre13 (approved):

useful workflows/womtool-to-cbas.py
```diff
@@ -0,0 +1,113 @@
+'''
+A script for converting the output of womtool's `inputs` command to a CBAS input definitions file.
+
+This script was written (quickly) to accelerate the test and development of CBAS features,
+and should not be considered to be actively maintained. Use at your own risk.
+
```
salonishah11, line 6: I think it would be great to add a comment on how to use this script as well as an example (like sample WomTool inputs and what it's CBAS input will look like through the script)

cjllanwarne (approved):
Didn't test myself but I know it's proven itself useful already, and isn't production code so happy to merge quickly and often to aid quick collaboration

useful workflows/womtool-to-cbas.py
mspector: done!

PR #88, opened 2022-11-22
[WM-1461] Store timezone information in timestamps in db and generate time in UTC

salonishah11:
The timestamps in database are stored with timezone information. The timestamps are passed to database in UTC time, and postgres for `timestamp with time zone` data type stores the timestamps internally in UTC format but when displayed converts the timestamp to local time ([reference](https://www.postgresql.org/docs/current/datatype-datetime.html#AEN5811)) and hence in database terminal it is displayed with offset. API responses have timestamps in UTC time and as a result the UI is able to correctly display timestamp based on the browser's timezone.

Previously
Submission created: ~ 5:09 PM EST
Value stored in DB: 2022-09-21 17:09:18.911989
Value returned in API response: 2022-09-21T17:09:18.911+00:00
Value shown in UI: 1:09 pm

Now,
Submission created: ~ 2:54PM EST
Value passed to DB: 2022-11-23T19:54:35.390733Z
Value stored in DB: 2022-11-23 14:54:35.390733-05
Value returned in API response (UTC timezone): 2022-11-23T19:54:35.390+00:00
Value shown in UI: 2:54 pm

Screen Shot 2022-11-23 at 3 03 11 PM

Closes https://broadworkbench.atlassian.net/browse/WM-1461

### DIFF
--- a/service/src/main/java/bio/terra/cbas/dao/RunDao.java
+++ b/service/src/main/java/bio/terra/cbas/dao/RunDao.java
@@ -50,7 +50,7 @@ public class RunDao {
   public int updateRunStatus(UUID runId, CbasRunStatus newStatus) {
     OffsetDateTime currentTimestamp = OffsetDateTime.now();
     String sql =
-        "UPDATE run SET status = :status, last_modified_timestamp =:lastModifiedTimestamp, last_polled_timestamp = :lastPolledTimestamp  WHERE id = :id";
+        "UPDATE run SET status = :status, last_modified_timestamp =:last_modified_timestamp, last_polled_timestamp = :last_polled_timestamp  WHERE id = :id";
     return jdbcTemplate.update(
         sql,
         new MapSqlParameterSource(
@@ -66,7 +66,7 @@ public class RunDao {
   }
 
   public int updateLastPolledTimestamp(UUID runID) {
-    String sql = "UPDATE run SET last_polled_timestamp = :lastPolledTimestamp  WHERE id = :id";
+    String sql = "UPDATE run SET last_polled_timestamp = :last_polled_timestamp  WHERE id = :id";
     return jdbcTemplate.update(
         sql,
         new MapSqlParameterSource(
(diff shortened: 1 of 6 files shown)


kpierre13 (approved):

aednichols (changes requested):
Wouldn't it be simpler to store UTC (with no TZ information) in the database and then convert to local time in the presentation layer only? In other words, all timezone info sent via JSON is UTC, and the browser converts to & from local.

service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -48,9 +49,9 @@ public List<Run> getRuns(String runSetId) {
   }
 
   public int updateRunStatus(UUID runId, CbasRunStatus newStatus) {
-    OffsetDateTime currentTimestamp = OffsetDateTime.now();
+    OffsetDateTime currentTimestamp = DateUtils.currentTimeInUTC();
     String sql =
-        "UPDATE run SET status = :status, last_modified_timestamp =:lastModifiedTimestamp, last_polled_timestamp = :lastPolledTimestamp  WHERE id = :id";
+        "UPDATE run SET status = :status, last_modified_timestamp = :last_modified_timestamp, last_polled_timestamp = :last_polled_timestamp WHERE id = :id";
```
salonishah11, line 54: This is a fix up from previous PR

service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -89,7 +91,7 @@ public Run mapRow(ResultSet rs, int rowNum) throws SQLException {
               rs.getString(Method.METHOD_URL_COL),
               rs.getString(Method.INPUT_DEFINITION_COL),
               rs.getString(Method.OUTPUT_DEFINITION_COL),
-              rs.getString(Method.OUTPUT_DEFINITION_COL));
+              rs.getString(Method.RECORD_TYPE_COL));
```
salonishah11, line 94: This is a fix up from previous PR

aednichols (approved):

mspector (approved):

PR #93, opened 2022-12-13
WM-1534: Metrics and lastPolled timestamps

cjllanwarne:
Adds metrics for WM-1534 and makes sure the last polled timestamps are up to date, even if nothing changes

### DIFF
--- a/service/src/main/java/bio/terra/cbas/dao/util/SqlPlaceholderMapping.java
+++ b/service/src/main/java/bio/terra/cbas/dao/util/SqlPlaceholderMapping.java
@@ -3,14 +3,12 @@ package bio.terra.cbas.dao.util;
 import bio.terra.cbas.util.Pair;
 import com.google.common.collect.Streams;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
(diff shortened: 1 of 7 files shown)


service/src/main/java/bio/terra/cbas/dao/util/SqlPlaceholderMapping.java
```diff
@@ -0,0 +1,44 @@
...
+
+import bio.terra.cbas.util.Pair;
+import com.google.common.collect.Streams;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public final class SqlPlaceholderMapping<A> {
```
mspector, line 11: curious to learn more about this -- is this your solution for "a neater way to build SQL queries"?

mspector (approved):

service/src/main/java/bio/terra/cbas/dao/util/SqlPlaceholderMapping.java
cjllanwarne: It's a small step in that direction, yeah. Still a little more bespoke than I'd like but it seems to be working for now, and definitely simplified the in-DAO logic, which was nice

kpierre13 (approved):

PR #95, opened 2022-12-22
WX-864 Enable workflow logs written to storage

kraefrei:
To enable workflow logs to be written to a storage container, the workflow must have a options.json that includes the option "final_workflow_log_dir". This change adds a CBAS config specifying a log directory to write to, and passes this option through the WES API. This is intended to mirror how rawls supplies this option in GCP. 

An outstanding change is required to make WES use this value for workflowOptions on submission.

There is an accompanying helm change to supply this configuration [here](https://github.com/broadinstitute/cromwhelm/pull/153)

### DIFF
--- a/service/src/main/java/bio/terra/cbas/config/CromwellServerConfiguration.java
+++ b/service/src/main/java/bio/terra/cbas/config/CromwellServerConfiguration.java
@@ -1,11 +1,12 @@
 package bio.terra.cbas.config;
 
-import cromwell.client.ApiClient;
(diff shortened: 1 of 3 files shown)


aednichols (approved):

jgainerdewar (commented):
I'll hold off on thumbing this because it should probably get a look from someone in WM, but this LGTM. 

This should make it easy to, in the future, put the workflow logs in a runset-specific subdirectory.

salonishah11 (approved):

PR #98, opened 2023-01-05
[WM-1612] Target workflow 4 pre-population

kpierre13:


service/src/main/resources/changelog/changesets/20230105_run_set_prepopulation_target_workflow_4.yaml
```diff
@@ -0,0 +1,227 @@
...
+                  value: Github
+              - column:
+                  name: last_run_set_id
+                  value: null
+        - insert:
+            tableName: method_version
+            columns:
+              - column:
+                  name: method_version_id
+                  value: 40000000-0000-0000-0000-000000000004
```
mspector, line 32: is this a convention we're using for the version? just curious where this version ID comes from

mspector (approved):

service/src/main/resources/changelog/changesets/20230105_run_set_prepopulation_target_workflow_4.yaml
```diff
@@ -0,0 +1,227 @@
...
+                  name: record_type
+                  value: FOO
+
+        - update:
+            tableName: run_set
+            columns:
+              - column:
+                  name: method_version_id
+                  value: 40000000-0000-0000-0000-000000000004
+            where: run_set_id='10000000-0000-0000-0000-000000000004'
```
cjllanwarne, line 227: I might be wrong but this looks like maybe it shouldn't be necessary? The `insert` looks like it's already adding it with this method version ID?

cjllanwarne (approved):
One comment, otherwise LGTM

service/src/main/resources/changelog/changesets/20230105_run_set_prepopulation_target_workflow_4.yaml
kpierre13: I think you're right. This is redundant.

PR #100, opened 2023-01-12
[WM-1600] DB unit tests

kpierre13:


service/src/main/java/bio/terra/cbas/BatchAnalysisWebApplication.java
```diff
@@ -8,7 +8,7 @@
 import org.springframework.context.annotation.ComponentScan.Filter;
 import org.springframework.context.annotation.FilterType;
 
-@SpringBootConfiguration
+// @SpringBootConfiguration
```
cjllanwarne, line 11: If it's unneeded, I would remove this

service/src/main/resources/application.yml
```diff
@@ -13,6 +13,14 @@ spring:
   # https://github.com/DataBiosphere/terra-common-lib/blob/480ab3daae282ddff0fef8dc329494a4422e32f1/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java#L118
   application.name: cbas
   application.version: ${cbas.version.gitHash:unknown}
+  datasource:
+    jdbc-url: jdbc:postgresql://${DATABASE_HOSTNAME:localhost}:5432/${DATABASE_NAME:cbas_db}
+    username: ${DATABASE_USER:cbas_user}
+    password: ${DATABASE_USER_PASSWORD:cbas_password}
+  sql:
+    init:
+      mode: always
```
cjllanwarne, line 22: Is this now making the other `cbas-database` (and maybe even `liquibase` too?) section redundant? If so, it might be nice to tidy them up. Right now we have 3 copy/pastes of the same set of database connectivity information within 30 lines of each other 😬 !

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
```diff
@@ -0,0 +1,81 @@
...
+    Method testMethod =
+        new Method(
+            methodId,
+            "assemble_refbased",
+            "assemble_refbased",
+            OffsetDateTime.parse("2023-01-27T19:21:24.542692Z"),
+            null,
+            "Github");
+
+    Method actual = methodDao.getMethod(methodId);
```
cjllanwarne, line 33: I saw that you added a new `createMethod` function in `MethodDao`. Was the idea to add a method before testing whether it existed or now? That might make this test more resilient to changes in the liquibase state in the future.

service/src/test/java/bio/terra/cbas/dao/TestRunSetDao.java
```diff
@@ -0,0 +1,264 @@
...
+
+@SpringBootTest
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class TestRunSetDao {
+
+  @Autowired RunSetDao runSetDao;
+  @Autowired MethodDao methodDao;
+  @Autowired MethodVersionDao methodVersionDao;
+
+  String inputDef =
```
cjllanwarne, line 25: Since these aren't actually being validated, I might be tempted to supply `null`s for these fields instead of copy/pasting these large definition strings here

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
kpierre13: That was the intention, but this became a bit more complicated for other Dao tests. I decided to keep the `createMethod` for now, since it will be useful in the future when we create more resilient tests.

service/src/main/resources/application.yml
kpierre13: I decided to keep the `cbas-database` config. There is a `CbasDatabaseConfguration.java` file that exists that uses this set up for its `@ConfigurationProperties`. Trying to point to the existing prefix `spring.datasource` caused a duplication error, so for now I opted to keep 2 copies of the database code. There is an option to add an `@Profile` annotation, but it seems that I'd still have to duplicate code in going that route.

cjllanwarne: Ok. If we can get some time with a spring expert, it might be nice to try to resolve some of these, but for now, duplicating isn't the worst thing in the world I guess

cjllanwarne: I assume removing the liquibase specific one didn't cause any problems? That's a good sign at least!

service/src/main/java/bio/terra/cbas/BatchAnalysisWebApplication.java
```diff
@@ -1,28 +0,0 @@
...
-import org.springframework.context.annotation.ComponentScan;
-import org.springframework.context.annotation.ComponentScan.Filter;
-import org.springframework.context.annotation.FilterType;
-
-// @SpringBootConfiguration
-@EnableAutoConfiguration
-@ComponentScan(
-    basePackages = {
-      "bio.terra.cbas",
-      "bio.terra.common.logging",
```
cjllanwarne, line 16: Did you really intend to delete this entire file? Was it just completely unnecessary?

cjllanwarne (approved):
Looks great, assuming deleting `service/src/main/java/bio/terra/cbas/BatchAnalysisWebApplication.java` really was necessary and safe?

service/src/main/java/bio/terra/cbas/BatchAnalysisWebApplication.java
kpierre13: I did mean to delete this. From my observation it looked like it was the original file to run the cbas app. Somewhere along the way it looks like the `App.java` file replaced it? The `App.java` file seemed to be doing everything this file was doing and more, so I opted to just get rid of it all together. I'm able to run cbas and run tests with this app deleted but I'm definitely not opposed to someone double checking in their local setup.

.github/workflows/build-and-test.yml
```diff
@@ -63,20 +63,19 @@ jobs:
...
         with:
           java-version: '17'
           distribution: 'temurin'
           cache: 'gradle'
 
-      - name: Make sure Postgres is ready and init
-        env:
-          PGPASSWORD: postgres
-        run: |
-          pg_isready -h localhost -t 10
```
salonishah11, line 77: I see the difference between this removed block and the added block on L66-71 is this line. Was it intentional to remove `pg_isready` check and move the block one step up?

service/src/main/java/bio/terra/cbas/config/BeanConfig.java
```diff
@@ -8,17 +8,10 @@
 import com.fasterxml.jackson.module.paramnames.ParameterNamesModule;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 
 @Configuration
 public class BeanConfig {
 
-  @Bean("jdbcTemplate")
-  public NamedParameterJdbcTemplate getNamedParameterJdbcTemplate(
```
salonishah11, line 17: Just to confirm, was this removal intentional?

service/src/main/resources/application.yml
```diff
@@ -28,9 +36,6 @@ spring:
           must-revalidate: true
         use-last-modified: false
   liquibase:
-    url: jdbc:postgresql://${DATABASE_HOSTNAME:localhost}:5432/${DATABASE_NAME:cbas_db}
-    user: ${DATABASE_USER:cbas_user}
-    password: ${DATABASE_USER_PASSWORD:cbas_password}
```
salonishah11, line 33: TOL: Would removal of this affect any config files in cromhelm repo for CBAS? I am not sure of it myself but just want to make sure we don't break CaaA.

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
salonishah11: TOL: Would it be better to create new methods/run sets for these DB tests (and at end of test suite remove them from DB) and use them for checks so that these tests don't break if and when we decide to remove the pre-populated workflows from database?

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
```diff
@@ -0,0 +1,81 @@
...
+            null,
+            "Github"));
+    allMethods.add(
+        new Method(
+            UUID.fromString("00000000-0000-0000-0000-000000000008"),
+            "fetch_sra_to_bam",
+            "fetch_sra_to_bam",
+            OffsetDateTime.parse("2023-01-27T19:21:24.563932Z"),
+            null,
+            "Github"));
```
salonishah11, line 75: Looks like variable `allMethods` isn't being used in the test. Can it be removed?

service/src/test/java/bio/terra/cbas/dao/TestRunDao.java
```diff
@@ -12,7 +12,11 @@
 import java.util.UUID;
 import java.util.stream.Collectors;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.springframework.boot.test.context.SpringBootTest;
 
+@SpringBootTest
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
salonishah11, line 19: Are these annotations needed?

.github/workflows/build-and-test.yml
kpierre13: Sorry I don't remember fully, as this was a change made at the beginning of this journey, but I know that the replacement of this code block was wo match more similarly what was in wds for my understanding. I think the `pg_isready` got lost in the process, and nothing broke without it, but my next commit will add it back.

service/src/main/resources/application.yml
kpierre13: This is a good point and I'm not sure either. Since I'm not familiar with cromwhelm, is there a way for me to check/test?

service/src/test/java/bio/terra/cbas/dao/TestRunDao.java
kpierre13: Yes. I wish I could tell you why, but without them the tests fail to run.

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
kpierre13: This was my initial approach that worked for some Daos, but became complicated with others. I think this is the goal for the second round of db tests to make them more resilient. I'll create a ticket for this work and put it in a placeholder swimlane.

service/src/main/resources/application.yml
salonishah11: I think this should be fine as it seems we are defining this [cbas-api.yaml](https://github.com/broadinstitute/cromwhelm/blob/189473ec60d9016c891d931519f32422fe0f495b/cromwell-helm/templates/cbas-api.yaml#L22-L26).

salonishah11 (approved):
I just have [one comment](https://github.com/DataBiosphere/cbas/pull/100#discussion_r1092560710) on which I would like your opinion. Otherwise LGTM! 🎉

PR #104, opened 2023-01-18
[WM-1623] assemble_refbased prepopulation

kpierre13:


mspector (approved):

service/src/main/resources/changelog/changesets/20230118_assemble_refbased_prepopulation.yaml
```diff
@@ -0,0 +1,1597 @@
...
+            columns:
+              - column:
+                  name: method_id
+                  value: 00000000-0000-0000-0000-000000000005
+              - column:
+                  name: name
+                  value: assemble_refbased
+              - column:
+                  name: description
+                  value: assemble_refbased / Target Workflow 2
```
mspector, line 17: wait, sorry, I know I approved this already but I'm actually kind of confused now about what this is -- are "assemble_refbased" and "Target Workflow 2" the same thing?

mspector: I see that they are - I think I was just thrown off by the PR names. This seems ok to me.

kpierre13: If it's confusing then I can get rid of it. Maybe that's better for public preview anyway.

service/src/main/resources/changelog/changesets/20230118_assemble_refbased_prepopulation.yaml
```diff
@@ -0,0 +1,1597 @@
...
+                  value: 50000000-0000-0000-0000-000000000005
+              - column:
+                  name: method_id
+                  value: 00000000-0000-0000-0000-000000000005
+              - column:
+                  name: method_version_name
+                  value: '1.0'
+              - column:
+                  name: method_version_description
+                  value: 'Target workflow assemble_refbased'
```
salonishah11, line 41: It might be good to remove `Target workflow` from here as well. Should we call it `Example  assemble_refbased` or `Sample assemble_refbased`?

service/src/main/resources/changelog/changesets/20230118_assemble_refbased_prepopulation.yaml
```diff
@@ -0,0 +1,1597 @@
...
+                  value: 50000000-0000-0000-0000-000000000005
+              - column:
+                  name: is_template
+                  value: true
+              - column:
+                  name: run_set_name
+                  value: assemble_refbased workflow
+              - column:
+                  name: run_set_description
+                  value: "assemble_refbased workflow"
```
salonishah11, line 62: Same comment here. Would it be better to call it `Example assemble_refbased workflow` or `Sample assemble_refbased workflow`?

cjllanwarne: My vote would be for `assemble_refbased sample submission` Ideally we will be hiding these from actual user view so it should be moot, but in case it leaks out somehow, that feels most "accurate" to me

PR #105, opened 2023-01-20
[WM-1649] Fix CBAS getRunSet return output

kpierre13:


service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -110,11 +110,21 @@ private RunSetDetailsResponse convertToRunSetDetails(RunSet runSet) {
...
     List<RunSet> runSets = runSetDao.getRunSets();
     List<RunSet> updatedRunSets = smartRunSetsPoller.updateRunSets(runSets);
-
-    List<RunSetDetailsResponse> runSetDetails =
-        updatedRunSets.stream().map(this::convertToRunSetDetails).toList();
-    RunSetListResponse response = new RunSetListResponse().runSets(runSetDetails);
-
+    RunSetListResponse response;
+
+    if (methodId != null) {
```
mspector, line 115: should it be possible for `methodId` to be null?

genuinely asking - I don't know if this is a pattern that we use/should use in this codebase. do we already use this method somewhere where `methodId` is null?

if we are willing to allow `methodId` to be null, then the fallback behavior you have here makes sense

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -110,11 +110,21 @@ private RunSetDetailsResponse convertToRunSetDetails(RunSet runSet) {
...
+      List<RunSet> filteredRunSet =
+          updatedRunSets.stream()
```
cjllanwarne, line 117: Instead of filtering in memory, I would be tempted to replace `runSetDao.getRunSets();` with (a new) `runSetDao.getRunSetsForMethod(methodId, pageSize);` call.

The reason is, if someone searches for run sets for a particular method, it feels a little weird to me that we would run the smart poller for _all_ run sets, then filter down in memory, and then return the result to them.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
kpierre13: Well, if we don't supply a `methodId`, then doesn't it get passed into CBAS as a null value?

cjllanwarne, line 115: 🤦 

So obvious in hindsight!!

mspector: yeah that sounds right. I was less questioning whether or not this will run as expected (I think you're right that it will) and more questioning whether we want to design methods to behave this way when arguments aren't supplied.

thinking about it ... this is more "food for thought" than something that should block acceptance. I'll be curious to hear @salonishah11 's or @cjllanwarne 's opinion

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
cjllanwarne: This would also let us respect the `pageSize` option.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
mspector: ^ I take this reaction to mean that, yes, we do want to handle missing arguments this way :)

cjllanwarne: Short answer: yes. This is how the spring/openapi libraries collaborate to implement optional parameters on APIs.

You might ask for the sense in allowing users to ask for _all run sets_ all at the same time... but in terms of implementing the API as currently defined, it is correct

service/src/main/java/bio/terra/cbas/dao/RunSetDao.java
```diff
@@ -27,18 +27,19 @@ public List<RunSet> getRunSets() {
     String sql =
         "SELECT * FROM run_set "
             + "INNER JOIN method_version ON run_set.method_version_id = method_version.method_version_id "
-            + "INNER JOIN method on method_version.method_id = method.method_id";
+            + "INNER JOIN method on method_version.method_id = method.method_id GROUP BY run_set.run_set_id, method_version.method_version_id, method.method_id ORDER BY MIN(run_set.submission_timestamp) DESC";
     return jdbcTemplate.query(sql, new RunSetMapper());
   }
 
-  public RunSet getRunSet(UUID runSetId) {
+  public RunSet getRunSet(UUID methodId) {
```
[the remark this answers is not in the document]
cjllanwarne: Unless there genuinely were no previous usages, I would suggest leaving this method unchanged, and instead adding a new `getRunSetByMethodId` alongside it.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -108,13 +108,22 @@ private RunSetDetailsResponse convertToRunSetDetails(RunSet runSet) {
...
-
+    List<RunSet> updatedRunSets;
+    RunSetListResponse response;
+
+    if (methodId != null) {
+      List<RunSet> filteredRunSet = Collections.singletonList(runSetDao.getRunSet(methodId));
+      updatedRunSets = smartRunSetsPoller.updateRunSets(filteredRunSet);
+      List<RunSetDetailsResponse> filteredRunSetDetails =
+          updatedRunSets.stream().map(this::convertToRunSetDetails).toList();
+      response = new RunSetListResponse().runSets(filteredRunSetDetails);
```
cjllanwarne, line 119: These three lines look(?) mostly the same to me... I think you could put only the choice of whether to use `getRunSet(methodId)` or `runSetDao.getRunSets()` in the `if` and keep the rest as a shared code path?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -112,13 +113,15 @@ public ResponseEntity<RunSetListResponse> getRunSets(UUID methodId, Integer page
     RunSetListResponse response;
 
     if (methodId != null) {
-      List<RunSet> filteredRunSet = Collections.singletonList(runSetDao.getRunSet(methodId));
+      List<RunSet> filteredRunSet =
+          Collections.singletonList(
+              runSetDao.getRunSetWithMethodId(methodId, Optional.ofNullable(pageSize).orElse(10)));
```
cjllanwarne, line 118: I'm not 100% sure how singletonList works, but if we're only expecting a single result back, we should probably force the `Optional.ofNullable(pageSize).orElse(10)` to be exactly 1.

Or if we want to allow more, we might want to remove that `singletonList` function.

service/src/main/java/bio/terra/cbas/dao/RunSetDao.java
```diff
@@ -23,23 +23,39 @@ public RunSetDao(NamedParameterJdbcTemplate jdbcTemplate) {
...
+        .query(sql, new MapSqlParameterSource("runSetId", runSetId), new RunSetMapper())
+        .get(0);
+  }
+
+  public RunSet getRunSetWithMethodId(UUID methodId, Integer pageSize) {
+    String sql =
+        "SELECT * FROM run_set "
+            + "INNER JOIN method_version ON run_set.method_version_id = method_version.method_version_id "
+            + "INNER JOIN method on method_version.method_id = method.method_id "
+            + "WHERE method.method_id = :methodId GROUP BY run_set.run_set_id, method_version.method_version_id, method.method_id "
```
cjllanwarne, line 52: This `GROUP BY` is a little surprising to me, since I would think we would want the entire results set rather than some grouped version

cjllanwarne (approved):
LGTM, with a few final comments / thoughts

service/src/main/java/bio/terra/cbas/dao/RunSetDao.java
kpierre13: I don't remember the exact details, but when trying to use `ORDER BY` I was getting an error on `run_set.run_set_id` about needing there to be an `ORDER BY` clause. After adding it, it gave the same error for the other two columns. Adding them solved the error. Apologies I can't remember exactly what is was.

mspector (approved):

PR #106, opened 2023-01-23
WM-1644: Handle last modified better

cjllanwarne:


service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -84,39 +86,53 @@ public List<Run> updateRuns(List<Run> runs) {
...
+          } finally {
+            recordOutboundApiRequestCompletion(
+                "wes/runSummary", getStatusStartNanos, getStatusSuccess);
+          }
+
+          CbasRunStatus newStatus = CbasRunStatus.INITIALIZING;
+          if (newWorkflowSummary != null) {
+            newStatus = CbasRunStatus.fromCromwellStatus(newWorkflowSummary.getStatus());
+          }
+          ;
```
mspector, line 117: this is getting picked up as a code smell, probably just a typo?

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -84,39 +86,53 @@ public List<Run> updateRuns(List<Run> runs) {
...
-
-      Set<Run> updatedRuns = new HashSet<>(runs);
-
-      for (Map.Entry<CbasRunStatus, List<Run>> engineStateEntry : engineStatuses.entrySet()) {
-        for (Run r : engineStateEntry.getValue()) {
-          updatedRuns.remove(r);
-          updatedRuns.add(updateDatabaseRunStatus(engineStateEntry, r));
+      Set<Run> updatedRuns = new HashSet<>();
+
+      for (Run r : runs) {
```
mspector, line 96: this block is getting picked up as a code smell for "cognitive complexity", but I think it's an improvement over what it's replacing

service/src/main/java/bio/terra/cbas/dao/RunSetDao.java
```diff
@@ -81,25 +82,45 @@ public int updateLastPolled(List<UUID> runSetIds) {
...
+                "run_count",
                 runCount,
-                RunSet.ERROR_COUNT_COL,
-                errorCount)));
+                "error_count",
+                errorCount,
+                "last_modified",
+                // If we hit the orelse, we shouldn't be adding lastModified to the DB query, so MAX
+                // should never actually be hitting
+                // the database. If it does, we have a bug!)
```
mspector, line 123: just to make sure I understand - you're saying that the `.orElse` should never actually be executed, if things are working properly?

does that mean it would be better to check whether `lastModified` is null first, and throw a descriptive error?

mspector (approved):
Just left one question for you, and I think that stray semicolon should be cleaned up, but otherwise 👍

service/src/main/java/bio/terra/cbas/dao/RunSetDao.java
cjllanwarne: The null is handled earlier in the function by not including it in the template. I suppose I could also not include it in the map unless it's not null either, to avoid the risk of database corruption if the logic is wrong. Let me try that

salonishah11 (approved):
There are 4 code smells being reported by SonarCloud. I think the one about `Cognitive Complexity` can be ignored if it is non-trivial to refactor it but other 3 seem easy to address. Otherwise code LGTM 👍

PR #107, opened 2023-01-23
[WM-1648] Remove target workflows

kpierre13:


cjllanwarne (approved):

service/src/main/resources/changelog/changesets/20230123_remove_target_workflows.yaml
```diff
@@ -0,0 +1,44 @@
...
+        - delete:
+            tableName: method_version
+            where: method_version_description = 'First version of target workflow 4'
+        - delete:
+            tableName: method
+            where: name = 'Target Workflow 4'
+
+        - delete:
+            tableName: run_set
+            where: run_set_name = 'Target workflow 3, run 1'
```
cjllanwarne, line 38: In the future, I would probably do this on ID rather than name, because (1) it's definitely unique and (2) it's less likely to have a typo, but in this case, this looks fine

salonishah11:
When running the branch locally, I am getting below error. Is it just my setup or something else?
```
Caused by: liquibase.exception.MigrationFailedException: Migration failed for change set changelog/changesets/20230123_remove_target_workflows.yaml::remove all target workflows::kpierre:
     Reason: liquibase.exception.DatabaseException: ERROR: update or delete on table "method_version" violates foreign key constraint "fk_run_set_method_version" on table "run_set"
  Detail: Key (method_version_id)=(20000000-0000-0000-0000-000000000001) is still referenced from table "run_set". [Failed SQL: (0) DELETE FROM public.method_version WHERE method_version_description = 'First version of target workflow 1']

```

service/src/main/resources/changelog/changesets/20230123_remove_target_workflows.yaml
kpierre13: That's what I was originally intending, but was getting an error about comparing `uuid to integer`. When I tried casting, it errored out saying that I couldn't do that either. Maybe there's a specific syntax that I could have used?

kpierre13:
> When running the branch locally, I am getting below error. Is it just my setup or something else?
> 
> ```
> Caused by: liquibase.exception.MigrationFailedException: Migration failed for change set changelog/changesets/20230123_remove_target_workflows.yaml::remove all target workflows::kpierre:
>      Reason: liquibase.exception.DatabaseException: ERROR: update or delete on table "method_version" violates foreign key constraint "fk_run_set_method_version" on table "run_set"
>   Detail: Key (method_version_id)=(20000000-0000-0000-0000-000000000001) is still referenced from table "run_set". [Failed SQL: (0) DELETE FROM public.method_version WHERE method_version_description = 'First version of target workflow 1']
> ```

@salonishah11  Hm, out of curiosity, did you drop and then create your db before running `bootRun`? Sorry if that's a dumb question. I didn't see this on my end when I ran it.

salonishah11:
No I didn't drop the database. My branch was up to date with main branch and then I pulled your branch and ran `gradlew bootRun`. @cjllanwarne would you be able to pull from main branch and then run this branch to see if you are seeing the same error as me?

service/src/main/resources/changelog/changesets/20230123_remove_target_workflows.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - changeSet:
+      id: "remove all target workflows"
+      author: kpierre
+      changes:
+        - delete:
+            tableName: run_set
+            where: run_set_id = '10000000-0000-0000-0000-000000000002'
+        - delete:
+            tableName: method_version
+            where: method_id = '00000000-0000-0000-0000-000000000002'
```
salonishah11, line 11: It might be good to use `method_version_id` instead of `method_id` here.

salonishah11 (approved):

PR #109, opened 2023-01-24
[WM-1640] add ParameterDefinitionNone

mspector:


cjllanwarne (approved):
LGTM. 

Something to consider: we have a `TestInputGeneratorBuildOptionalInputs` which handles specifically building input sets which include optional inputs. Would adding specific tests there be able to give you the same level of confidence that the changes are working well, without having to set up full "connected" tests via the API layer too? Or are you getting some specific reassurance by seeing it all fit together end-to-end?

mspector:
doh, no specific reason for implementing tests at the API level (btw, I double checked our test definitions glossary and apparently this is actually a "[functional check](https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2464251944/DSP+Testing+Definitions#Functional-Checks)", not a connected test, so I will change that too)

I'm happy to add a test to `TestInputGeneratorBuildOptionalInputs`. I don't think I want to remove these new functional checks, though, unless you disagree strongly

kpierre13 (approved):

PR #111, opened 2023-01-25
[WM-1658] add changeset to update fetch_sra_to_bam input/output definitions

mspector:
(this is just part of WM-1658)

Note that there is a discrepancy between the input configuration in the dev workspace vs. the data tables that Beth provided in Slack:

In the dev workspace, the variable `SRA_ID` is supplied by the attribute `accession`, but in Beth's TSV file, it's supplied by the attribute `sra_id`. This PR assumes that Beth's TSV is most accurate.


kpierre13 (approved):

cjllanwarne (changes requested):
Sorry to do this to you... could you rebase this on my output definition changes branch to make sure we're using the new style outputs?

PR #113, opened 2023-01-25
WM-1660: Send nothing instead of null when CBAS chooses not to supply an input

cjllanwarne:
With apologies to reviewers for a few stowaway tidy ups:
1. Using `instanceof` (with assignment) instead of type name comparisons and casting to determine which type of input is being supplied
2. Fall back case in case an unhandled new input type is provided
3. Refactor of input handling exceptions to be more like output exceptions


service/src/main/java/bio/terra/cbas/runsets/inputs/InputGenerator.java
```diff
@@ -51,12 +52,15 @@ public static Map<String, Object> buildInputs(
...
 
-      // Convert into an appropriate CbasValue:
-      CbasValue cbasValue = CbasValue.parseValue(param.getInputType(), parameterValue);
-
-      params.put(parameterName, cbasValue.asSerializableValue());
+      if (parameterValue != null) {
+        // Convert into an appropriate CbasValue:
+        CbasValue cbasValue = CbasValue.parseValue(param.getInputType(), parameterValue);
+        params.put(parameterName, cbasValue.asSerializableValue());
+      }
```
cjllanwarne, line 63: For reviewers: these are the lines that actually implement the requested feature update

kpierre13 (approved):

PR #114, opened 2023-01-26
[WM-1658] match inputs/outputs for assemble_refbased to test workspace

mspector:


service/src/main/resources/changelog/changesets/20230125_assemble_refbased_io_updates.yaml
```diff
@@ -0,0 +1,1417 @@
...
+                      },
+                      {
+                        "input_name": "assemble_refbased.reference_fasta",
+                        "input_type": {
+                          "type": "primitive",
+                          "primitive_type": "File"
+                        },
+                        "source": {
+                          "type": "record_lookup",
+                          "record_attribute": "workspace.ncov_reference_genome_fasta"
```
mspector, line 37: important: workspace data isn't implemented yet. what should we do here?

service/src/main/resources/changelog/changesets/20230125_assemble_refbased_io_updates.yaml
```diff
@@ -0,0 +1,1417 @@
...
+                        "input_name": "assemble_refbased.plot_self_coverage.binning_summary_statistic",
+                        "input_type": {
+                          "type": "optional",
+                          "optional_type": {
+                            "type": "primitive",
+                            "primitive_type": "String"
+                          }
+                        },
+                        "source": {
+                          "type": "none"
```
kpierre13, line 64: Definitely not a blocker, but out of curiosity, does case matter for type 'none'?

kpierre13 (approved):

service/src/main/resources/changelog/changesets/20230125_assemble_refbased_io_updates.yaml
mspector: I don't know - that's a good question. all lowercase is what we want, but maybe other cases would work too?

mspector:
closing; redundant with https://github.com/DataBiosphere/cbas/pull/115

PR #115, opened 2023-01-26
[WM-1658] [WM-1669] assemble_refbased and sarscov2_nextstrain I/O updates

mspector:


kpierre13 (approved):

service/src/main/resources/changelog/changesets/20230125_assemble_refbased_io_updates.yaml
```diff
@@ -0,0 +1,1417 @@
...
+                    [
+                      {
+                        "input_name": "assemble_refbased.reads_unmapped_bams",
+                        "input_type": {
+                          "type": "array",
+                          "array_type": {
+                            "type": "primitive",
+                            "primitive_type": "File"
+                          },
+                          "non_empty": true
```
salonishah11, line 22: Sorry if this is a naive question but does this basically say that this is not an optional Array input?

cjllanwarne: Non empty means it's an error if the array is empty (eg the difference between `null` vs `[]`)

service/src/main/resources/changelog/changesets/20230126_sarscov2_nextstrain_io_updates.yaml
```diff
@@ -0,0 +1,1485 @@
...
+                            "type": "primitive",
+                            "primitive_type": "String"
+                          }
+                        },
+                        "source": {
+                          "type": "none"
+                        }
+                      },
+                      {
+                        "input_name": "sarscov2_nextstrain.derived_cols.table_map",
```
cjllanwarne, line 297: The UI is showing this as "undefined(optional)", which we might want to dig into. I suspect it's a UI display issue rather than anything wrong in prestaging, just something to be aware of

service/src/main/resources/changelog/changesets/20230125_assemble_refbased_io_updates.yaml
```diff
@@ -0,0 +1,1417 @@
...
+                          "record_attribute": "dist_to_ref_snps"
+                        }
+                      },
+                      {
+                        "output_name": "assemble_refbased.ivar_trim_stats",
+                        "output_type": {
+                          "type": "array",
+                          "array_type": {
+                            "type": "primitive",
+                            "primitive_type": "String"
```
cjllanwarne, line 1214: In our sample workspace this is `Array[Map[String, String]]`. 

Since we don't want to support maps yet, and the output is set to "none", this is probably ok for now... but I wonder whether we'd be better off defining the map type and just failing if people try to use it, rather than pretending that maps are strings?

cjllanwarne (approved):

service/src/main/resources/changelog/changesets/20230125_assemble_refbased_io_updates.yaml
cjllanwarne: I'll make a jira ticket for it

service/src/main/resources/changelog/changesets/20230126_sarscov2_nextstrain_io_updates.yaml
mspector: ah, so it is... interesting.

I wonder if it's worth holding off on merging this until we know for sure that it's just a UI display issue

service/src/main/resources/changelog/changesets/20230125_assemble_refbased_io_updates.yaml
salonishah11: Ah I see 👍

salonishah11 (approved):

service/src/main/resources/changelog/changesets/20230125_assemble_refbased_io_updates.yaml
```diff
@@ -0,0 +1,1426 @@
...
+                      },
+                      {
+                        "input_name": "assemble_refbased.reference_fasta",
+                        "input_type": {
+                          "type": "primitive",
+                          "primitive_type": "File"
+                        },
+                        "source": {
+                          "type": "literal",
+                          "parameter_value": "gs://fc-7ae14861-7808-44e7-8a0a-0a288d691bc3/reference_files/ref-sarscov2-NC_045512.2.fasta"
```
mspector, line 37: this is going to need to be changed on Monday 1/30

service/src/main/resources/changelog/changesets/20230125_assemble_refbased_io_updates.yaml
```diff
@@ -0,0 +1,1426 @@
...
+                            "map_type": {
+                              "key_type": {
+                                "type": "primitive",
+                                "primitive_type": "String"
+                              },
+                              "value_type": {
+                                "type": "primitive",
+                                "primitive_type": "String"
+                              }
+                            }
```
cjllanwarne, line 1223: It turned out the `map_type` wrapper around key_type and value_type is unnecessary, so we just need key and values when the type is specified as `map`

```suggestion
                            "type": "map",
                            "key_type": {
                              "type": "primitive",
                              "primitive_type": "String"
                            },
                            "value_type": {
                              "type": "primitive",
                              "primitive_type": "String"
                            }
```

cjllanwarne (changes requested):

cjllanwarne (approved):

PR #116, opened 2023-01-27
[WM-1657] Workflow order presentation

kpierre13:
…d date

### DIFF
--- a/service/src/main/java/bio/terra/cbas/dao/MethodDao.java
+++ b/service/src/main/java/bio/terra/cbas/dao/MethodDao.java
@@ -34,7 +34,7 @@ public class MethodDao {
   }
 
   public List<Method> getMethods() {
-    String sql = "SELECT * FROM method GROUP BY method.method_id ORDER BY MIN(created) DESC";
+    String sql = "SELECT * FROM method ORDER BY created DESC";
     return jdbcTemplate.query(sql, new MethodMapper());
   }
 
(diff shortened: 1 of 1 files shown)


mspector (approved):
this does seem simpler :)

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
```diff
@@ -34,7 +34,7 @@ public Method getMethod(UUID methodId) {
   }
 
   public List<Method> getMethods() {
-    String sql = "SELECT * FROM method";
+    String sql = "SELECT * FROM method GROUP BY method.method_id ORDER BY MIN(created) DESC";
```
cjllanwarne, line 37: It feels really surprising that you need the group by here... are you sure this doesn't work?
```suggestion
    String sql = "SELECT * FROM method ORDER BY created DESC";
```

cjllanwarne (approved):
LGTM albeit with an eyebrow raised...

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
kpierre13: This totally works. I don't know why I thought I needed to use `MIN()`.  That was forcing me to use the `GROUP BY` method.

cjllanwarne (approved):
Re-thumb from me!

PR #124, opened 2023-02-08
[WM-1688] move non_empty field from optional to array

mspector:


service/src/main/resources/changelog/changesets/20230207_sarscov2_nextstrain_nonempty_optionals.yaml
```diff
@@ -0,0 +1,1312 @@
...
+                          "optional_type": {
+                            "type": "array",
+                            "array_type": {
+                              "type": "optional",
+                              "optional_type": {
+                                "type": "primitive",
+                                "primitive_type": "File"
+                              }
+                            },
+                            "non_empty": false
```
mspector, line 309: since you can't see it in a diff: this is an example of the kind of change made in this changeset.

Previously, `non_empty` appeared as a property of `input_type` -- but `optional` doesn't have a `non_empty` property, so it caused an error.

this change moves `non_empty` to be a property of `optional_type`, which is an `array`

kpierre13 (approved):

service/src/main/resources/changelog/changesets/20230207_sarscov2_nextstrain_nonempty_optionals.yaml
```diff
@@ -0,0 +1,1313 @@
...
+                          }
+                        },
+                        "source": {
+                          "type": "record_lookup",
+                          "record_attribute": "combined_assemblies"
```
mspector, line 54: this was the other change. I was told that the `assembly_fastas` input variable defaults to a Google Cloud file, so we need to set this to read from the `combined_assemblies` data table attribute instead.

salonishah11 (approved):
Looked at the actual changes made and they LGTM!

PR #126, opened 2023-02-14
[WM-1713] Don't return templated run sets in GET /run_sets API

salonishah11:
Tested locally:
GET /run_sets doesn't return templated run sets and  UI only displays 1 submission that I had run as expected
Screen Shot 2023-02-14 at 3 24 06 PM

Closes https://broadworkbench.atlassian.net/browse/WM-1713

### DIFF
--- a/service/src/test/java/bio/terra/cbas/dao/TestRunSetDao.java
+++ b/service/src/test/java/bio/terra/cbas/dao/TestRunSetDao.java
@@ -79,5 +79,8 @@ class TestRunSetDao {
     // DB has no non-templated run sets, so it will return empty list
     List<RunSet> runSets = runSetDao.getRunSets(10, false);
     assertEquals(0, runSets.size());
+
+    List<RunSet> templateRunSets = runSetDao.getRunSets(10, true);
+    assertEquals(3, templateRunSets.size());
   }
 }
--- a/service/src/test/java/bio/terra/cbas/dao/TestRunSetDao.java
+++ b/service/src/test/java/bio/terra/cbas/dao/TestRunSetDao.java
@@ -76,8 +76,8 @@ class TestRunSetDao {
 
   @Test
   void retrievesAllRunSets() {
-
-    List<RunSet> runSets = runSetDao.getRunSets(10);
-    assertEquals(3, runSets.size());
+    // DB has no non-templated run sets, so it will return empty list
+    List<RunSet> runSets = runSetDao.getRunSets(10, false);
+    assertEquals(0, runSets.size());
   }
 }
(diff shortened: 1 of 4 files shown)


service/src/test/java/bio/terra/cbas/dao/TestRunSetDao.java
```diff
@@ -76,8 +76,8 @@ void retrievesSingleRunSet() {
 
   @Test
   void retrievesAllRunSets() {
-
-    List<RunSet> runSets = runSetDao.getRunSets(10);
-    assertEquals(3, runSets.size());
+    // DB has no non-templated run sets, so it will return empty list
```
cjllanwarne, line 79: Do you want to keep the "returns everything" version of this test case too (the one that would return 3 if the `isTemplate` is true)?

cjllanwarne (approved):

service/src/test/java/bio/terra/cbas/dao/TestRunSetDao.java
salonishah11: Yeah I was debating whether to keep it or not. If you think it should be present I can keep it. 

One thing to note that I had to wrap my head around is that it would be "return templated run sets" as compared to "return everything". If we want to have "return everything" for this API then the implementation of [getRunSets()](https://github.com/DataBiosphere/cbas/blob/9624648eae6b60940fc913152bcf29e3d8c01b1f/service/src/main/java/bio/terra/cbas/dao/RunSetDao.java#L27) will be different. I choose to do this way because I thought that templated run sets are side-effect of how we store inputs & outputs definiton which is a Tech Debt and might go away, and hence we might never need to return in API response. What do you think?

cjllanwarne: +1 to not having to modifying the API. But since the switch exists in the DAO function, testing it at the DAO layer feels sensible (in case someone finds and relies on it later)

mspector (approved):
LGTM

A non-blocking thought: do you think adding the boolean `isTemplate` argument is better than an entirely new function that just gets templates?

The way you have it, we reuse code better, but it also kind of smashes together two concepts (run sets vs. templates) that are mostly similar because of the way we implement this right now.

Just thinking out loud. This is a simple enough situation that it probably doesn't matter much in practice

salonishah11:
@mspector that is a valid point. I choose to do this way because it avoids code redundancy. And since template run sets are tech debt item, I don't think we might need to just get template run sets. But mainly I did this to avoid code redundancy since the query for get template run sets and get non-templated run sets would be the same.

PR #129, opened 2023-02-23
[WM-1737] Remove hard coded page size limit from GET /run_sets

salonishah11:
Pagination works as expected
Screen Shot 2023-02-23 at 5 09 18 PM

Closes https://broadworkbench.atlassian.net/browse/WM-1737

### DIFF
--- a/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
@@ -45,7 +45,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.UUID;
 import org.databiosphere.workspacedata.client.ApiException;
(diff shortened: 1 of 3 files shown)


service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -116,14 +115,13 @@ public ResponseEntity<RunSetListResponse> getRunSets(UUID methodId, Integer page
 
     if (methodId != null) {
       List<RunSet> filteredRunSet =
-          Collections.singletonList(
-              runSetDao.getRunSetWithMethodId(methodId, Optional.ofNullable(pageSize).orElse(10)));
+          Collections.singletonList(runSetDao.getRunSetWithMethodId(methodId));
```
salonishah11, line 118: We only return latest run set associated with the method hence there is no limit to pass in page size here and it's associated DAO method.

mspector (approved):

PR #144, opened 2023-03-19
[WM-1796] Validate Methods

kpierre13:


service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -46,7 +46,7 @@
 @Controller
 public class MethodsApiController implements MethodsApi {
   private final CromwellService cromwellService;
-  private final MethodDao methodDao;
+  private static MethodDao methodDao;
```
cjllanwarne, line 49: This looks super suspicious to me, especially since you're still supplying a `methodDao` and overwriting the value in the constructor. What was the motivation behind this, and is it possible to revert it back to `final` and not `static`?

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
```diff
@@ -95,4 +95,28 @@ public int updateLastRunWithRunSet(RunSet runSet) {
                 "run_set_id", runSet.runSetId(),
                 "method_id", runSet.methodVersion().method().methodId())));
   }
+
+  public int getNumberOfMethods(
```
cjllanwarne, line 99: I would maybe call this `countMethods` to be closer to the SQL action in question

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
```diff
@@ -95,4 +95,28 @@ public int updateLastRunWithRunSet(RunSet runSet) {
...
+        new MapSqlParameterSource(
+            Map.of(
+                "name", methodName,
+                "methodVersionUrl", methodUrl,
+                "methodVersionName", methodVersion,
+                "methodSource", methodSource));
+
+    Integer count = jdbcTemplate.queryForObject(sql, params, Integer.class);
+
+    return count;
```
cjllanwarne, line 120: You could probably do this all in one line `return jdbcTemplate.queryForObject...`

cjllanwarne (changes requested):

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
kpierre13: I wasn't sure what the best way to handle the error that I was getting. When this is `final` line 245 throws an error ```Non-static field 'methodDao' cannot be referenced from a static context```. Making the function non-static seemed like not the right thing to do either, so I'd be curious what the right thing to do here is in this situation.

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -334,6 +336,32 @@ void validPostRequest() throws Exception {
...
+        new ArrayList<>(List.of("Method method1 already exists. Please select a new method."));
+
+    when(methodDao.getNumberOfMethods(
+            duplicateMethodRequest.getMethodName(),
+            duplicateMethodRequest.getMethodUrl(),
+            duplicateMethodRequest.getMethodVersion(),
+            duplicateMethodRequest.getMethodSource().toString()))
+        .thenReturn(1);
+
+    List<String> actualErrors = MethodsApiController.validateMethod(duplicateMethodRequest);
```
cjllanwarne, line 361: There's potentially a super-subtle [TOCTTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) race condition if we use `validateMethod` here. 

Imagine someone writes a buggy script to import workflows and it sends two requests at exactly the same time (or perhaps a user clicks an import button in the UI twice in quick succession). This check for "does it exist" might pass in both threads, and then two instance of the method might still be written to the database in parallel.

The "validate" method was until now mostly concerned with general validity/invalidity of methods, but is now checking validity in a way that changes over time (a workflow which is valid at time t0 might no longer be valid at time t10 or t100).

This isn't particularly bad here (the worst case scenario is both very unlikely to happen, AND is hardly the end of the world!), but it's good to be in the habit of thinking about these things. But something to think about: if the intention is to never allow duplicate entries containing the same set of these values, an alternative would be to create a unique key constraint across those fields (see [example](https://dba.stackexchange.com/questions/316319/how-to-check-record-if-exists-before-insert-with-postgresql)). That way, the insert would fail automatically if another row already contained the values, and no special check would be needed ahead of time.

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
cjllanwarne: In this situation, the other method _should_ be non-static, because it's now relying on a class field (methodDao) that gets supplied at construction time. So it should be fine to make that other method an instance method (ie non-static)

This actually neatly dovetails with my other comment (about time of check / time of use race conditions). The reason the other validate method is no longer static is because it's now checking live database state, instead of inherent properties of the method.

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
kpierre13: I think this makes complete sense. Thank you for pointing this out. I think it'll be best to create a new ticket for a changelog that adds the `UNIQUE` key constraint on these columns.

salonishah11: I am wondering if we can add the `UNIQUE` constraints as part of this PR. The reason is that if we add the constraints in this ticket itself, we might not need to do this validation check in the `validateMethod` as the database itself would throw and error when creating duplicate methods and we can catch that and return a "duplicate method..." error back to the user? Plus it fits within the scope of this ticket itself. What do you think @kpierre13 @cjllanwarne?

service/src/main/resources/changelog/changesets/20230323_add_unique_constraint_to_methods_columns.yaml
```diff
@@ -0,0 +1,14 @@
+databaseChangeLog:
+  - changeSet:
+      id: "addUniqueConstraint to Method and MethodVersion table"
```
salonishah11, line 3: 🎉

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
```diff
@@ -70,6 +70,7 @@ passing in a current_timestamp() value, causing the test to fail.
   @Test
   void retrievesAllMethods() {
     List<Method> allMethods = methodDao.getMethods();
+    System.out.println(allMethods);
```
salonishah11, line 73: This can probably be removed

salonishah11 (approved):
Thank you for looking into adding the constraints as part of this PR itself. Great job! 💯

PR #145, opened 2023-03-22
[WM-1777] POST /run_sets/abort endpoint

kpierre13:


common/openapi.yml
```diff
@@ -707,6 +734,22 @@ components:
           type: string
         output_definition:
           type: string
+
+    AbortRunSetResponse:
+      type: object
+      required: [ run_set_id, runs ]
+      properties:
```
salonishah11, line 741: Maybe we can also return the status of the request? Probably return `Canceling` state? What do you think?

kpierre13: I think the `Canceling` state would be goo to return here.

salonishah11 (approved):

mspector (approved):

PR #146, opened 2023-03-24
[WM-1778] Abort endpoint calls cromwell

kpierre13:


service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -123,4 +123,8 @@ public static String getErrorMessage(List<FailureMessage> failureMessages) {
       return result;
     }
   }
+
+  public RunId submitAbortWorkflow(Run run) throws ApiException {
```
cjllanwarne, line 127: This naming feels a little obtuse... is there a reason to not also call this "cancelRun"?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -231,14 +232,22 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+        runDao.getRuns(new RunDao.RunsFilters(runSetId, NON_TERMINAL_STATES));
+    List<String> submittedAbortWorkflows = new ArrayList<>();
 
-    aborted.runSetId(runSetId);
+    for (Run run : runningWorkflows) {
+      try {
+        cromwellService.submitAbortWorkflow(run);
+        submittedAbortWorkflows.add(run.engineId());
+      } catch (cromwell.client.ApiException e) {
+        log.error("Unable to abort workflow %s".formatted(run.engineId()), e);
```
cjllanwarne, line 246: To think about, do we want to:
- Return this error (or perhaps a set of errors) to the user?
- Let the user know that not everything went according to plan?
- Mark these errors against the workflows?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -231,14 +232,22 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+    aborted.runSetId(runSetId);
 
-    dummyRunsList.add(UUID.randomUUID());
-    dummyRunsList.add(UUID.randomUUID());
+    List<Run> runningWorkflows =
+        runDao.getRuns(new RunDao.RunsFilters(runSetId, NON_TERMINAL_STATES));
+    List<String> submittedAbortWorkflows = new ArrayList<>();
 
-    aborted.runSetId(runSetId);
+    for (Run run : runningWorkflows) {
```
cjllanwarne, line 241: To think about in the future:
- We will probably need to consider what happens when we overflow the number of workflows we can abort all at the same time

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
kpierre13: By the last bullet point, do you mean adding the message to the `errors` columns in the db?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -231,14 +232,22 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+      try {
+        cromwellService.submitAbortWorkflow(run);
+        submittedAbortWorkflows.add(run.engineId());
+      } catch (cromwell.client.ApiException e) {
+        log.error("Unable to abort workflow %s".formatted(run.engineId()), e);
+      }
+    }
+    aborted.runs(submittedAbortWorkflows);
```
salonishah11, line 249: I think we should also mark the Run Ids in `Canceling` state for which calling the abort endpoint was successful. For Run Ids where we received the ApiException we shouldn't mark them in `Canceling` state.

Related to this, the Run Set should also be marked in `Canceling` state in database.

salonishah11:
TOL: In the Run Sets poller, for a Run Set that is in Canceling state we should check:
- if all the Runs in the Run Set are  in terminal state, then mark the Run Set as `Canceled`
- if any Runs are still in either `Running, Initializing, Queued or Unknown` state we should submit an abort request to Cromwell for those Runs (these might be Runs for which we received Exception when the initial abort request was submitted)

What do you think @kpierre13 @cjllanwarne?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
kpierre13: Does this happen automatically in the `SmartRunsPoller`? When I testing this in CBAS-UI, the status automatically went from `Canceling` to `Canceled` for both run sets and runs by itself.

cjllanwarne (approved):

cjllanwarne:
@kpierre13 I think Saloni is specifically talking about the run sets poller, and makes a great point. The logic would presumably be somethig like:

- When CBAS receives a cancel request:
  - Mark the run set as cancelling immediately.
  - Try to cancel all of the workflows immediately
    - For successful cancel calls, mark the run state as Cancelling
    - If calling cancel in Cromwell fails, don't update the run state
  - In the run set smart poller:
    - If the run set is cancelling but an individual run is still running, try again to cancel it
    - And if successful this time, mark it as Cancelling
    - When a "cancelling" run finishes, mark it as canceled
    - When all runs in a "Cancelling" run set are complete, mark the run set as Cancaled.

I'd be happy to see you make the run set changes in a separate ticket/PR, but I do agree with Saloni that we need to tidy up the runs sets somehow as part of the Abort Workflows epic.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Ah right, that would be because the UI was polling for status updates every 1 min and Cromwell was returning `Canceling` state for each run. But I think it would be better to also mark the Runs here itself in Canceling state. CBAS should not rely on UI to poll for status update of Runs to have the Runs transition to Canceling state.

And since Cromwell doesn't know about Run Sets, we should mark the Run Set in Canceling state here too.

salonishah11:
>In the run set smart poller:
If the run set is cancelling but an individual run is still running, try again to cancel it
.....

This is a  much better explanation to what I was trying to say. Thank you @cjllanwarne! I am happy for this to be added in a separate ticket and I agree with Chris that it should definitely be part of Abort functionality epic.

kpierre13:
I created a follow-up ticket to do the work both @salonishah11 and @cjllanwarne detailed in the comments [here](https://broadworkbench.atlassian.net/browse/WM-1847)

common/openapi.yml
```diff
@@ -751,6 +751,10 @@ components:
           items:
             type: string
             format: uuid
+        errors:
+          type: array
+          items:
+            type: string
```
salonishah11, line 757: It seems in other places in this file `errors` is just a string instead of Array of string. I think we should make it same as others. What do you think?

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -499,6 +505,112 @@ void getRunSetsApiTest() throws Exception {
     assertEquals(
         CbasRunSetStatus.toCbasRunSetApiState(CbasRunSetStatus.RUNNING), runSetDetails2.getState());
   }
+
+  @Test
+  void testRunSetAbort() throws Exception {
```
salonishah11, line 510: It would be great to add another test case where abort request for 1 run fails and the other one succeeds. We should verify in that test case that the status of that particular run whose abort request failed was not Canceling.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -231,14 +232,47 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+
+    for (Run run : runningWorkflows) {
+      // Trying inside the for-loop in case a single run fails to be updated
+      try {
+        cromwellService.cancelRun(run);
+        submittedAbortWorkflows.add(run.runId());
+        // Update each run to have a 'canceling' run state
+        runDao.updateRunStatus(run.runId(), CbasRunStatus.CANCELING, OffsetDateTime.now());
+      } catch (cromwell.client.ApiException e) {
+        String msg = "Unable to abort workflow %s.".formatted(run.runId());
```
salonishah11, line 262: nit: Adding this message prefix `Unable to abort workflow` here will make it repetitive for each run. So if 3 runs had exceptions the final error message that user will see would be
``` 
Run set canceled with errors: Unable to abort workflow X. Unable to abort workflow Y. Unable to abort workflow Z
```
To make it DRY this prefix message can be appended on L271. Something like:
```
Run set canceled with errors. Unable to abort workflow(s): X, Y, Z
```

salonishah11 (approved):

PR #149, opened 2023-04-07
[WM-1814] provider verification test

mspector:


service/build.gradle
```diff
@@ -52,10 +52,26 @@ dependencies {
...
+    } 
+}
+
+task verifyPacts( type: Test ) {
+    useJUnitPlatform()
+    systemProperty 'pact.provider.version', System.getenv('PACT_PROVIDER_COMMIT')
+    systemProperty 'pact.provider.branch',  System.getenv('PACT_PROVIDER_BRANCH')
+    systemProperty 'pact.verifier.publishResults', true
+    filter {
+        includeTestsMatching "*VerifyPacts*"
```
mspector, line 72: looking for feedback on whether it makes sense to filter by a naming convention like this

kpierre13: This block of code is adding two tasks, one for our existing unit tests and the other specifically for Pact tests? If so, I think this makes sense, especially seeing the logic for `excludingTestsMatching`. I think separating these two kinds of tests will make debugging in the future a lot easier.

kpierre13 (approved):

service/build.gradle
mspector: > This block of code is adding two tasks, one for our existing unit tests and the other specifically for Pact tests?

Yes, exactly. There are some DAO tests that fail when I run them together, which is something we should address at some point, but I think there's value in separating these tests anyway

salonishah11: Maybe a naive question, but does this mean Pact tests won't run on PRs?

mspector: no, I think they should run on PRs. the github action has been updated to make that happen

salonishah11 (approved):

PR #150, opened 2023-04-20
Mls wm 1811 post run sets provider verification

mspector:


service/src/test/java/bio/terra/cbas/controllers/VerifyPactsRunSetsApiController.java
```diff
@@ -65,20 +72,80 @@ void before(PactVerificationContext context) {
     context.setTarget(new MockMvcTestTarget(mockMvc));
   }
 
+  @State({"initialize dependencies"})
```
cjllanwarne, line 75: These state names could probably be more descriptive... like "Exactly one run set with ID ...." (since we're pinning it, we might as well make it easy to find?)

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsRunSetsApiController.java
```diff
@@ -65,20 +72,80 @@ void before(PactVerificationContext context) {
...
+    when(runDao.createRun(any())).thenReturn(1);
+  }
+
+  @State({"initialize run_set"})
+  public HashMap<String, String> postRunSets() throws Exception {
+    String fixedRunSetUUID = "11111111-1111-1111-1111-111111111111";
+    String fixedRunUUID = "22222222-2222-2222-2222-222222222222";
+    when(uuidSource.generateUUID())
+        .thenReturn(UUID.fromString(fixedRunSetUUID))
+        .thenReturn(UUID.fromString(fixedRunUUID));
```
cjllanwarne, line 121: This feels like it's assuming how the state will be used rather than just setting up non-opinionated state. Is that going to be brittle? Like if we use the `initialize run_set` state to create 2 run sets in a consumer test instead of 1, will everything after the second call to generateUUID just return the same value?

cjllanwarne (approved):
LGTM for now, with a slight concern about naming and a slight concern about brittleness

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsRunSetsApiController.java
mspector: I don't think it's possible to use this state twice in the way you're suggesting, but I understand your point about it being named/identified in the wrong way. Something like your above suggestion "Exactly one run state with ID..." is probably the way to go.

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsRunSetsApiController.java
mspector: good point, I will change that!

cjllanwarne (approved):

salonishah11 (approved):

PR #151, opened 2023-04-20
WM-1764, WM-1862: Ad hoc WDS URL and authorized requests

cjllanwarne:


mspector (approved):
Based on your suggestions in your walkthrough, I focused on `findUrlForWds` and the associated tests. It looks good to me, based on what I know about the algorithm.

The only thing that stuck out to me was the choice to focus the tests on `findUrlForWds` as opposed to the inner function `findBestAppForWds`. Just curious if that was a conscious choice, or if `findUrlForWds` just felt like the right minimum testable unit

service/src/main/java/bio/terra/cbas/dependencies/leonardo/AppUtils.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public AppUtils(
+      LeonardoServerConfiguration leonardoServerConfiguration,
+      WdsServerConfiguration wdsServerConfiguration) {
+    this.leonardoServerConfiguration = leonardoServerConfiguration;
+    this.wdsServerConfiguration = wdsServerConfiguration;
+  }
+
+  int appComparisonFunction(ListAppResponse a, ListAppResponse b) {
+    // Name scores: correct name always wins
```
salonishah11, line 30: It seems [in Terra UI app type along with app name check](https://github.com/DataBiosphere/terra-ui/blob/09fc15b510521e7443c2ab9fda2bf994b78bb27d/src/libs/ajax/data-table-providers/WdsDataTableProvider.ts#L107-L112) takes precedence over other methods to find the WDS proxy url. Maybe we should also include that here so that `WDS` app type in given higher precedence?

cjllanwarne: 👍

cjllanwarne:
@mspector re 
> "The only thing that stuck out to me was the choice to focus the tests on findUrlForWds as opposed to the inner function findBestAppForWds. Just curious if that was a conscious choice, or if findUrlForWds just felt like the right minimum testable unit"

Basically, yes. I'm treating this `findUrlForWds` as a function from "list of apps" to "single URL" and testing it as a unit. I think that's (a) acceptable because it hasn't got any external dependencies and could be thought of as a pure mathematical function, and (b) appropriate because it tests the interface to this class, which is unlikely to ever change, even if the internal function layout moves around. But there is a risk that I'm not testing some edge case because of all the setup involved, so let me know if you think there's some coverages missing that would be easy enough to test at a lower level?

service/src/main/java/bio/terra/cbas/dependencies/common/ScheduledHealthChecker.java
```diff
@@ -0,0 +1,53 @@
...
+      CromwellService cromwellService, WdsService wdsService, LeonardoService leonardoService) {
+    this.healthCheckStatuses = new ConcurrentHashMap<>();
+    this.healthCheckSystems =
+        Map.of(
+            "cromwell", cromwellService,
+            "wds", wdsService,
+            "leonardo", leonardoService);
+  }
+
+  @Scheduled(fixedDelay = 5 * 60 * 1000)
```
salonishah11, line 38: Should this be made configurable?

mspector:
> so let me know if you think there's some coverages missing that would be easy enough to test at a lower level?

No need as far as I'm concerned. Thanks for explaining!

service/src/main/java/bio/terra/cbas/dependencies/leonardo/AppUtils.java
```diff
@@ -29,7 +29,16 @@ public AppUtils(
   }
 
   int appComparisonFunction(ListAppResponse a, ListAppResponse b) {
-    // Name scores: correct name always wins
+
+    // First criteria: Prefer apps with the WDS app type.
+    // NB: Negative because lower index is better
```
mspector, line 34: 👍

service/src/test/java/bio/terra/cbas/dependencies/leonardo/TestAppUtils.java
```diff
@@ -110,6 +109,14 @@ void throwIfBestAppHasNoWDS() {
     assertThrows(DependencyNotAvailableException.class, () -> au.findUrlForWds(apps));
   }
 
+  private void permuteAndTest(List<ListAppResponse> apps, String expectedUrl) throws Exception {
```
mspector, line 109: I like this

PR #155, opened 2023-05-02
[WM-1963] Validator for Jira tags in PR names

cjllanwarne:


kpierre13 (approved):

.github/workflows/validate_pr_name.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate PR title
+        id: validate
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const title = context.payload.pull_request.title;
+            const regex = /[A-Z]+-\d+/;
```
salonishah11, line 20: Do we want to make the regex same as the one we use to ensure a JIRA ticket number is present in [tag-and-publish.yaml](https://github.com/DataBiosphere/cbas/blob/7b63b9b7b3e66bb7a22c87d85d33cff1b2011550/.github/workflows/tag-and-publish.yml#L27)?

salonishah11 (approved):

PR #161, opened 2023-05-04
WM-1971: Make API debug logging a config option

cjllanwarne:


kpierre13 (approved):
Not a blocker: in the future, all configurations (cromwell, wds, leo, etc) we want to see debugging for, will need to manually be changed to true in their respective files? Or is simply changing the `application.yml` file sufficient?

salonishah11 (approved):

cjllanwarne:
@kpierre13 
> In the future, all configurations (cromwell, wds, leo, etc) we want to see debugging for, will need to manually be changed to true in their respective files? Or is simply changing the application.yml file sufficient?

Yeah, we'll need to add the equivalent to the `apiClient.setDebugging(leonardoServerConfiguration.debugApiLogging());` line for every dependency, and wire it through from the appropriate part of the CBAS config. But then when we want to turn it on in a dev environment, we can just update the app config and no longer need to do any re-coding in CBAS itself. Does that answer the question?

PR #162, opened 2023-05-08
WM-1974: Fix string truncation for error messages

cjllanwarne:


service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -82,6 +82,14 @@ public int updateRunStatus(
                 currentTimestamp)));
   }
 
+  static String truncatedErrorMessage(String errorMessage) {
```
mspector, line 85: is this being used anywhere other than the tests?

mspector (approved):

salonishah11 (approved):

PR #163, opened 2023-05-11
[WM-1847] Tidy up run sets

kpierre13:


common/openapi.yml
```diff
@@ -778,7 +778,6 @@ components:
           type: array
           items:
             type: string
-            format: uuid
```
salonishah11, line 781: Just out of curiosity, any specific reason to remove this?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -238,40 +241,12 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
-        runDao.updateRunStatus(run.runId(), CbasRunStatus.CANCELING, OffsetDateTime.now());
-      } catch (cromwell.client.ApiException e) {
-        String msg = "Unable to abort workflow %s.".formatted(run.runId());
-        log.error(msg, e);
-        failedRunIds.add(run.runId().toString());
-        // Add the error message against the run
-        run.withErrorMessage(msg);
-      }
-    }
+    Map<String, List<String>> abortDetails = abortManager.abortRunSet(runSetId);
```
salonishah11, line 247: It might be a good idea to convert abort details `Map<String, List<String>>` into its own class called `AbortRequestDetails` with 2 properties `abortRequestFailedIds` and `abortRequestSubmittedIds`. This way it would be easier to access their values rather than using the Map key.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunSetsPoller.java
```diff
@@ -99,6 +102,30 @@ private RunSet updateRunSet(RunSet rs, OffsetDateTime limitedEndTime) {
       smartRunsPoller.updateRuns(updateableRuns, Optional.of(runPollUpdateEndTime));
 
       StatusAndCounts newStatusAndCounts = newStatusAndErrorCounts(rs);
+
+      if (rs.status() == CbasRunSetStatus.CANCELING) {
+        abortManager.abortRunSet(rs.runSetId());
```
salonishah11, line 107: I think calling `abortRunSet` here might be pre-mature. If all the Runs in the RunSet are in `Canceled` state then we don't need to call this function here. I think this should be called in the `else` section of `if` block on L119.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunSetsPoller.java
```diff
@@ -99,6 +102,30 @@ private RunSet updateRunSet(RunSet rs, OffsetDateTime limitedEndTime) {
...
+        List<Run> allRuns = runDao.getRuns(new RunDao.RunsFilters(rs.runSetId(), null));
+        int canceledRuns = 0;
+
+        // Check how many runs in the run set are canceled; increment.
+        for (Run run : allRuns) {
+          if (run.status() == CbasRunStatus.CANCELED) {
+            canceledRuns += 1;
+          }
+        }
```
salonishah11, line 116: Can this information be fetched from the database itself? I think the `getRunStatusCounts` can be used to filter by RunSet Id and Run status.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunSetsPoller.java
salonishah11: One thing to also note is that in `abortRunSet` method on L41 the status of RunSet is set to Canceling. This should only happen the first time when a request for aborting a RunSet comes in. We shouldn't reset it when its called from the SmartRunSetPoller as we will loose the original timestamp of when the abort request was made.

common/openapi.yml
kpierre13: To keep things concise, rather than try and return two different map types, I opted to change this back to a string, so I could return a Map<String List<String>> but this may change looking through more of the PR comments.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunSetsPoller.java
kpierre13: I don't think that I'm setting the runSet to cancelling again in the `SmartRunSetPoller`. I think I'm only setting the run set to a cancelling state when the request comes in. Is that right? Should I be looking somewhere else?

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -564,27 +564,12 @@ void testRunSetAbort() throws Exception {
...
-    when(runSetDao.updateStateAndRunDetails(
-            eq(returnedRunSet1Running.runSetId()),
-            eq(CbasRunSetStatus.CANCELING),
-            eq(2),
-            eq(0),
-            any()))
-        .thenReturn(1);
-
-    when(runDao.getRuns(any())).thenReturn(runs);
+    HashMap<String, List<String>> abortResults = new HashMap<>();
```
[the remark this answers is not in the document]
kpierre13: Yes, I've created that class and applied it everywhere where the Map was being called.

common/openapi.yml
kpierre13: With the latest push, this is now back to being a uuid

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunSetsPoller.java
salonishah11: I think [this L40](https://github.com/DataBiosphere/cbas/blob/5f3aca90a9e05a1c4a0ce1dc04b2a900f95dc4d2/service/src/main/java/bio/terra/cbas/runsets/monitoring/RunSetAbortManager.java#L40) inside the `abortManager.abortRunSet()`  sets the RunSet to `Canceling` state. When the abort request comes in, the RunSet state should be changed to `Canceling`. But in the SmartRunSetPoller, if the RunSet still has some runs to be cancelled, it calls the `abortManager.abortRunSet()` again which would set the RunSet state and timestamp to Cancelling again. Is my understanding of the code right?

kpierre13: Ah, my mistake. So I should make that check conditional, so it's not constantly being set to a 'CANCELING` state. Sorry for my confusion.

salonishah11 (approved):
💯

PR #164, opened 2023-05-18
WM-1987: Fix jersey memory leak

cjllanwarne:
Bonus extra fixes (see [slides](https://docs.google.com/presentation/d/1paOnzU0jjRjV1n133j0GA_M-wfEeFTPSbV9AcAFEwYQ/edit?usp=sharing)):
- Stops Relay/Listener RESETs which spammed the logs with exception stack traces, and caused WDS status to flip/flop.
- Retry logic around WDS functions, to recover in case the ProcessingException happens for some other reason (or the `Connection: close` header stops being respected by WDS in the future.


service/src/main/java/bio/terra/cbas/dependencies/common/ScheduledHealthChecker.java
```diff
@@ -35,7 +35,9 @@ public ScheduledHealthChecker(
             "leonardo", leonardoService);
   }
 
-  @Scheduled(fixedDelay = 5 * 60 * 1000)
+  @Scheduled(
+      fixedDelayString = "${cbas.scheduler.healthCheckIntervalSeconds}",
+      timeUnit = java.util.concurrent.TimeUnit.SECONDS)
```
cjllanwarne, line 40: This is a stowaway change to allow the health check interval to be configurable

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsClient.java
```diff
@@ -45,6 +48,7 @@ protected ApiClient getApiClient()
         "Authorization",
         "Bearer " + credentialLoader.getCredential(CredentialLoader.CredentialType.AZURE_TOKEN));
     apiClient.setDebugging(wdsServerConfiguration.debugApiLogging());
+    apiClient.setHttpClient(commonHttpClient);
```
cjllanwarne, line 51: This is the actual change which fixed the bug

service/src/main/java/bio/terra/cbas/dependencies/wds/ScheduledWdsClientRefresher.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.concurrent.Callable;
+import javax.ws.rs.client.Client;
+import org.databiosphere.workspacedata.client.ApiClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ScheduledWdsClientRefresher {
```
cjllanwarne, line 14: This allows us to schedule WDS client refreshes to avoid the listener force-closing the connection

service/src/main/java/bio/terra/cbas/dependencies/wds/ScheduledWdsClientRefresher.java
```diff
@@ -0,0 +1,88 @@
...
+
+  private final Object synchronizer = new Object();
+
+  private final Logger logger = LoggerFactory.getLogger(ScheduledWdsClientRefresher.class);
+
+  private final List<Client> retiredClients;
+  private final List<Client> expiredClients;
+
+  private Client currentClient;
```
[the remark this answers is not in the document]
pshapiro4broad: Oops, I see you have already made the scheduled tasks synchronized. That's good but I think you still have an issue with the getters since they're not synchronized. Maybe making a copy of the lists before returning them would be sufficient.

service/src/main/java/bio/terra/cbas/dependencies/wds/ScheduledWdsClientRefresher.java
```diff
@@ -0,0 +1,88 @@
...
+    } catch (Exception ex) {
+      logger.error("Error creating client for {}", clientName, ex);
+    }
+  }
+
+  @Scheduled(
+      fixedDelayString = "${cbas.scheduler.httpClientRefreshSeconds}",
+      timeUnit = java.util.concurrent.TimeUnit.SECONDS)
+  public void closeOldClients() {
+    synchronized (synchronizer) {
```
pshapiro4broad, line 56: Personally I would use method level synchronization here (and use `synchronized(this)` for the synchronized block above) rather than adding an explicit sync object. Is there a reason to use an explicit object here?

service/src/main/java/bio/terra/cbas/dependencies/wds/ScheduledWdsClientRefresher.java
```diff
@@ -0,0 +1,88 @@
...
+      timeUnit = java.util.concurrent.TimeUnit.SECONDS)
+  public void refreshClient() {
+    synchronized (synchronizer) {
+      logger.debug("Refreshing client for {} and marking current client as retired.", clientName);
+      if (currentClient != null) {
+        retiredClients.add(currentClient);
+      }
+    }
+    try {
+      currentClient = clientFactory.call();
```
pshapiro4broad, line 46: In this case it doesn't make a difference since you're calling the service on a timer for status checks but normally I would write this by dirtying the cache and then fill the cache on demand, rather than preemptively filling the cache.

service/src/main/java/bio/terra/cbas/dependencies/wds/ScheduledWdsClientRefresher.java
```diff
@@ -0,0 +1,88 @@
...
+      // Clear out the expired clients and move the retired clients to the expired list to be
+      // closed down next cycle:
+      expiredClients.clear();
+      expiredClients.addAll(retiredClients);
+      retiredClients.clear();
+    }
+  }
+
+  public Client getCurrentClient() {
+    return currentClient;
```
pshapiro4broad, line 78: How does this code prevent a client from being closed while it's still in use? Is the idea that the client refresh seconds is long enough that any api call must have already completed?

service/src/test/java/bio/terra/cbas/dependencies/wds/TestScheduledWdsClientRefresher.java
```diff
@@ -0,0 +1,88 @@
...
+  @Test
+  void buildsUpRetiredClients() {
+    ScheduledWdsClientRefresher scheduledWdsClientRefresher = new ScheduledWdsClientRefresher();
+
+    List<Integer> clientHashes = new ArrayList<>();
+
+    for (int i = 0; i < 10; i++) {
+      clientHashes.add(scheduledWdsClientRefresher.getCurrentClient().hashCode());
+      scheduledWdsClientRefresher.refreshClient();
+    }
```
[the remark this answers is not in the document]
pshapiro4broad: I think `generate()` could work here too since `i` is unneeded
```
    List<Integer> clientHashes = Stream.generate(() -> {
      int hash = scheduledWdsClientRefresher.getCurrentClient().hashCode()
      scheduledWdsClientRefresher.refreshClient();
      return hash;
    }).limit(10);
```

service/src/test/java/bio/terra/cbas/dependencies/wds/TestScheduledWdsClientRefresher.java
```diff
@@ -0,0 +1,88 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+public class TestScheduledWdsClientRefresher {
+
+  @Test
+  void canGetCurrentClient() {
+    ScheduledWdsClientRefresher scheduledWdsClientRefresher = new ScheduledWdsClientRefresher();
+    assert (scheduledWdsClientRefresher.getCurrentClient() != null);
```
pshapiro4broad, line 20: Unless you run with assertions enabled (and most terra code doesn't do this), `assert` doesn't do anything. See https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html#enable-disable. You need to use `assertTrue()` or (my preference) a hamcrest matcher
```suggestion
    assertThat(scheduledWdsClientRefresher.getCurrentClient(), nullValue());
```

service/src/main/java/bio/terra/cbas/dependencies/wds/ScheduledWdsClientRefresher.java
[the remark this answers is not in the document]
cjllanwarne: The lists are not intended to be accessed outside the class. Good point on adding synchronization around the `getCurrentClient`... though I'm not sure it would actually be possible for it to ever be in an invalid state.

service/src/test/java/bio/terra/cbas/dependencies/wds/TestScheduledWdsClientRefresher.java
cjllanwarne: 🤦 ... and even better, `assertNotNull`

service/src/test/java/bio/terra/cbas/runsets/monitoring/TestSmartRunsPollerFunctional.java
```diff
@@ -348,14 +349,16 @@ void updatingOutputFails_UpdatingDataTable() throws Exception {
...
+            new WdsServiceApiException(
+                new org.databiosphere.workspacedata.client.ApiException("Bad WDS update")));
     when(runsDao.updateRunStatusWithError(eq(runningRunId1), eq(SYSTEM_ERROR), any(), any()))
         .thenReturn(1);
     var actual = smartRunsPoller.updateRuns(List.of(runToUpdate1, runAlreadyCompleted));
 
     verify(cromwellService).runSummary(runningRunEngineId1);
     String expectedErrorMessage =
-        "Error while updating data table attributes for record %s from run %s (engine workflow ID %s): Bad WDS update"
+        "Error while updating data table attributes for record %s from run %s (engine workflow ID %s): WdsServiceApiException: Bad WDS update"
```
cjllanwarne, line 361: Maybe not necessary

service/src/main/java/bio/terra/cbas/retry/RetryLoggingListener.java
```diff
@@ -0,0 +1,36 @@
...
+@Component("retryLoggingListener")
+public class RetryLoggingListener implements RetryListener {
+
+  private final Logger logger = LoggerFactory.getLogger(getClass());
+
+  @Override
+  public <T, E extends Throwable> void close(
+      RetryContext context, RetryCallback<T, E> callback, Throwable throwable) {
+    if (context.getRetryCount() > 1) {
+      logger.debug("Retryable method closing ({}th retry)", context.getRetryCount());
```
mspector, line 19: nit: these `({}th retry)` messages will look funny for retry number 1, 2, and 3! I wouldn't block approval over this, but you could do `(retry number {})` or `(retry #{})`

mspector (approved):
thanks for the walkthrough!

kpierre13 (approved):

PR #165, opened 2023-05-24
[WM-1981] Support for Dockstore methods in POST /method endpoint

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/WM-1981


service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -243,21 +276,26 @@ public List<String> validateMethod(PostMethodRequest methodRequest) {
...
-
-        if (!doesUrlMatchPattern) {
-          errors.add("method_url is invalid. URL doesn't match pattern format");
-        } else if (!SUPPORTED_URL_HOSTS.contains(url.getHost())) {
-          errors.add("method_url is invalid. Supported URI host(s): " + SUPPORTED_URL_HOSTS);
+      // we only verify if URL is valid for GitHub source here. For Dockstore methods a workflow
+      // path which is not completely a valid URL is sent as method url, and it's validity is
+      // checked while fetching the raw GitHub url for the workflow path
+      if (methodRequest.getMethodSource() == PostMethodRequest.MethodSourceEnum.GITHUB) {
+        // verify that URL is valid, and it's host is supported
```
cjllanwarne, line 283:
```suggestion
        // verify that URL is valid, and its host is supported
```

cjllanwarne: Looks like this wasn't even your comment originally, so this is me being extra-double pedantic... 😆

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -84,19 +87,49 @@ public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMetho
...
+      return new ResponseEntity<>(new PostMethodResponse().error(errorMsg), HttpStatus.BAD_REQUEST);
+    } catch (bio.terra.dockstore.client.ApiException e) {
+      String errorMsg =
+          "Error while importing Dockstore workflow. Error: %s".formatted(e.getMessage());
+      log.error(errorMsg, e);
+      return new ResponseEntity<>(new PostMethodResponse().error(errorMsg), HttpStatus.BAD_REQUEST);
+    }
+
     // call Cromwell's /describe endpoint to get description of the workflow along with inputs and
     // outputs
```
cjllanwarne, line 122: TOL: This method is getting a little long and unwieldy. I'm finding it a little hard to keep track of everything that it's doing, and I think this kind of comment is also maybe a bit of a sign that we could break it down into a couple of sub-functions, to keep the top-level coordination function shorter and easier to read?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -171,6 +180,40 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+                      methodVersion.method().methodSource())),
+              methodVersion.name(),
+              dockstoreService);
+
+      // this could happen if there was no url or empty url received in the Dockstore workflow's
+      // descriptor response
+      if (rawMethodUrl == null || rawMethodUrl.isEmpty()) {
+        String errorMsg =
+            "Error while retrieving WDL url for Dockstore workflow. No workflow url found specified path.";
+        log.error(errorMsg);
```
cjllanwarne, line 201: Same comment about `log.error`

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -171,6 +180,40 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+        return new ResponseEntity<>(
+            new RunSetStateResponse().errors(errorMsg), HttpStatus.BAD_REQUEST);
```
cjllanwarne, line 203: If this does happen, do we really want to fail the request, or should we instead consider creating a runset which is immediately in the `Failed` state (so that we keep the attempt and the error in the run history?)

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -171,6 +180,40 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+        | MalformedURLException
+        | bio.terra.dockstore.client.ApiException e) {
+      // the flow shouldn't reach here since if it was invalid URL it should have been caught in
+      // when method was imported
+      String errorMsg =
+          "Something went wrong while submitting workflow. Error: %s".formatted(e.getMessage());
+      log.error(errorMsg, e);
+      return new ResponseEntity<>(
+          new RunSetStateResponse().errors(errorMsg), HttpStatus.INTERNAL_SERVER_ERROR);
+    }
```
cjllanwarne, line 215: And again here

cjllanwarne: Although... given the "should have been caught in when method was imported" comment, maybe we _DO_ want to make this a more strong type or error log...

service/src/main/java/bio/terra/cbas/dependencies/dockstore/DockstoreService.java
```diff
@@ -0,0 +1,25 @@
...
+  private final DockstoreClient dockstoreClient;
+  private static final String WDL_TYPE = "WDL";
+  private static final String WORKFLOW_IDENTIFIER = "#workflow/";
+
+  public DockstoreService(DockstoreClient dockstoreClient) {
+    this.dockstoreClient = dockstoreClient;
+  }
+
+  private Ga4Ghv1Api ga4Ghv1Api() {
+    return new Ga4Ghv1Api(dockstoreClient.getApiClient());
```
cjllanwarne, line 19: In the other dependencies I think we put these API getters into the equivalent of `DockstoreClient.java`. Not sure whether that makes much difference... but I guess it might make dependency injection harder in certain test cases?

service/src/test/java/bio/terra/cbas/app/HelloWorldTest.java
```diff
@@ -1,13 +0,0 @@
-package bio.terra.cbas.app;
-
-import static org.junit.jupiter.api.Assertions.assertEquals;
-
-import org.junit.jupiter.api.Test;
-
-/** Placeholder test case to prove the github action integration. */
-class HelloWorldTest {
```
cjllanwarne, line 8: Goodbye World! 👋 🌎

service/src/test/java/bio/terra/cbas/common/TestMethodUtil.java
```diff
@@ -0,0 +1,122 @@
...
+import bio.terra.dockstore.model.ToolDescriptor;
+import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ContextConfiguration;
+
+@WebMvcTest
```
cjllanwarne, line 26: These tests look more like straight unit tests than MVC (API-level) tests?

service/src/test/java/bio/terra/cbas/common/TestMethodUtil.java
```diff
@@ -0,0 +1,122 @@
...
+@ContextConfiguration(classes = MethodsApiController.class)
+class TestMethodUtil {
+  @MockBean private CromwellService cromwellService;
+  @MockBean private DockstoreService dockstoreService;
+  @MockBean private MethodDao methodDao;
+  @MockBean private MethodVersionDao methodVersionDao;
+  @MockBean private RunSetDao runSetDao;
```
cjllanwarne, line 33: These don't look like they're used as beans in this test? The look like they could be mocked/stubbed/null'd out per test, according to need?

service/src/test/java/bio/terra/cbas/common/TestMethodUtil.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ContextConfiguration;
+
+@WebMvcTest
+@ContextConfiguration(classes = MethodsApiController.class)
```
cjllanwarne, line 27: Also not sure this is used?

service/src/test/java/bio/terra/cbas/common/TestMethodUtil.java
```diff
@@ -0,0 +1,122 @@
...
+class TestMethodUtil {
```
cjllanwarne, line 28: I'd strongly encourage adding the `@ExtendWith(MockitoExtension.class)` annotation to this class. Amongst other things, it checks (a) did all my mocked out methods actually get used, and (b) did any unmocked-out methods get called unexpectedly

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -44,6 +47,7 @@ class TestMethodsApiController {
   private static final String API = "/api/batch/v1/methods";
 
   @MockBean private CromwellService cromwellService;
+  @MockBean private DockstoreService dockstoreService;
```
cjllanwarne, line 50: I'd also suggest adding `@ExtendWith(MockitoExtension.class)` to this test class

(you didn't change the class signature line, so this comment is just on the first changed line in the file)

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -213,7 +217,7 @@ void returnErrorForInvalidPostRequest() throws Exception {
       }
       """;
     String expectedError =
-        "Bad user request. Error(s): method_name is required. method_source is required and should be one of: [GitHub]. method_version is required. method_url is invalid. Supported URI host(s): [raw.githubusercontent.com]";
+        "Bad user request. Error(s): method_name is required. method_source is required and should be one of: [GitHub, Dockstore]. method_version is required";
```
cjllanwarne, line 220: Do we want to add a test for a Github type workflow with an invalid github url to make sure that part of the message is still shown sometimes?

cjllanwarne (commented):
I'm a little inclined to agree with the sonar code smell:
> Refactor this method to reduce its Cognitive Complexity from 18 to the 15 allowed

dockstore-client/swagger.gradle
```diff
@@ -0,0 +1,49 @@
...
+
+    implementation 'io.swagger.core.v3:swagger-annotations'
+    swaggerCodegen 'io.swagger.codegen.v3:swagger-codegen-cli'
+}
+
+def artifactGroup = "${group}.dockstore"
+
+tasks.register('downloadDockstoreYaml', Download) {
+    // this was the latest git commit hash at development time
+    src "https://raw.githubusercontent.com/dockstore/dockstore/bff1a2e0459f2d95faf0c8acfa575cf41e46e342/dockstore-webservice/src/main/resources/openapi3/openapi.yaml"
```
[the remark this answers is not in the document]
salonishah11: I actually used this link so that we can pin the version for openapi.yaml file. It seems there is an openapi.yaml hosted at `https://dockstore.org/openapi.yaml`. Would that be better to use?

service/src/main/java/bio/terra/cbas/common/MethodUtil.java
```diff
@@ -0,0 +1,42 @@
...
+import bio.terra.dockstore.model.ToolDescriptor;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.List;
+
+public final class MethodUtil {
+  private static final String GITHUB_URL_HOST = "github.com";
+  private static final String RAW_GITHUB_UTL_HOST = "raw.githubusercontent.com";
```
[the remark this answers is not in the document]
salonishah11: Oops my bad

service/src/main/java/bio/terra/cbas/config/DockstoreServerConfiguration.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.cbas.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "dockstore")
+public record DockstoreServerConfiguration(String baseUri) {
+
+  public String getBaseUri() {
+    return baseUri;
+  }
```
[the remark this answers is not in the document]
salonishah11: Oh I didn't know that. TIL. I will remove this explicit getter method. Thank you!

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -84,19 +87,49 @@ public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMetho
...
+      }
+    } catch (URISyntaxException | MalformedURLException e) {
+      String errorMsg =
+          "Bad user request. Method url has invalid value. Error: %s".formatted(e.getMessage());
+      log.error(errorMsg, e);
+      return new ResponseEntity<>(new PostMethodResponse().error(errorMsg), HttpStatus.BAD_REQUEST);
+    } catch (bio.terra.dockstore.client.ApiException e) {
+      String errorMsg =
+          "Error while importing Dockstore workflow. Error: %s".formatted(e.getMessage());
+      log.error(errorMsg, e);
```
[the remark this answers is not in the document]
salonishah11: Yes I have updated this and at the other places where its user error to be `log.warn` instead

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
salonishah11: I had only touched the code inside this block to replace the method.url with the rawUrl. But I can try and refactor this method to make it more readable 👍

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Hmm that's a good question. I think if it is a bad request / user error (which it is in this case) we should not create a RunSet in failed state. The request was not valid and hence we should only fail the request. We are logging the error from the request so there is a trail for the request. What do you think?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: I would prefer to leave this as `log.error` so that we are alerted about the flow reaching this point?

service/src/main/java/bio/terra/cbas/dependencies/dockstore/DockstoreService.java
salonishah11: I was following what we did for [Leonardo AppsV2Api](https://github.com/DataBiosphere/cbas/blob/dbcc106e4edfbd9daad116a8d64a8f6afd7d9d3a/service/src/main/java/bio/terra/cbas/dependencies/leonardo/LeonardoService.java#L27). Should I be looking at some other service class?

service/src/test/java/bio/terra/cbas/common/TestMethodUtil.java
salonishah11: Without the `@WebMvcTest` annotation the tests were failing as it was receiving the `dockstoreService` mock bean as null. But I am not very familiar with Spring magic and maybe there is something else that can be done to fix the failures instead of adding this particular annotation?

service/src/test/java/bio/terra/cbas/common/TestMethodUtil.java
salonishah11: I added this because the tests were failing with `Failed to load ApplicationContext` error after adding `@WebMvcTest` annotation. Maybe there is some Spring magic that I can use but I just don't know about? Do you know what I can do to fix the `dockstoreService` being null error without introducing the `@WebMvcTest` annotation?

service/src/test/java/bio/terra/cbas/common/TestMethodUtil.java
salonishah11: I get the same `Failed to load ApplicationContext` error if I don't add this annotation 🤔

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
salonishah11: I have added another test to cover case of invalid method url with unsupported host

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
salonishah11: I have refactored this method a bit. I think because of various exceptions handlings it is a little verbose but maybe its little more readable now?

dockstore-client/swagger.gradle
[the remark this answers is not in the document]
cjllanwarne: Hmm, I'm not sure. Something fixed feels better than something floating. Something we keep as a copy feels _best_, if we check it into the repo ourselves, since we know it will never change... is that an option?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
cjllanwarne: Two thoughts:

1. From a user's point of view, I don't know how visible logs are going to be.
2. When we switch from "do everything synchronously in the user request" to "store in the DB and process later", we'll almost certainly have records of things like this against failed runsets. But I don't know that we need to be too worried about how we handle it right now. As long as the user can see the error message easily when they need to, I think it's ok.

service/src/main/java/bio/terra/cbas/dependencies/dockstore/DockstoreService.java
cjllanwarne: Oh, ok then, never mind! 😆

service/src/test/java/bio/terra/cbas/common/TestMethodUtil.java
cjllanwarne: Since you're not using this for MVC testing, you don't need to worry about indirectly injecting mocks into the spring context. You can just provide them directly when you call the methods. So, I think you can remove this annotation _and_ the `@MockBean`s below, and when you need a `dockstoreService` object for one of your tests, just create one and configure it as you need it

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -102,19 +102,19 @@ public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMetho
...
       String errorMsg =
           "Bad user request. Method url has invalid value. Error: %s".formatted(e.getMessage());
-      log.error(errorMsg, e);
+      log.warn(errorMsg, e);
       return new ResponseEntity<>(new PostMethodResponse().error(errorMsg), HttpStatus.BAD_REQUEST);
     } catch (bio.terra.dockstore.client.ApiException e) {
       String errorMsg =
           "Error while importing Dockstore workflow. Error: %s".formatted(e.getMessage());
-      log.error(errorMsg, e);
+      log.warn(errorMsg, e);
```
cjllanwarne, line 117: This kind of thing still feels like a metric to track it would be super helpful.

cjllanwarne (commented):
Looks good, with one request for dockstore error metrics, and I think you can make the unit tests work without the mock mvc framework.

service/src/main/java/bio/terra/cbas/dependencies/dockstore/DockstoreClient.java
```diff
@@ -0,0 +1,24 @@
...
+import bio.terra.cbas.config.DockstoreServerConfiguration;
+import bio.terra.dockstore.client.ApiClient;
+import javax.ws.rs.client.Client;
+import org.springframework.stereotype.Component;
+
+@Component
+public class DockstoreClient {
+
+  private final DockstoreServerConfiguration dockstoreServerConfiguration;
+  private final Client commonHttpClient = new ApiClient().getHttpClient();
```
salonishah11, line 12: Follow same convention of using singleton client. See [this PR](https://github.com/DataBiosphere/cbas/pull/164/files#diff-0e12086dab69c580ca388826597f9e392e5f42410050a69151ed5a9002164831).

dockstore-client/swagger.gradle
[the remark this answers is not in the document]
salonishah11: Yeah I think checking it into the cbas repo should also work. It seems kind of best of both worlds for now so I will do that.

service/src/test/java/bio/terra/cbas/common/TestMethodUtil.java
salonishah11: Ah I see. Okay. So I have mocked out the `DockstoreService` class and removed all the unnecessary annotations 👍

dockstore-client/swagger.gradle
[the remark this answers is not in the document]
salonishah11: I have checked-in the dockstore openapi.yaml file.

service/src/main/java/bio/terra/cbas/dependencies/dockstore/DockstoreClient.java
salonishah11: Updated the code

salonishah11:
Added metric counter for error when importing dockstore method:

Screenshot 2023-06-01 at 2 41 53 PM

service/build.gradle
```diff
@@ -80,8 +80,7 @@ task verifyPacts( type: Test ) {
     filter {
         includeTestsMatching "*VerifyPacts*"
     }
-
-}
+ }
```
cjllanwarne, line 83: Is this right?

cjllanwarne (approved):

service/build.gradle
salonishah11: I think this was automatic based on the linter/spotless. I didn't manually make this change. Would you like me to revert it?

mspector: just for my own learning... does that mean that linter or spotless identified the `dockstore-client` directory with the openAPI YAML in it, and added this line as a result?

salonishah11: L38 was actually manually added by me. The removal of new line on L81 was automatic by spotless I think.

mspector: I see - thanks for clarifying

PR #166, opened 2023-05-25
[WM-2009] abort endpoint pact test

kpierre13:


cjllanwarne (approved):

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsRunSetsApiController.java
```diff
@@ -184,4 +188,48 @@ public void runSetsData() throws Exception {
...
+            null,
+            null);
+
+    AbortRequestDetails abortDetails = new AbortRequestDetails();
+    abortDetails.setFailedIds(List.of());
+    abortDetails.setSubmittedIds(List.of(runToBeCancelled.runId()));
+
+    when(runSetDao.getRunSet(runSetId)).thenReturn(runSetToBeCancelled);
+    when(runDao.getRuns(new RunDao.RunsFilters(runSetId, any())))
+        .thenReturn(Collections.singletonList(runToBeCancelled));
```
cjllanwarne, line 231: TOL: If these are always returning the "to be cancelled" version of the run and run set, then what sets the status to `CANCELING` to satisfy the expected response on the client side of the pact?

cjllanwarne (approved):

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsRunSetsApiController.java
```diff
@@ -184,4 +188,48 @@ public void runSetsData() throws Exception {
     when(smartRunSetsPoller.updateRunSets(response))
         .thenReturn(new TimeLimitedUpdater.UpdateResult<>(response, 1, 1, true));
   }
+
+  @State({"a run set with UUID 20000000-0000-0000-0000-000000000002 exists"})
```
mspector, line 192: we had talked about consolidating the run and run set setup into a single state function, but actually, I think that is not as straightforward as I thought it would be -- so I think we can leave figuring that out for another time.

mspector (approved):
nice work!

PR #169, opened 2023-05-31
[WM-1992] add statement to README.md and DEVELOPMENT.md about including Jira IDs in PR titles

mspector:


kpierre13 (approved):

DEVELOPMENT.md
```diff
@@ -2,6 +2,11 @@
 
 TODO: Complete me! Should be of similar quality to the ECM documentation.
 
+## Contributing
+
+For compliance reasons, all pull requests **must** be submitted with a Jira ID in the pull request title.
+Include the Jira ID at the beginning of the title, enclosed in square brackets.
```
salonishah11, line 8: small nit: maybe we can remove the requirement of "enclosed in square brackets". I don't think that's mandatory no?

README.md
```diff
@@ -1,5 +1,11 @@
 # Composite Batch Analysis Service (C-BAS)
 
+## Contributing
+
+For compliance reasons, all pull requests **must** be submitted with a Jira ID in the pull request title.
+Include the Jira ID at the beginning of the title, enclosed in square brackets.
```
salonishah11, line 6: Same comment here

DEVELOPMENT.md
cjllanwarne: Yeah I never put them in square brackets 😆

cjllanwarne: I'd also say, if there is more than one relevant ticket, include them all, eg `WM-1997, WM-2002, WM-2005: Fix for many bugs with the same root cause`, etc.

cjllanwarne (approved):
LGTM, with a few minor suggestions

PR #174, opened 2023-06-21
[WM-2038] fix pact verification

mspector:
I've annotated this PR more heavily than usual, because there are multiple interconnected issues involved. Please let me know if you have questions.

Issues covered:
- Pacts that were previously unpublished (due to now-fixed problems with CBAS-UI's CI) were revealed to be failing - those have been fixed.
- CBAS should only verify with its main branch
- MethodSource error message was unclear - now fixed
- Verification class now covers all controllers, not just the RunSetsApiController


.github/workflows/build-and-test.yml
```diff
@@ -104,19 +104,6 @@ jobs:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Configure pact
```
mspector, line 107: I'm moving the pact verification step(s) from the `build-and-test` action so that they only run on merges to main.

.github/workflows/verify-pacts-main.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Build all projects without running tests
+        run: ./gradlew --build-cache build -x test
+
+  verify-pacts:
+    needs: [ build ]
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
```
mspector, line 30: I don't think postgres should be needed, but I'm leaving it in for now because we will need to test this action after this PR gets merged. In a nutshell, the reason it has to be merged first is that the GitHub API won't let us trigger this action on *any* branch until a version of it is merged into `main`. So once we merge this, we can iterate in a new PR.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -189,13 +188,15 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+          PostMethodRequest.MethodSourceEnum.fromValue(methodVersion.method().methodSource());
+
+      if (methodSourceEnum == null) {
+        throw new UnknownMethodSourceException(methodVersion.method().methodSource());
+      }
       rawMethodUrl =
           MethodUtil.convertToRawUrl(
-              methodVersion.url(),
-              Objects.requireNonNull(
-                  convertToMethodSourceEnum(methodVersion.method().methodSource())),
```
mspector, line 196: a bad `MethodSource` field was the root cause of most of the verification failures. It showed up as a vague null pointer exception, which was thrown by the above `requireNonNull` call.

I'm replacing this `requireNonNull` with the `if` statement above that throws the more specific `UnknownMethodSourceException`

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
```diff
@@ -45,13 +46,18 @@
...
+import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.web.servlet.MockMvc;
 
 @WebMvcTest
-@ContextConfiguration(classes = {RunSetsApiController.class, CbasApiConfiguration.class})
+@ContextConfiguration(
+    classes = {RunSetsApiController.class, MethodsApiController.class, CbasApiConfiguration.class})
+@TestPropertySource(properties = "cbas.cbas-api.runSetsMaximumRecordIds=100")
+@TestPropertySource(properties = "cbas.cbas-api.maxWorkflowInputs=100")
+@TestPropertySource(properties = "cbas.cbas-api.maxWorkflowOutputs=40")
```
mspector, line 57: these configurations were needed before, but their absence was masked by the fact that none of the pacts that need them were ever being verified.

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
```diff
@@ -45,13 +46,18 @@
...
 @Provider("cbas")
 @PactBroker()
-class VerifyPactsRunSetsApiController {
+class VerifyPactsAllControllers {
```
mspector, line 60: I couldn't find a way to effectively separate these tests into one class per controller - and I'm not sure if that would have benefitted us much anyway. Instead, this file/class has been renamed to reflect the fact that it verifies *all* controllers.

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
```diff
@@ -107,7 +113,7 @@ public void initializeDAO() throws Exception {
                 "myMethod description",
                 OffsetDateTime.now(),
                 methodVersionUUID,
-                "myMethod source"),
+                "GitHub"),
```
mspector, line 116: this was the cause of the null pointer exception (see comment above)

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
```diff
@@ -123,6 +129,13 @@ public void initializeDAO() throws Exception {
     when(runDao.createRun(any())).thenReturn(1);
   }
 
+  @State({"cromwell initialized"})
```
mspector, line 132: I think we should try to manage our states in a more modular way - this is the start of an attempt to do so

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
mspector: from @salonishah11 : can we just use the methodSource enum instead of a raw string?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
mspector: we need to preserve the use of `convertToMethodSource`

dillydally414: What happened to the explicit `UnknownMethodSourceException` throw?

dillydally414 (approved):
Small question regarding the NullPointer fix (but it seems to be fine given the tests are passing)

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
mspector: turns out `convertToMethodSource` throws it! I didn't realize that before

kpierre13 (approved):

PR #176, opened 2023-06-30
[WM-2010] Provider side contract test for POST /methods endpoint

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/WM-2010

### DIFF
--- a/service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
+++ b/service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
@@ -21,7 +21,6 @@ import bio.terra.cbas.model.MethodLastRunDetails;
 import bio.terra.cbas.model.MethodListResponse;
 import bio.terra.cbas.model.PostMethodResponse;
 import bio.terra.cbas.models.*;
-import bio.terra.cbas.util.UuidSource;
 import bio.terra.dockstore.model.ToolDescriptor;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import cromwell.client.model.WorkflowDescription;
@@ -58,7 +57,6 @@ class TestMethodsApiController {
   @MockBean private MethodDao methodDao;
   @MockBean private MethodVersionDao methodVersionDao;
   @MockBean private RunSetDao runSetDao;
-  @MockBean private UuidSource uuidSource;
 
   // This mockMVC is what we use to test API requests and responses:
   @Autowired private MockMvc mockMvc;
@@ -434,8 +432,6 @@ class TestMethodsApiController {
     when(cromwellService.describeWorkflow(validRawWorkflow))
         .thenReturn(workflowDescForValidWorkflow);
 
-    when(uuidSource.generateUUID()).thenReturn(UUID.randomUUID()).thenReturn(UUID.randomUUID());
-
     MvcResult response =
         mockMvc
             .perform(
@@ -467,8 +463,6 @@ class TestMethodsApiController {
         .thenReturn(mockToolDescriptor);
     when(cromwellService.describeWorkflow(validRawWorkflow))
         .thenReturn(workflowDescForValidWorkflow);
-
-    when(uuidSource.generateUUID()).thenReturn(UUID.randomUUID()).thenReturn(UUID.randomUUID());
 
     MvcResult response =
         mockMvc
--- a/service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
+++ b/service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
@@ -21,6 +21,7 @@ import bio.terra.cbas.model.MethodLastRunDetails;
 import bio.terra.cbas.model.MethodListResponse;
 import bio.terra.cbas.model.PostMethodResponse;
 import bio.terra.cbas.models.*;
+import bio.terra.cbas.util.UuidSource;
 import bio.terra.dockstore.model.ToolDescriptor;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import cromwell.client.model.WorkflowDescription;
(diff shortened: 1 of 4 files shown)


dillydally414 (approved):

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -464,6 +468,8 @@ void validDockstoreMethodPostRequest() throws Exception {
     when(cromwellService.describeWorkflow(validRawWorkflow))
         .thenReturn(workflowDescForValidWorkflow);
 
+    when(uuidSource.generateUUID()).thenReturn(UUID.randomUUID()).thenReturn(UUID.randomUUID());
```
cjllanwarne, line 471: If this is fine being random, I would imagine we don't need to overwrite the default implementation here?

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
```diff
@@ -76,6 +76,20 @@ class VerifyPactsAllControllers {
   // This mockMVC is what we use to test API requests and responses:
   @Autowired private MockMvc mockMvc;
 
+  // mock objects
+  UUID fixedMethodVersionUUID = UUID.fromString("90000000-0000-0000-0000-000000000009");
```
cjllanwarne, line 80: Just checking... is this assuming the current prestaged database state? Or just a very similar looking UUID?

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
```diff
@@ -248,4 +259,24 @@ public void postAbort() throws Exception {
...
+  public HashMap<String, String> setFixedMethodId() throws Exception {
+    when(uuidSource.generateUUID()).thenReturn(fixedMethodUUID);
+
+    // This value is returned so that it can be injected into variables in the Pact(s)
+    HashMap<String, String> providerStateValues = new HashMap<>();
+    providerStateValues.put("method_id", fixedMethodUUID.toString());
+    return providerStateValues;
+  }
+
+  @State({"returns a fixed run set ID"})
```
cjllanwarne, line 273: This feels a little like we're maybe misusing the whole pact `state` concept a little bit?

cjllanwarne (approved):
LGTM with a little food for thought thrown in

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
mspector: If I understand correctly, there's no database being used during pact verification. It's probably worth changing this UUID to one that is more obviously different

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
mspector: I suggested this change, because I thought it was a better use of state! I'm curious to hear your perspective, I'm not absolutely confident about this

salonishah11: From what I understood, in the provider side of pact we setup the state for the contract test to run. Since the consumer side [expects method id and run set id in the response](https://github.com/DataBiosphere/cbas-ui/blob/160454ace85d48024adfa9cd8856bfeb965e376f/src/libs/ajax.test.js#L321-L322), we fix them here and set it in the provider state

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
salonishah11: I think this is just a very similar looking ID. It is fixed at `90000000-0000-0000-0000-000000000009` because that's what the consumer tests expect in the response. For example [here](https://github.com/DataBiosphere/cbas-ui/blob/160454ace85d48024adfa9cd8856bfeb965e376f/src/libs/ajax.test.js#L128).

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
salonishah11: I added this because without this the test was receiving null for `uuidSource`. Is there something else I should be doing?

service/src/test/java/bio/terra/cbas/controllers/VerifyPactsAllControllers.java
salonishah11: Actually putting my understanding in words made me think if we need to fix the method id and run set id at all? It seems the consumer side only checks that the method_id and run_set_id in the response are non-null which would already happen when a post /method endpoint request is made. So maybe we don't need to fix these ids at all? 🤔

mspector: I think there's a decision here for us to make. The reason for fixing the IDs is so that we can verify that the response contains the ID that was generated. It guards against a case where e.g. a random run set ID is generated, but a different UUID is returned.

We can make this setup simpler by only checking that the returned UUID is non-null (or that it matches a UUID regex). I don't have a strong opinion, but I wanted to point out that the options are subtly different

mspector (approved):

cjllanwarne (approved):

salonishah11:
I re-ran the Verify Pacts after the cbas-ui side of changes were merged. The new verification runs on the [6135c2837a4ea0565964f3b43f658a365216667d](https://github.com/DataBiosphere/cbas-ui/commit/6135c2837a4ea0565964f3b43f658a365216667d) commit which is the version bump that occurred after the [cbas-ui PR was merged](https://github.com/DataBiosphere/cbas-ui/pull/194). See below screenshot.

Screenshot 2023-07-10 at 9 12 31 AM

PR #180, opened 2023-07-13
[WX-1193] Add Call Caching to RunSet

THWiseman:
This PR updates RunSets to include an additional boolean parameter `call_caching_enabled`. CBAS will use this to parameterize the workflow options it sends to Cromwell. 

If `call_caching_enabled` isn't provided, CBAS will default to `false`. This is consistent with GCP behavior.


service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -376,6 +379,14 @@ public static List<String> validateRequestInputsAndOutputs(
...
   }
 
+  public static List<String> validateRequestContainsCallCachingParameter(RunSetRequest request) {
+    List<String> errorList = new ArrayList<>();
+    if (request.isCallCachingEnabled() == null) {
+      errorList.add("Request must contain boolean call_caching_enabled parameter");
+    }
+    return errorList;
+  }
+
```
aednichols, line 389: Mild preference for backwards compatible API that defaults to... on or off, idk

THWiseman: Defaulted to `off` as we do for GCP.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -128,6 +127,35 @@ public static String getErrorMessage(List<FailureMessage> failureMessages) {
     }
   }
 
+  /**
+   * Cromwell accepts an object "Workflow Options" to specify additional configuration for a
+   * workflow. Here, we build that object with the parameters we care about. final_workflow_log_dir
+   * specifies the path where outputs will be written. write_to_cache and read_from_cache are both
+   * related to call caching. When a user enables call caching, these should both be set to true.
```
salonishah11, line 134: Clarifying question: wouldn't setting `write_to_cache` to false confuse users that come from GCP Terra? I think in GCP Terra Rawls always sets `write_to_cache` to true and the call caching checkbox in UI refers to reading from call cache. So when unchecking the call caching box in Terra Azure they would assume that the `write_to_cache` will still be enabled even if read is disabled?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -128,6 +127,35 @@ public static String getErrorMessage(List<FailureMessage> failureMessages) {
...
+   * @return A string formatted as a JSON object that can be used as cromwell's Workflow Options.
+   * @throws JsonProcessingException
+   */
+  public static String buildWorkflowOptionsJson(
+      Optional<String> finalWorkflowLogDir, Boolean isCallCachingEnabled)
+      throws JsonProcessingException {
+    if (isCallCachingEnabled == null) {
+      log.warn("Null call caching parameter provided. Defaulting to false.");
+      isCallCachingEnabled = false;
+    }
```
salonishah11, line 148: Maybe this check can be moved to RunSetsApiController since the field is associated with RunSet and additionally this doesn't get checked for all the workflows in the RunSet?

salonishah11: Additionally maybe `buildWorkflowOptionsJson` can also be refactored to run once per RunSet if the workflow options are going to be same for all workflows in the RunSet?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
THWiseman: This is a good catch. I think you're right that it makes sense to always `write_to_cache`. I'll make that swap.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
THWiseman: Makes sense to me, happy to make that change.

common/openapi.yml
```diff
@@ -249,6 +249,10 @@ components:
           type: string
           format: uuid
           description: ID of the method version to run
+        call_caching_enabled:
+          type: boolean
+          default: false
```
aednichols, line 254: Default should be `true` IMO. Yes, we could default the GUI checkbox to on and have it override the `false`, but it just seems easier to have the the whole stack swimming in the same direction.

(I think in a DM I even suggested we should not expose the ability to disable call caching, because it overshadows [superior solutions like `volatile`; but that's a topic for the GUI/another day.](https://cromwell.readthedocs.io/en/stable/optimizations/VolatileTasks/))

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -128,6 +126,46 @@ public static String getErrorMessage(List<FailureMessage> failureMessages) {
...
+
+    // Path for cromwell to write workflow logs to.
+    Optional<String> finalWorkflowLogDir = cromwellClient.getFinalWorkflowLogDirOption();
+    if (finalWorkflowLogDir.isPresent()) {
+      workflowOptions.put("final_workflow_log_dir", finalWorkflowLogDir.get());
+    }
+
+    // Call caching fields
+    workflowOptions.put("write_to_cache", true);
+    workflowOptions.put("read_from_cache", isCallCachingEnabled);
```
aednichols, line 155: It seems like surely there is some way to use a `Map` literal as the basis for  the JSON, instead of side-effecting.

aednichols (approved):
Seems fine to me, feel free to take or leave my curmudgeonly opinions 🙂

salonishah11 (approved):

PR #181, opened 2023-07-27
[WM-2146] Logic for CBAS to find right Cromwell

kpierre13:


service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
```diff
@@ -1,34 +1,69 @@
...
+  private ApiClient getApiClient()
+      throws DependencyNotAvailableException, AzureAccessTokenException {
+    String uri;
+
+    String baseUriFromConfig = cromwellServerConfiguration.baseUri();
+    if (!baseUriFromConfig.isEmpty()) {
+      uri = baseUriFromConfig;
+    } else {
+      uri = dependencyUrlLoader.loadDependencyUrl(DependencyUrlLoader.DependencyUrlType.CROMWELL);
+    }
```
cjllanwarne, line 43: So this is going to be a little different from the other lookups.

For _read_ operations we will always want to use the URL in config, for _write_ operations we will always want to look up the right per-user cromwell instance.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
```diff
@@ -1,34 +1,69 @@
...
+
+    ApiClient apiClient = new ApiClient().setBasePath(uri);
+    apiClient.setHttpClient(singletonHttpClient);
+    apiClient.addDefaultHeader(
+        "Authorization",
+        "Bearer " + credentialLoader.getCredential(CredentialLoader.CredentialType.AZURE_TOKEN));
+    // By closing the connection after each request, we avoid the problem of the open connection
+    // being force-closed ungracefully by the Azure Relay/Listener infrastructure:
+    apiClient.addDefaultHeader("Connection", "close");
```
cjllanwarne, line 52: I think we only need this for the writer/external Cromwell instance... but I guess leaving it there for the local/reader instance is fine too

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -150,9 +154,7 @@ public String buildWorkflowOptionsJson(Boolean isCallCachingEnabled) {
 
     // Path for cromwell to write workflow logs to.
     Optional<String> finalWorkflowLogDir = cromwellClient.getFinalWorkflowLogDirOption();
-    if (finalWorkflowLogDir.isPresent()) {
-      workflowOptions.put("final_workflow_log_dir", finalWorkflowLogDir.get());
-    }
+    finalWorkflowLogDir.ifPresent(s -> workflowOptions.put("final_workflow_log_dir", s));
```
cjllanwarne, line 157: I like this change, but it seems slightly odd to see it happening in the middle of this PR? 😆

service/src/main/resources/application.yml
```diff
@@ -69,7 +69,7 @@ wds:
 
 leonardo:
   baseUri: "https://leonardo.dsde-dev.broadinstitute.org/"
-  wdsAppTypeNames: ['WDS', 'CROMWELL']
+  appTypeNames: ['WDS', 'CROMWELL']
```
cjllanwarne, line 72: So this was originally an order of preference for "which app do we look for to represent WDS", instead of just "what apps do we know about". We probably will want something similar for CROMWELL when we migrate from CROMWELL to (WORKFLOWS + CROMWELL_RUNNER). If someone is in the "old" world they will want CROMWELL, and new world will want CROMWELL_RUNNER.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
kpierre13: It should be easy enough to move it into a writer function

service/src/main/resources/application.yml
kpierre13: So in the "new" world, these appTypes will be `CROMWELL_RUNNER, WORKFLOWS, CROMWELL` only?

service/src/main/java/bio/terra/cbas/dependencies/leonardo/AppUtils.java
```diff
@@ -34,21 +35,25 @@ public AppUtils(
...
       return appTypeScoreA - appTypeScoreB;
     }
 
-    // Second criteria: Prefer apps with the WDS name
+    // Second criteria: Prefer apps with the expected app type name name
     int nameScoreA =
-        Objects.equals(a.getAppName(), "wds-%s".formatted(wdsServerConfiguration.instanceId()))
+        Objects.equals(
+                a.getAppName(),
+                "%s-%s".formatted(a.getAppType(), wdsServerConfiguration.instanceId()))
```
salonishah11, line 50: I think the `"%s-%s".formatted(a.getAppType(), wdsServerConfiguration.instanceId())` check might not work for Cromwell apps because for Cromwell/CBAS apps, the name of app follows the pattern `terra-app-<random-uuid>`. It might be worth testing out these changes against an a running Cromwell app in Terra dev.

salonishah11 (changes requested):

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -504,6 +506,14 @@ private List<RunStateResponse> buildInputsAndSubmitRun(
...
             storeRun(runId, null, runSet, record.getId(), SYSTEM_ERROR, e.getMessage()));
+      } catch (AzureAccessTokenException e) {
+        log.warn("Unrecognized token", e);
+        runStateResponseList.add(
+            storeRun(runId, null, runSet, record.getId(), SYSTEM_ERROR, e.getMessage()));
+      } catch (DependencyNotAvailableException e) {
+        log.warn("Dependency not available", e);
+        runStateResponseList.add(
+            storeRun(runId, null, runSet, record.getId(), SYSTEM_ERROR, e.getMessage()));
       }
```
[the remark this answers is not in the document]
kpierre13: I had them as 3 different catch blocks since the `log.warn` message was different amongst them all. Would it be cleaner to combine into one catch block and change the message based on which exception is being thrown?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
kpierre13: Not quite sure where this change actually came from myself!

dillydally414 (approved):

service/src/main/java/bio/terra/cbas/dependencies/leonardo/AppUtils.java
```diff
@@ -33,26 +34,41 @@ public AppUtils(
   }
 
   int appComparisonFunction(ListAppResponse a, ListAppResponse b) {
```
mspector, line 36: is this function still being used?

kpierre13: Currently no, I left it thinking that it may be able to be reused in a following ticket. It can be removed for clarity, however.

service/src/main/java/bio/terra/cbas/dependencies/leonardo/AppUtils.java
```diff
@@ -93,24 +93,27 @@ ListAppResponse findBestAppForWds(List<ListAppResponse> apps)
...
                     logger.info(
-                        "Not using app {} for WDS because it is in workspace {}, not {}",
+                        "Not using app {} for {} because it is in workspace {}, not {}",
                         app.getAppName(),
+                        appType,
                         app.getWorkspaceId(),
                         wdsServerConfiguration.instanceId());
                   }
-                  var b = leonardoServerConfiguration.wdsAppTypes().contains(app.getAppType());
+                  var b = leonardoServerConfiguration.appTypeNames().contains(app.getAppType());
```
salonishah11, line 102: When we add new app types for Workflows app and Cromwell runner app, does this line mean that we would be looking for WDS urls in these new app types as well?

kpierre13: From my understanding, this function in getting a list of suitable apps from Leonardo. This line in particular I think is checking that the app that is currently being looked at is a valid app type in Leonardo.

Since WDS is now a stand-alone app, this change was mostly a naming covention change, as now we're getting an array of app types and not just a list of app types for the sole purpose of getting the wds proxy url.

As far as I'm aware, the only way to get the WDS urls will be from the wds app type. The new app types won't have any knowledge of wds urls like the cromwell app type does today.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
kpierre13: Looking at the changelog, this was the result of `spotlessApply`

service/src/main/resources/application.yml
salonishah11: Does this require an update too to cromwhelm and terra-helmfile charts?

salonishah11 (approved):

PR #182, opened 2023-07-28
[WM-2094] Get User from API Call

dillydally414:
- Added `ControllerUtils` to extract out common logic
- Added `BearerTokenFilter` to move the token from a header into request attributes (if it exists)
- Added simple tests for both of the above
- See usage example in [`RunSetsApiController`](https://github.com/DataBiosphere/cbas/commit/a542c11bb687db6109a48b9bbf572ea8e2233627#diff-1bd3bc7318e38eee77c2fce2eff273155ad4195dcb9bed577968795ee1baac6f)


service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
```diff
@@ -0,0 +1,43 @@
...
+  private final ServletRequest servletRequest;
+  private final SamService samService;
+
+  @Autowired
+  ControllerUtils(ServletRequest servletRequest, SamService samService) {
+    this.servletRequest = servletRequest;
+    this.samService = samService;
+  }
+
+  Optional<String> getUserToken() {
```
dillydally414, line 25: When we move to the 1-hour token proposition we should hopefully be able to update this method without changing usage

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
```diff
@@ -0,0 +1,43 @@
...
+  Optional<UserStatusInfo> getSamUser() {
+    return getUserToken()
+        .map(
+            token -> {
+              try {
+                return samService.getUserStatusInfo(token);
+              } catch (InterruptedException e) {
+                logger.error(e.getMessage(), e.getCause());
+                Thread.currentThread().interrupt();
+                return null;
```
dillydally414, line 39: I'm not totally sure how to handle this InterruptedException that comes from Sam...

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -71,6 +73,7 @@
 @Controller
 public class RunSetsApiController implements RunSetsApi {
 
+  private final ControllerUtils controllerUtils;
```
dillydally414, line 76: Adding ControllerUtils should be the only necessary update to any existing controllers - this also means we can modify ControllerUtils fields without worrying about updating every usage

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
cjllanwarne: 💯

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+public class ControllerUtils {
+  private static final Logger logger = LoggerFactory.getLogger(ControllerUtils.class);
+  private final ServletRequest servletRequest;
+  private final SamService samService;
+
+  @Autowired
+  ControllerUtils(ServletRequest servletRequest, SamService samService) {
```
cjllanwarne, line 20: OOI, what is a ServletRequest and how are we auto-wiring it?

I believe `@Component`s are singletons throughout the system, and not recycled. Will that be a problem if more than one user is making requests and the `ServletRequest` is generated once and then shared?

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
cjllanwarne: If in doubt (or if different callers might want to handle it differently), feel free to just declare it as thrown by the method and then handle it further down the call stack

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
```diff
@@ -28,4 +36,13 @@ public Result checkHealth() {
       return new Result(false, e.getMessage());
     }
   }
+
+  public UserStatusInfo getUserStatusInfo(String accessToken) throws InterruptedException {
+    UsersApi usersApi = samUsersApi(accessToken);
+    try {
+      return SamRetry.retry(usersApi::getUserStatusInfo);
+    } catch (ApiException apiException) {
+      throw SamExceptionFactory.create("Error getting user status info from Sam", apiException);
```
cjllanwarne, line 45: Is this what's creating the `InterruptedException`? Is there a reason not to just let this throw the `ApiException` directly?

service/src/main/java/bio/terra/cbas/util/BearerTokenFilter.java
```diff
@@ -0,0 +1,53 @@
...
+import javax.servlet.ServletResponse;
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import org.springframework.stereotype.Component;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+
+/**
+ * Servlet filter that inspects the incoming request for an LeoToken=... cookie, and saves any token
+ * it finds into the current thread-local RequestContextHolder.
```
cjllanwarne, line 21: This is probably not the right way to go any more. The `LeoToken` cookie was a bit of a hack to allow UIs to work between browser navigations to different hosts, but might disappear soon (because 1. we don't need them any more, and 2. browsers are moving away from allowing these kinds of cookies). Luckily, we should be able to assume that the requests reaching CBAS have a much more standard `Authorization` header on them, which will be of the form `Authorization: Bearer <token>`

cjllanwarne (changes requested):

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
dillydally414: Copied from https://github.com/DataBiosphere/terra-workspace-manager/blob/144c9c0aa4df21fdefd11f77f00b9df7f476a339/service/src/main/java/bio/terra/workspace/service/iam/SamService.java#L191-L200. When I dug into SamExceptionFactory I found this:
```
/**
 * Factory class for creating common Sam exceptions. This class pulls the error message out of the
 * Sam ApiException, which could be buried a layer down, and then inspects the HTTP Status code to
 * generate the right type of exception.
 */
```

service/src/main/java/bio/terra/cbas/util/BearerTokenFilter.java
dillydally414: I was wondering about this... since we send an Authorization header from Terra-UI but a LeoToken from CBAS-UI. I will change this back to Authorization header, which will agree with WDS as well

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
dillydally414: Found some more info [here](https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/sam/exception/SamExceptionFactory.java#L75). Seems like we should catch and rethrow our own exception for consistency's sake rather than deal with Java's builtin

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
```diff
@@ -28,4 +36,17 @@ public Result checkHealth() {
       return new Result(false, e.getMessage());
     }
   }
+
+  public UserStatusInfo getUserStatusInfo(String accessToken) {
```
dillydally414, line 40: @salonishah11 is this an example of a SamService method that should be tested? I'm assuming I would need to mock out the API calls?

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
dillydally414: This method now handles the InterruptException by rethrowing the SamException version for consistency's sake, so the only type of exception coming out of this method will be an [`ErrorReportException`](https://github.com/DataBiosphere/terra-common-lib/blob/eaaf6217ec0f024afa45aac14d21c8964c0f27c5/src/main/java/bio/terra/common/exception/ErrorReportException.java#L15)

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
dillydally414: https://github.com/DataBiosphere/terra-workspace-data-service/blob/b479672d36545d0b45766b3e0e09564d077a19b8/service/src/main/java/org/databiosphere/workspacedataservice/activitylog/ActivityEventBuilder.java#L47-L48

Using WDS as reference, I removed the autowired request in favor for RequestContextAttributes, which seems to be scoped to each request (and because we're not storing anything hopefully less vulnerable to race conditions)

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
salonishah11: I think so. You can probably mock out the response from Sam and add unit tests for this function.

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
```diff
@@ -0,0 +1,31 @@
...
+import bio.terra.cbas.util.BearerTokenFilter;
+import java.util.Optional;
+import org.broadinstitute.dsde.workbench.client.sam.model.UserStatusInfo;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+
+@Component
+public class ControllerUtils {
```
salonishah11, line 13: Clarifying question: I am wondering if these here can be moved inside SamService as they seem dependent on it? Or is there a reason that they should be in a separate class? 🤔

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
```diff
@@ -19,6 +23,10 @@ private StatusApi getStatusApi() {
     return new StatusApi(samClient.getApiClient());
   }
 
+  private UsersApi samUsersApi(String accessToken) {
```
salonishah11, line 26: small nit: to keep this method consistent with other methods as well as prefix it with a verb, maybe it can be renamed to `getUsersApi`?

service/src/main/java/bio/terra/cbas/util/BearerTokenFilter.java
salonishah11, line 1: Since this is specific to Sam maybe it can be moved inside `service/src/main/java/bio/terra/cbas/dependencies/sam/`?

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
```diff
@@ -0,0 +1,31 @@
...
+  private final SamService samService;
+
+  @Autowired
+  ControllerUtils(SamService samService) {
+    this.samService = samService;
+  }
+
+  Optional<String> getUserToken() {
+    Object token =
+        RequestContextHolder.currentRequestAttributes()
```
cjllanwarne, line 23: Since I had to go look this up, I would maybe add a comment here along the lines of (stolen from the javadoc):
```
Exposes the web request in the form of a thread-bound [RequestAttributes](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/context/request/RequestAttributes.html) object.
```

So as long as we call this function from the _thread_ that's servicing a request, we should be certain to get the right request's headers. Which is **_great_**, that was the main thing I was afraid of.

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
```diff
@@ -0,0 +1,31 @@
...
+import bio.terra.cbas.dependencies.sam.SamService;
+import bio.terra.cbas.util.BearerTokenFilter;
+import java.util.Optional;
+import org.broadinstitute.dsde.workbench.client.sam.model.UserStatusInfo;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+
+@Component
```
cjllanwarne, line 12: For you to think about, my inclination would be to remove the `Component`-ness from this and make this a more pure utility class (ie, no instances, all static functions, each function gets the values it needs to do its job as arguments).

Right now the only reason for this to be a Component is to get a `SamService`, but the callers are more naturally in the "component" space so they should be able to autowire in a SamService in a way that I _think_ would feel more natural.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -161,6 +166,13 @@ public ResponseEntity<RunSetListResponse> getRunSets(UUID methodId, Integer page
 
   @Override
   public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
+    Optional<UserStatusInfo> user = controllerUtils.getSamUser();
+    if (user.isPresent()) {
+      log.info("User ID: {}", user.get().getUserSubjectId());
```
cjllanwarne, line 171: I assume this is here for now as a placeholder until your next ticket? We probably want to have some standard form (and maybe even utility functions) for logging these, like "User X did action Y on resource Z"

service/src/main/java/bio/terra/cbas/util/BearerTokenFilter.java
```diff
@@ -0,0 +1,50 @@
...
+ * Servlet filter that inspects the incoming request for an Authorization: Bearer ... token, and
+ * saves any token it finds into the current thread-local RequestContextHolder.
+ *
+ * <p>Note that this filter does not validate or inspect the token; it just extracts it from the
+ * request, allowing it to be sent as-is from CBAS to other services such as Sam.
+ *
+ * <p>Adapted from <a
+ * href="https://github.com/DataBiosphere/terra-workspace-data-service/blob/b479672d36545d0b45766b3e0e09564d077a19b8/service/src/main/java/org/databiosphere/workspacedataservice/sam/BearerTokenFilter.java">WDS</a>
+ */
+@Component
```
cjllanwarne, line 26: Since I found this being a `Component` a little confusing too... could you add a quick comment along the lines of how Spring uses these `Filter` components to populate the request attributes with this sort of pseudo-attribute?

And TOL for the same of simplicity... you could probably also have had this be a static utility function that takes the list of attributes as an argument and returns the auth token, without any need for all this magic filter component instantiation, like
```
public static String getAuthToken(RequestAttributes attributes) { ... }
```

service/src/test/java/bio/terra/cbas/controllers/TestControllerUtils.java
```diff
@@ -0,0 +1,89 @@
...
+          .userEmail("realuser@gmail.com")
+          .userSubjectId("user-id-foo")
+          .enabled(true);
+
+  @BeforeEach
+  void init() {
+    SamService samService = mock(SamService.class);
+    when(samService.getUserStatusInfo(tokenValue)).thenReturn(mockUser);
+    when(samService.getUserStatusInfo(expiredTokenValue))
+        .thenThrow(new SamUnauthorizedException("Unauthorized :("));
```
cjllanwarne, line 36: Was going to comment negatively about the `:(` but then I saw that this is in a test, and instead: 💯 🥇

service/src/test/java/bio/terra/cbas/util/TestBearerTokenFilter.java
```diff
@@ -0,0 +1,64 @@
...
+    RequestContextHolder.resetRequestAttributes();
+    response = new MockHttpServletResponse();
+    chain = new MockFilterChain();
+  }
+
+  @Test
+  void testRequestWithAuthHeader() throws ServletException, IOException {
+    String tokenValue = "foo-token";
+    MockHttpServletRequest request = new MockHttpServletRequest();
+    request.addHeader(HttpHeaders.AUTHORIZATION, String.format("Bearer %s", tokenValue));
```
cjllanwarne, line 38: Since you have the conditional in the code, you might as well test a case where there's no `Bearer` in the value to remove

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+  Optional<String> getUserToken() {
+    Object token =
+        RequestContextHolder.currentRequestAttributes()
+            .getAttribute(BearerTokenFilter.ATTRIBUTE_NAME_TOKEN, RequestAttributes.SCOPE_REQUEST);
+    return Optional.ofNullable((String) token);
+  }
+
+  Optional<UserStatusInfo> getSamUser() {
+    return getUserToken().map(samService::getUserStatusInfo);
```
cjllanwarne, line 29: FYI the token is almost certainly a JWT, and might have some of these fields available without a secondary call out to Sam, if we can find the right decoder

service/src/main/java/bio/terra/cbas/util/BearerTokenFilter.java
dillydally414: My thought here was that the filter may be useful for logging purposes later down the line, since it's automatically applied once to every incoming request (whereas the static utility function would be called each time we get the user token, and I'm not sure about the consistency across requests). I think I may be looking ahead too much though, since I imagine logging may end up being more complicated and request-dependent.

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
dillydally414: @cjllanwarne does this approach make sense over the `ControllerUtils` static methods suggested above? (Could just get rid of ControllerUtils)

service/src/main/java/bio/terra/cbas/controllers/ControllerUtils.java
dillydally414: It actually has _more_ fields than the SamUser call... except the user ID, which is (I suspect) the most important 🙃 When we implement 1 hour tokens from Sam we would be talking to Sam anyways though, so I'm not sure if it's terrible to have to talk to them here.

service/src/main/java/bio/terra/cbas/util/BearerTokenFilter.java
dillydally414: Tried moving into a helper/utility method - the problem that arises is suddenly we are dealing with `request` objects again, since headers are not in the attributes. I think this `Filter` may be a necessary evil if we want to restrict the number of somewhat opaque request objects being passed around. I will add/update comments for better context

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -292,6 +308,34 @@ void setupFunctionalChecks() throws Exception {
                 "ApiException thrown on purpose for testing purposes."));
     when(cromwellService.submitWorkflow(eq(workflowUrl), eq(workflowInputsMap3), any()))
         .thenReturn(new RunId().runId(cromwellWorkflowId3));
+
+    when(samService.getSamUser()).thenReturn(Optional.of(mockUser));
+  }
+
+  @Test
+  void runSetWithSamException() throws Exception {
```
dillydally414, line 316: Added a test here that might get tossed out or significantly changed when we actually start doing things with this data but is a good sanity check for exception messaging/codes.

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -559,7 +603,7 @@ void tooManyRecordIds() throws Exception {
     String inputSourceAsString = "{ \"type\" : \"none\", \"record_attribute\" : null }";
     String request =
         requestTemplate.formatted(
-            workflowUrl,
+            methodVersionId,
```
dillydally414, line 606: Turns out this test was actually going sideways but since we were checking for an error it didn't pick up on it until I added the GlobalExceptionHandler, which makes thrown errors handled more accurately (and doesn't seem to impact ResponseEntities with error status codes)

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
```diff
@@ -19,6 +26,23 @@ private StatusApi getStatusApi() {
...
+  public Optional<String> getUserToken() {
+    // RequestContextHolder exposes the web request in the form of a *thread-bound*
+    // [RequestAttributes](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/context/request/RequestAttributes.html) object.
+    Object token =
+        RequestContextHolder.currentRequestAttributes()
+            .getAttribute(BearerTokenFilter.ATTRIBUTE_NAME_TOKEN, RequestAttributes.SCOPE_REQUEST);
+    return Optional.ofNullable((String) token);
+  }
+
+  public Optional<UserStatusInfo> getSamUser() {
```
dillydally414, line 42: Noticed during tests: If there's no token, this method won't throw an error, it'll just return empty user object. That feels like... not the expected behavior? No token should presumably be handled similarly to expired token, where an error is thrown (probably 401). @cjllanwarne am I thinking about this correctly?

service/src/main/java/bio/terra/cbas/controllers/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,23 @@
...
+ * href="https://github.com/DataBiosphere/terra-java-project-template/blob/43cff86354e8d295ed96c29a7faf60eadd11e502/service/src/main/java/bio/terra/javatemplate/controller/GlobalExceptionHandler.java">Taken
+ * from Terra template</a> This enables proper logging and response codes when errors are thrown
+ * within the code (primarily when interacting with Sam). Without this, any errors are reported to
+ * the user as a 500 Internal Server Error.
+ */
+@RestControllerAdvice
+public class GlobalExceptionHandler extends AbstractGlobalExceptionHandler<ErrorReport> {
+
+  @Override
+  public ErrorReport generateErrorReport(Throwable ex, HttpStatus statusCode, List<String> causes) {
```
salonishah11, line 20: I am having a little trouble understanding where this is used. Can you point me to it? Does this apply to all APIs?

dillydally414: My understanding is that this is automatically hooked into controllers and called when controller methods throw exceptions, which is useful because not all of our error cases return nice ResponseEntities, but we don't want every exception classified as an InternalServerError.

dillydally414: It is similar to the BearerTokenFilter in that we do not explicitly call it, but Spring will call this method on outgoing responses with Exceptions. Here's a link with more info: [Spring Exception handling](https://www.baeldung.com/exception-handling-for-rest-with-spring), check section titled "Solution 3". This is used in many of the other Terra java projects as well for reference

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
dillydally414: I think I am leaning more towards the exception because we wouldn't be calling getUser from within a method unless we needed an authenticated user to work with - in which case we would want to error out if we don't have one. I don't foresee us having a situation where we want to allow a non-logged in user to call our endpoints (excluding status api), and I actually think the relay listener will reject before it even gets to this point. If we did have a weird case where we wanted to check the user but didn't care if they existed, I think I'd prefer to have to manually catch the exception to signify we're altering the typical behavior.

dillydally414: My thought is that we may be able to just replace this method with `getUserStatusInfo` since that will throw whatever error Sam throws when the token is missing instead of us creating our own exception handling/routing.

salonishah11: I think the `getUserStatusInfo` or `getSamUser` would most likely be used when fetching the user ID to either log or store it in database. For checking the permissions we might only be using `getUserToken` method.

dillydally414: This would probably mean that our controllers would block the requests before even getting to the logging/storing point, since if the user doesn't have a token they presumably won't have any permissions. We can talk more about this this afternoon, I think the Optionals may not be necessary.

dillydally414:
Comments/to-dos from walkthrough:

- [ ] Look into [WSM](https://github.com/DataBiosphere/terra-workspace-manager)'s strategy and compare with current/WDS strategy
- [ ] Look into caching of user info from token so as not to call Sam for that on each request

salonishah11:
> Look into caching of user info from token so as not to call Sam for that on each request

@dillydally414 I wonder if it makes sense to create a separate ticket for this? This way the scope of current ticket doesn't increase and the PR size doesn't get large? What do you think?

dillydally414:
@salonishah11 sorry just saw this - it ended up being a pretty minor change so I included it here. I'm not sure how to actually test this but I was able to verify the caching by running tests against the same user token cached and it took 1ms the 2nd-4th time I ran the method vs 474ms the 1st time.

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
```diff
@@ -19,6 +27,35 @@ private StatusApi getStatusApi() {
...
+    // [RequestAttributes](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/context/request/RequestAttributes.html) object.
+    return (String)
+        RequestContextHolder.currentRequestAttributes()
+            .getAttribute(BearerTokenFilter.ATTRIBUTE_NAME_TOKEN, RequestAttributes.SCOPE_REQUEST);
+  }
+
+  // this cache uses token.hashCode as its key. This prevents any logging such as
+  // in CacheLogger from logging the raw token.
+  // Taken from WDS
+  @Cacheable(cacheNames = "tokenResolution", key = "#root.target.getUserToken().hashCode()")
```
dillydally414, line 45: Not sure how concerned we should be about `hashCode` collisions... Wondering if there is a better way we can encode the token to avoid logging raw token but without the potential for collisions.

service/src/main/resources/application.yml
```diff
@@ -13,6 +13,9 @@ spring:
   # https://github.com/DataBiosphere/terra-common-lib/blob/480ab3daae282ddff0fef8dc329494a4422e32f1/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java#L118
   application.name: cbas
   application.version: ${cbas.version.gitHash:unknown}
+  cache:
+    cache-names: tokenResolution
+    caffeine.spec: expireAfterWrite=1h
```
dillydally414, line 18: Unfortunately it doesn't seem like there's a way to custom-set expiration based on whether the tokens have hit their expiry time, but given that Terra-UI and the to-be-implemented permission checks will both handle expired tokens it's probably fine to let the token to ID conversion hang around a little after the token itself expires.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -161,6 +166,9 @@ public ResponseEntity<RunSetListResponse> getRunSets(UUID methodId, Integer page
 
   @Override
   public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
+    UserStatusInfo user = samService.getSamUser();
+    log.info("User ID: {}", user.getUserSubjectId());
```
cjllanwarne, line 170: I would either remove this or add a `// TODO: remove/update in ticket WM-XXXX` comment, because I don't _think_ this is something we want there long term?

service/src/main/java/bio/terra/cbas/dependencies/sam/BearerTokenFilter.java
```diff
@@ -0,0 +1,58 @@
...
+ * <a href="https://docs.oracle.com/javaee/6/api/javax/servlet/Filter.html">Servlet filter</a> that
+ * inspects the incoming request for an Authorization: Bearer ... token, and saves any token it
+ * finds into the current thread-local RequestContextHolder.
+ *
+ * <p>Filters are contructed once (hence the @Component tag), while doFilter is called for each
+ * request/response pair. Read the Javadoc for `doFilter` to learn more about its function. By
+ * isolating the request to this class, we can access thread-specific RequestContextHolders without
+ * worrying about whether a given `request` object is up-to-date. The filter is necessary to inspect
+ * certain request properties like headers, while the RequestContextHolders serves more as a
+ * dictionary for specific keyed attributes.
```
cjllanwarne, line 26: Great comments, thank you!

service/src/main/java/bio/terra/cbas/dependencies/sam/BearerTokenFilter.java
```diff
@@ -0,0 +1,58 @@
...
+ * <p>Adapted from <a
+ * href="https://github.com/DataBiosphere/terra-workspace-data-service/blob/b479672d36545d0b45766b3e0e09564d077a19b8/service/src/main/java/org/databiosphere/workspacedataservice/sam/BearerTokenFilter.java">WDS</a>
+ */
+@Component
+public class BearerTokenFilter implements Filter {
+  public static final String ATTRIBUTE_NAME_TOKEN = "bearer-token-attribute";
+  private static final String BEARER_PREFIX = "Bearer ";
+
+  @Override
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
```
pshapiro4broad, line 40: I hadn't seen this before so this approach is new to me and I'm not sure of the pros/cons vs the approach I took with adding support for `BearerToken`. This code uses a filter to populate the request context but then don't make use of using the context to directly inject a request scoped bean, which seems like a missed opportunity to me.

But I think a bigger issue is that you don't need a filter to do this, which makes things more complicated. You can instead define a custom request scoped bean which produces `BearerToken` and inject it directly where needed.

The full PR is here https://github.com/DataBiosphere/terra-data-catalog/pull/37

But the basic idea is to add a bean generator to your `Configuration`:
```
  @Bean
  @RequestScope
  public BearerToken bearerToken(HttpServletRequest request) {
    return new BearerTokenFactory().from(request);
  }
```

where `BearerTokenFactory` is from terra-common-lib. Once you've done that, you can inject the bearer token directly in your code, like this:
```
  @Autowired
  public SamService(SamConfiguration samConfig, SamClient samClient, BearerToken bearerToken) {
    this.samConfig = samConfig;
    this.samClient = samClient;
    this.bearerToken = bearerToken;
  }
```

Note that since `BearerToken` is a request scoped bean, each request gets is own copy even though `SamService` is a singleton bean.

service/src/main/resources/application.yml
cjllanwarne: Yeah I think this is ok. As long as we keep the authz cache short, I don't think there's any risk here. Maybe we can rename the cache to `tokenToUserIdResolution` since there might be other token-y resolutions to cache and we don't want to re-use this one inappropriately.

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
cjllanwarne: I think I'd feel better if the cached function took arguments like `(String token, String tokenHash)` and it was _that_ hash that was the key to the cache. It should be easy enough to wrap the call in a way that lets you extract the relevant fields and it would reduce the amount of magic (like... what is "`root`"?)

By the way, in java `hashCode`, unless overridden, gives you a value depending on memory address, so this might not always give the same hashcode for the same "value".

service/src/test/java/bio/terra/cbas/dependencies/sam/TestSamService.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Test
+  void testGetUserToken() {
+    setTokenValue(tokenValue);
+    String userToken = samService.getUserToken();
+    assertEquals(tokenValue, userToken);
+  }
+
+  @Test
+  void testGetSamUserNoToken() {
```
cjllanwarne, line 78: I would add a test for the caching too, in particular making sure that two different instances of the same token value uses the cache. (by the way, I think that's another reason for making the getSamUser function take parameters rather than relying on threadlocal magic - it makes these tests a lot easier to write!)

cjllanwarne (commented):
Mostly good, a couple of hopefully small comments

service/src/test/java/bio/terra/cbas/dependencies/sam/TestSamService.java
dillydally414: I did a lot of digging and configuration changes to try to test the caching and it seems that caching doesn't agree well with spring boot tests. Willing to discuss further tomorrow but I kept running into a brick wall today and I can't seem to find examples in other Terra repos or online.

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
dillydally414: Looks like `BearerToken.hashCode()` uses `Objects.hash`, which should be stable for Strings

cjllanwarne (approved):

salonishah11 (approved):

PR #185, opened 2023-07-28
WM-2100: Correct Jira ID in terra-helmfile publisher

cjllanwarne:


kpierre13 (approved):

cjllanwarne:
Bypassing double code review requirement because this PR is not going to production, and the chart it updates in terra-helmfile is also not going to production.

PR #187, opened 2023-08-08
[WM-2093] Add user id to RunSet

dillydally414:
This PR adds a `user_id` column to the `run_set` table, which is populated based on the token provided in the request. There are many files that were updated just to include the extra parameter in the RunSet constructor, so I have commented where the important changes were made.

Note that this PR does not expose `user_id` on API responses - it's simply responsible for storing this new field in the database.

### DIFF
--- a/service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
@@ -268,6 +268,7 @@ public class MethodsApiController implements MethodsApi {
             0,
             objectMapper.writeValueAsString(inputs),
             objectMapper.writeValueAsString(outputs),
+            null,
             null);
 
     methodDao.createMethod(method);
--- a/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
@@ -166,9 +166,6 @@ public class RunSetsApiController implements RunSetsApi {
 
   @Override
   public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
-    UserStatusInfo user = samService.getSamUser();
-    log.info("User ID: {}", user.getUserSubjectId()); // TODO: remove in WM-2093
-
     captureRequestMetrics(request);
 
     // request validation
@@ -228,6 +225,8 @@ public class RunSetsApiController implements RunSetsApi {
           new RunSetStateResponse().errors(errorMsg), HttpStatus.INTERNAL_SERVER_ERROR);
     }
 
+    UserStatusInfo user = samService.getSamUser();
+
     // Create a new run_set
     UUID runSetId = this.uuidSource.generateUUID();
     RunSet runSet;
@@ -249,7 +248,8 @@ public class RunSetsApiController implements RunSetsApi {
               0,
               objectMapper.writeValueAsString(request.getWorkflowInputDefinitions()),
               objectMapper.writeValueAsString(request.getWorkflowOutputDefinitions()),
-              request.getWdsRecords().getRecordType());
+              request.getWdsRecords().getRecordType(),
+              user.getUserSubjectId());
     } catch (JsonProcessingException e) {
       log.warn("Failed to record run set to database", e);
       return new ResponseEntity<>(
--- a/service/src/main/java/bio/terra/cbas/dao/RunSetDao.java
+++ b/service/src/main/java/bio/terra/cbas/dao/RunSetDao.java
@@ -66,8 +66,8 @@ public class RunSetDao {
 
   public int createRunSet(RunSet runSet) {
     return jdbcTemplate.update(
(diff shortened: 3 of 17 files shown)


service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -228,6 +225,8 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
           new RunSetStateResponse().errors(errorMsg), HttpStatus.INTERNAL_SERVER_ERROR);
     }
 
+    UserStatusInfo user = samService.getSamUser();
```
dillydally414, line 228: important addition here

service/src/main/java/bio/terra/cbas/dao/RunSetDao.java
```diff
@@ -66,8 +66,8 @@ public RunSet getRunSetWithMethodId(UUID methodId) {
 
   public int createRunSet(RunSet runSet) {
     return jdbcTemplate.update(
-        "insert into run_set (run_set_id, method_version_id, run_set_name, run_set_description, call_caching_enabled, is_template, status, submission_timestamp, last_modified_timestamp, last_polled_timestamp, run_count, error_count, input_definition, output_definition, record_type)"
-            + " values (:runSetId, :methodVersionId, :name, :description, :callCachingEnabled, :isTemplate, :status, :submissionTimestamp, :lastModifiedTimestamp, :lastPolledTimestamp, :runCount, :errorCount, :inputDefinition, :outputDefinition, :recordType)",
+        "insert into run_set (run_set_id, method_version_id, run_set_name, run_set_description, call_caching_enabled, is_template, status, submission_timestamp, last_modified_timestamp, last_polled_timestamp, run_count, error_count, input_definition, output_definition, record_type, user_id)"
```
dillydally414, line 69: important addition here

service/src/main/java/bio/terra/cbas/dao/mappers/RunSetMapper.java
```diff
@@ -32,6 +32,7 @@ public RunSet mapRow(ResultSet rs, int rowNum) throws SQLException {
         rs.getInt(RunSet.ERROR_COUNT_COL),
         rs.getString(RunSet.INPUT_DEFINITION_COL),
         rs.getString(RunSet.OUTPUT_DEFINITION_COL),
-        rs.getString(RunSet.RECORD_TYPE_COL));
+        rs.getString(RunSet.RECORD_TYPE_COL),
+        rs.getString(RunSet.USER_ID_COL));
```
dillydally414, line 36: important addition here

service/src/main/java/bio/terra/cbas/models/RunSet.java
```diff
@@ -18,7 +18,8 @@ public record RunSet(
     Integer errorCount,
     String inputDefinition,
     String outputDefinition,
-    String recordType) {
+    String recordType,
+    String userId) {
```
dillydally414, line 22: important addition here

service/src/main/resources/changelog/changesets/20230808_add_user_id_column.yaml
```diff
@@ -0,0 +1,11 @@
+databaseChangeLog:
```
dillydally414, line 1: important addition here

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -379,6 +379,7 @@ void runSetWith1FailedRunTest() throws Exception {
     assertEquals(recordType, newRunSetCaptor.getValue().recordType());
     assertEquals(outputDefinitionAsString, newRunSetCaptor.getValue().outputDefinition());
     assertEquals(isCallCachingEnabled, newRunSetCaptor.getValue().callCachingEnabled());
+    assertEquals(mockUser.getUserSubjectId(), newRunSetCaptor.getValue().userId());
```
dillydally414, line 382: actual test here

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -268,6 +268,7 @@ private void createNewMethod(
             0,
             objectMapper.writeValueAsString(inputs),
             objectMapper.writeValueAsString(outputs),
+            null,
```
mspector, line 271: my only question is: is it ok for this value to be null, instead of an empty string?

I assume the answer is "yes" (because I see RecordType - also a string - was also being set to null) so I'm approving. But I'm curious to understand, if you have any comment

mspector (approved):

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
dillydally414: I think it should be fine for 2 reasons: 1) any database entries prior to this change will have null user IDs associated, so this column should be accepting of null values, and 2) we probably don't want to associate a user id with a template run set anyways. Tbh, I don't think empty string vs null makes a huge difference, but I stuck with null for consistency with RecordType

salonishah11 (approved):

PR #188, opened 2023-08-09
[WM-2125] CBAS checks user auth with Sam for all APIs

salonishah11:
This PR adds the ability in CBAS to check whether the user has proper permissions to access each endpoint in CBAS by calling Sam's `resourcePermissionV2` API. The appropriate access needed for each API was documented in [WM-2126](https://broadworkbench.atlassian.net/browse/WM-2126). The main logic for this is in `SamService.java` file.

Closes https://broadworkbench.atlassian.net/browse/WM-2125

### DIFF
--- a/service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
@@ -9,11 +9,13 @@ import static bio.terra.cbas.util.methods.WomtoolToCbasInputsAndOutputs.womToCba
 import bio.terra.cbas.api.MethodsApi;
 import bio.terra.cbas.common.DateUtils;
 import bio.terra.cbas.common.MethodUtil;
+import bio.terra.cbas.common.exceptions.ForbiddenException;
 import bio.terra.cbas.common.exceptions.WomtoolValueTypeProcessingException.WomtoolValueTypeNotFoundException;
 import bio.terra.cbas.dao.MethodDao;
 import bio.terra.cbas.dao.MethodVersionDao;
 import bio.terra.cbas.dao.RunSetDao;
 import bio.terra.cbas.dependencies.dockstore.DockstoreService;
+import bio.terra.cbas.dependencies.sam.SamService;
 import bio.terra.cbas.dependencies.wes.CromwellService;
 import bio.terra.cbas.model.MethodDetails;
 import bio.terra.cbas.model.MethodInputMapping;
@@ -45,6 +47,8 @@ import java.util.UUID;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
@@ -53,19 +57,23 @@ import org.springframework.stereotype.Controller;
 public class MethodsApiController implements MethodsApi {
   private final CromwellService cromwellService;
   private final DockstoreService dockstoreService;
+  private final SamService samService;
   private final MethodDao methodDao;
   private final MethodVersionDao methodVersionDao;
   private final RunSetDao runSetDao;
+  private static final Logger logger = LoggerFactory.getLogger(MethodsApiController.class);
 
   public MethodsApiController(
       CromwellService cromwellService,
       DockstoreService dockstoreService,
+      SamService samService,
       MethodDao methodDao,
       MethodVersionDao methodVersionDao,
       RunSetDao runSetDao,
       ObjectMapper objectMapper) {
     this.cromwellService = cromwellService;
     this.dockstoreService = dockstoreService;
+    this.samService = samService;
     this.methodDao = methodDao;
     this.methodVersionDao = methodVersionDao;
     this.runSetDao = runSetDao;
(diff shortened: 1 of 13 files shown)


service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -76,6 +85,14 @@ public MethodsApiController(
 
   @Override
   public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMethodRequest) {
+    // check if current user has write permissions on the workspace
+    if (!samService.hasWritePermission()) {
+      String errorMsg =
+          getSamForbiddenExceptionMsg(SamService.WRITE_ACTION, SamService.RESOURCE_TYPE_WORKSPACE);
+      logger.info(errorMsg);
```
dillydally414, line 92: Do we need to log messages as well as throw the exception? When I run this locally, I see this logged message, but I also see the messages logged by the `GlobalExceptionHandler`, and the response contains the proper message as well. I think we may be able to get rid of the `logger.info` and let GlobalExceptionHandler deal with the messageExtraction. Then we could move the ForbiddenExceptionMsg to the ForbiddenException itself, and pass in the Action and ResourceType to the constructor. I think this would help reduce duplication across controller/endpoints as well.

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
```diff
@@ -54,4 +103,16 @@ public Result checkHealth() {
       return new Result(false, e.getMessage());
     }
   }
+
+  public boolean hasReadPermission() {
```
dillydally414, line 107: Opening up to discussion - what do we think about `checkReadPermission` instead of `hasReadPermission`, where `checkReadPermission` throws an exception if user doesn't have permission and does nothing if user has permission. Then we wouldn't need to explicitly throw the same exception at the top of each endpoint.

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
```diff
@@ -27,10 +44,42 @@ private StatusApi getStatusApi() {
...
+    } catch (ApiException e) {
+      throw SamExceptionFactory.create(
+          "Error checking %s permissions on workspace from Sam".formatted(actionType), e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw SamExceptionFactory.create(
+          "Request interrupted while checking %s permissions on workspace from Sam"
+              .formatted(actionType),
+          e);
+    }
```
dillydally414, line 72: This looks very similar to the `getUserStatusInfo` method body. Maybe outside of the scope of this PR but it could be nice to abstract out the similarities if not too complex.

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
salonishah11: Yeah I added this log message with the thought that when we fetch user IDs we can log the user ID here, but not send the user ID back to the user via the exception message.

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
salonishah11: Yeah that is true. Abstracting out the try block and different error messages passed to `SamExceptionFactory.create` makes sense but I am not sure how it will affect the readability of the code. 🤔 If you have strong preference I can try generalizing it in this PR?

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
dillydally414: The SamException would still be displayed differently from the ForbiddenException even if we throw ForbiddenException from inside SamService. I do agree that this method allows for better testing and wider use cases, I'm just not sure how many cases we will want to check "does the user have the given access" and proceed even if the user doesn't. Also interested in hearing from other reviewers

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
```diff
@@ -27,10 +44,42 @@ private StatusApi getStatusApi() {
     return new StatusApi(samClient.getApiClient());
   }
 
+  boolean hasPermission(String actionType) {
+    // don't check auth access with Sam if "sam.checkAuthAccess" is false
```
dillydally414, line 48: OOC - what is the use case for having `sam.checkAuthAccess` to false? Is this a temporary placeholder so that we don't start checking these permissions until we're at a further stage of collaboration?

salonishah11: `sam.checkAuthAccess` can be set to false when working locally so that you don't have to pass bearer token to requests.

dillydally414: Ok, I will edit the RunSetsApiController to look at this configuration option before adding the user ID to RunSet, otherwise that will still error out locally.

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
salonishah11: Since each server log, including exceptions would be logged with user ID in a later PR, the log messages added were redundant. Hence I have removed it.

dillydally414 (approved):
Still interested to hear other's thoughts on the previous comment, but overall looks good either way 😄

kpierre13 (approved):

service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
salonishah11: @kpierre13 as the second reviewer can you share your thoughts on this?

dillydally414: That was my thought. `hasXPermission` is beneficial if we want to check access but not throw, but I think we will always want to throw?

salonishah11: The only place I can think of at the moment where we don't want to simply throw an exception is when we want to update the Run and Run set status after getting a callback from Cromwell via the new callback API. In that case, after we check if the user still has write permission, we still might want to clean up the status of Run/Workflow to Failed and possibly return a new/combined error back to user saying that they didn't have permission to update status of Run X. 🤔

dillydally414: Ah, that makes sense. In that case I'm good with leaving as-is

PR #190, opened 2023-08-11
WM-2087: Create placeholder api for run status update 

zykonda:
Place holder API created for run state update.

### DIFF
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -314,11 +314,9 @@ components:
           format: uuid
         state:
           $ref: '#/components/schemas/RunState'
-        workflow_output_definitions:
-          description: Declarative output definition for writing back to WDS.
-          type: array
-          items:
-            $ref: '#/components/schemas/WorkflowOutputDefinition'
+        outputs:
+          type: object
+          description: Map of output keys to output values
 
     RunResultsResponse:
       type: object
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -291,20 +291,7 @@ components:
           items:
             $ref: '#/components/schemas/WorkflowOutputDefinition'
           description: Declarative output definition for writing back to WDS.
-    RunStateRequest:
-      type: object
-      required: [runId, state]
-      properties:
-        runId:
-          type: string
-          format: uuid
-        state:
-          $ref: '#/components/schemas/RunState'
-        workflow_output_definitions:
-          description: Declarative output definition for writing back to WDS.
-          type: array
-          items:
-            $ref: '#/components/schemas/WorkflowOutputDefinition'
+
     RunResultsRequest:
       type: object
       required: [ run_id, state ]
@@ -910,8 +897,3 @@ components:
         - CANCELED
         - CANCELING
       type: string
-
-  securitySchemes:
-    bearerAuth:
-      type: http
-      scheme: bearer
--- a/common/openapi.yml
+++ b/common/openapi.yml
@@ -288,9 +288,9 @@ components:
 
     RunResultsRequest:
       type: object
-      required: [ run_id, state ]
+      required: [ workflowId, state ]
       properties:
-        run_id:
+        workflowId:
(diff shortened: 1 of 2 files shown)


common/openapi.yml
```diff
@@ -158,6 +158,23 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/batch/v1/runs/{runId}/state:
+    parameters:
+      - $ref: '#/components/parameters/RunId'
```
dillydally414, line 163: I think `runId` would be taken care of by the `workflow-id` in the request body. It would also be nice to have a stable endpoint to pass into Cromwell config rather than needing to pass in an override each time a workflow is submitted.

common/openapi.yml
```diff
@@ -158,6 +158,23 @@ paths:
...
+    patch:
```
dillydally414, line 164: Will Cromwell be sending a patch or a post?

common/openapi.yml
zykonda: Correct, I got carried away by implementing the inner logic. Will update PR soon. :)

common/openapi.yml
zykonda: Post it is.

common/openapi.yml
```diff
@@ -157,6 +157,35 @@ paths:
           $ref: '#/components/responses/UserError'
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/batch/v1/run_sets/results:
+    post:
+      tags: [ runs ]
```
mspector, line 162: should this be `run_sets`?

zykonda: updated

common/openapi.yml
```diff
@@ -265,6 +317,77 @@ components:
...
+    RunStateRequest:
+      type: object
+      required: [runId, state]
+      properties:
+        runId:
+          type: string
+          format: uuid
+        state:
+          $ref: '#/components/schemas/RunState'
+        workflow_output_definitions:
```
zykonda, line 329: @jgainerdewar: I think I got this wrong and instead of output definition, the body should have wds_records here coming from cromwell. Please confirm - I'll update the request bodies.

common/openapi.yml
```diff
@@ -157,6 +157,35 @@ paths:
           $ref: '#/components/responses/UserError'
         '500':
           $ref: '#/components/responses/ServerError'
+  /api/batch/v1/run_sets/results:
```
jgainerdewar, line 160: As far as we know we don't have plans to use this type of endpoint. Is it being added just in case it's useful later, or maybe I'm missing something?

zykonda: From the doc: "The callback could be one-per-workflow or could group together batches of workflows"
I can remove this API for now - we can add this later anyway.

common/openapi.yml
jgainerdewar: Agreed, I think it makes sense to use a new scheme.

common/openapi.yml
jgainerdewar: Ah, yeah, we certainly could do that in the future - but no plans to do it immediately.

common/openapi.yml
```diff
@@ -266,6 +292,31 @@ components:
...
+      type: object
+      required: [ run_id, state ]
+      properties:
+        run_id:
+          type: string
+          format: uuid
+        state:
+          $ref: '#/components/schemas/RunState'
+        error:
+          type: string
```
jgainerdewar, line 318: I don't think we need an error message here, if there's an error CBAS should return a non-200 code and either a `UserError` or `ServerError` response. In fact, in the case of success I don't think think we need to return any response body at all, just the success code. Realizing this wasn't specified in the design doc, sorry about that!

zykonda: updated!

jgainerdewar (approved):

PR #191, opened 2023-08-11
[WM-2095] User ID in Logs

dillydally414:
This PR adds user awareness to the existing logging system. Check the code comment in `SamService.java` for further explanation - it is very similar to how we already include `requestId` in our logs (though the heavy lifting is done by terra-common-lib there).


service/src/main/java/bio/terra/cbas/dependencies/sam/SamService.java
```diff
@@ -48,6 +49,13 @@ boolean hasPermission(String actionType) {
     // don't check auth access with Sam if "sam.checkAuthAccess" is false
     if (!samClient.checkAuthAccessWithSam()) return true;
 
+    /* Associate user id with log context for this thread/request. This will be included with
+    request ID within any logs for the remainder of the request. See
+    https://github.com/DataBiosphere/terra-common-lib/blob/eaaf6217ec0f024afa45aac14d21c8964c0f27c5/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java#L129-L132
+    for details on how this is included in cloud logs, and logback.xml for console logs
+    */
+    MDC.put("user", getSamUser().getUserSubjectId());
+
```
dillydally414, line 58: Here is where we actually add the user id to the thread context. It makes sense to me to include it in the permission checks, ensuring it is called at the top of every user-aware endpoint, since we want the user ID attached to all logs for the request.

service/src/main/resources/application.yml
```diff
@@ -1,6 +1,5 @@
 logging:
   level.bio.terra.cbas: ${BATCHANALYSIS_LOG_LEVEL:info}
-  pattern.level: '%X{requestId} %5p'
```
dillydally414, line 3: This was being overriden by terra-common-lib configuration anyways, and is now replaced by `logback.xml`

service/src/main/resources/logback.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dillydally414, line 1: This file is very similar to the default spring configuration, with an addition to the console log pattern to include values from the [MDC](https://logback.qos.ch/manual/mdc.html) (requestId, user). Technically we could do without this as user will still get logged to cloud logs, but for consistency and local testing I think it's beneficial to include this in our console logs, in addition to any other request information we may want to log in the future.

service/src/test/java/bio/terra/cbas/dependencies/sam/TestSamService.java
```diff
@@ -64,15 +72,24 @@ void init() throws ApiException {
...
-                  if (bearerToken != null
-                      && bearerToken.getToken().equals(tokenWithCorrectAccess)) {
-                    return mockUser;
-                  } else if (bearerToken != null
-                      && bearerToken.getToken().equals(tokenCausingInterrupt)) {
-                    throw new InterruptedException();
-                  } else {
-                    // expired or no token
-                    throw new ApiException(401, "Unauthorized :(");
+                  if (bearerToken == null) {
```
dillydally414, line 75: Cleaned this up to include the remaining tokens and be more explicit about our test cases

service/src/test/java/bio/terra/cbas/dependencies/sam/TestSamService.java
```diff
@@ -176,13 +200,53 @@ void testGetSamUserExpiredToken() {
...
   void testGetSamUserInterruptingToken() {
-    setTokenValue(tokenCausingInterrupt);
+    setTokenValue(tokenCausingUserInterrupt);
     SamInterruptedException e =
         assertThrows(SamInterruptedException.class, () -> samService.getSamUser());
     assertEquals("Request interrupted while getting user status info from Sam", e.getMessage());
     assertEquals(HttpStatus.INTERNAL_SERVER_ERROR, e.getStatusCode());
   }
 
+  @Test
```
dillydally414, line 210: These are the actual tests for this PR - ensuring that the proper user id is getting stored in the MDC and is included in the output message. This is true even when the user does not have permission to complete the action.

service/src/test/java/bio/terra/cbas/dependencies/sam/TestSamService.java
```diff
@@ -280,12 +308,78 @@ void testHasPermissionForRequestWithNoToken() {
...
+    setTokenValue(validTokenCausingAccessInterrupt);
     SamInterruptedException exception =
         assertThrows(SamInterruptedException.class, () -> samService.hasComputePermission());
     assertEquals(
         "Request interrupted while checking compute permissions on workspace from Sam",
         exception.getMessage());
     assertEquals(HttpStatus.INTERNAL_SERVER_ERROR, exception.getStatusCode());
   }
+
+  // Tests for including user IDs in logs once permissions have been checked
```
dillydally414, line 320: These are the actual tests for this PR - ensuring that the proper user id is getting stored in the MDC and is included in the output logs. This is true even when the user does not have permission to complete the action, and is included both in our human-readable-logging and normal logging.

service/src/test/java/bio/terra/cbas/dependencies/sam/TestSamService.java
```diff
@@ -64,15 +76,24 @@ void init() throws ApiException {
...
-                  if (bearerToken != null
-                      && bearerToken.getToken().equals(tokenWithCorrectAccess)) {
-                    return mockUser;
-                  } else if (bearerToken != null
-                      && bearerToken.getToken().equals(tokenCausingInterrupt)) {
-                    throw new InterruptedException();
-                  } else {
-                    // expired or no token
-                    throw new ApiException(401, "Unauthorized :(");
+                  if (bearerToken == null || bearerToken.getToken() == null) {
```
dillydally414, line 79: Cleaned this up to include the remaining tokens and be more explicit about our test cases

dillydally414:
This got into a weird state, moved to #193

PR #195, opened 2023-08-17
AJ-1173: new okhttp-based WDS client

davidangb:
This PR updates CBAS to use the new, OkHttp-powered WDS Java client.

See https://github.com/DataBiosphere/terra-workspace-data-service/pull/321 for more detail.


service/build.gradle
```diff
@@ -21,14 +21,14 @@ dependencies {
...
-    implementation 'org.databiosphere:workspacedataservice-client:0.2.57-SNAPSHOT'
+    implementation 'org.databiosphere:workspacedataservice-client-okhttp-javax:0.2.95-SNAPSHOT'
     implementation 'org.springframework.boot:spring-boot-starter-data-jdbc'
     implementation 'org.springframework.boot:spring-boot-starter-web'
     implementation 'org.springframework.boot:spring-boot-starter-thymeleaf'
     implementation 'org.springframework.retry:spring-retry'
     implementation 'org.broadinstitute.dsde.workbench:sam-client_2.12:0.1-61135c7'
     implementation 'bio.terra:datarepo-client:1.313.0-SNAPSHOT'
-    implementation 'javax.ws.rs:javax.ws.rs-api:2.1.1'
+    // implementation 'javax.ws.rs:javax.ws.rs-api:2.1.1'
```
davidangb, line 31: I believe the direct dependency on `javax.ws.rs-api` is not necessary

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -72,7 +72,8 @@ public void updateOutputAttributes(Run run)
     List<WorkflowOutputDefinition> outputDefinitionList =
         objectMapper.readValue(run.runSet().outputDefinition(), new TypeReference<>() {});
     Object outputs = cromwellService.getOutputs(run.engineId());
-    RecordAttributes outputParamDef = OutputGenerator.buildOutputs(outputDefinitionList, outputs);
+    Map<String, Object> outputParamDef =
+        OutputGenerator.buildOutputs(outputDefinitionList, outputs);
```
davidangb, line 76: `RecordAttributes` is gone, replaced by `Map<String, Object>` as a result of changing from swagger-codegen to openapi-generator. I am still investigating if this can be reverted.

davidangb: this is (hopefully) no longer true; resolving

service/build.gradle
```diff
@@ -21,14 +21,13 @@ dependencies {
...
     implementation 'org.apache.commons:commons-text:1.10.0'
-    implementation 'org.databiosphere:workspacedataservice-client:0.2.57-SNAPSHOT'
+    implementation 'org.databiosphere:workspacedataservice-client-okhttp-javax:0.2.97-SNAPSHOT'
     implementation 'org.springframework.boot:spring-boot-starter-data-jdbc'
     implementation 'org.springframework.boot:spring-boot-starter-web'
     implementation 'org.springframework.boot:spring-boot-starter-thymeleaf'
     implementation 'org.springframework.retry:spring-retry'
     implementation 'org.broadinstitute.dsde.workbench:sam-client_2.12:0.1-61135c7'
     implementation 'bio.terra:datarepo-client:1.313.0-SNAPSHOT'
-    implementation 'javax.ws.rs:javax.ws.rs-api:2.1.1'
```
davidangb, line 31: I believe the direct dependency on javax.ws.rs-api is not necessary.

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -680,7 +680,10 @@ void wdsRecordIdNotFoundTest() throws Exception {
     assertThat(
         result.getResponse().getContentAsString(),
         containsString(
-            "Error while fetching WDS Records for Record ID(s): {MY_RECORD_ID_2=ApiException thrown for testing purposes.}"));
+            "Error while fetching WDS Records for Record ID(s): {MY_RECORD_ID_2=Message: ApiException thrown for testing purposes.\\n"
+                + "HTTP response code: 400\\n"
+                + "HTTP response body: null\\n"
+                + "HTTP response headers: null}"));
```
davidangb, line 686: the new client appears to have changed error messages. I haven't tracked this down - does this need changing?

cjllanwarne: It's a little clunky but I don't think it's a blocker... unless the headers include things we don't want to be showing users or storing/logging

cjllanwarne (approved):

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
pshapiro4broad: I'm a bystander here but I would advocate making the string comparison less strict. Unless someone somewhere is dependent on exactly what the error message looks like (that is, the message is part of the API contract) I don't think you get value out of strictness here.

On the other hand, maybe we want to customize the response so it doesn't include all this extra text? Unless that's hard to do.

pshapiro4broad (approved):
Not sure I have approval on PRs here but I've been watching this from the sidelines as I plan to update other services to use the openapi codegen. LGTM

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
salonishah11: I like the idea of not having the extra text in the error response message but I am not sure how hard it would be to change the response class. In the short term maybe we can update the test to be less strict?

salonishah11 (approved):
Left 1 comment but LGTM. Thank you for making this change!

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
davidangb: I will look into this a bit more, see if I can at least figure out where/why this changed.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -403,20 +401,6 @@ public static List<String> validateRequestInputsAndOutputs(
...
-      if (error != null) {
-        return error.getMessage();
-      } else {
-        return exception.getMessage();
-      }
-    } catch (Exception e) {
-      return exception.getMessage();
-    }
-  }
-
```
davidangb, line 419: the new version of this moved to `WdsClientUtils` so it can be reused from multiple places

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -207,7 +208,7 @@ private Run updateDatabaseRunStatus(
                         updatableRun.recordId(),
                         updatableRun.runId(),
                         updatableRun.engineId(),
-                        e.getMessage());
+                        WdsClientUtils.extractErrorMessage(e.getMessage()));
```
davidangb, line 211: I'm not sure if there's anywhere else in the CBAS codebase you'd want to use this. I don't see any other test failures, but perhaps there are exception cases that don't have test coverage.

cjllanwarne (approved):
One thought about documentation via test cases but otherwise looks good to me. I like the reduction in noise from the messages

salonishah11 (approved):

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsClientUtils.java
```diff
@@ -0,0 +1,24 @@
...
+    if (excMsg == null) {
+      return "null";
+    }
+    int endPos = excMsg.indexOf("HTTP response code: ");
+    // if the exception message starts with "Message: ",
+    // endPos will be at least 9 characters from the start
+    if (endPos > 9 && excMsg.startsWith("Message: ")) {
+      return excMsg.substring(9, endPos - 1);
+    }
+    return excMsg;
```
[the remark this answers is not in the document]
davidangb: done! in 774840e14e8a8f88d608b89828e265c7cd71db92

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsClientUtils.java
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * Given an ApiException message from the WDS client, attempt to parse a human-readable error out
+   * of the formatted message. ApiException.getMessage() returns a String in the format: "Message:
+   * %s%nHTTP response code: %s%nHTTP response body: %s%nHTTP response headers: %s" and we'd like to
+   * get just that first %s, representing the actual message.
+   *
+   * @param excMsg the WDS client exception message
+   * @return the human-readable string, or the identity if it could not be parsed
+   */
+  public static String extractErrorMessage(String excMsg) {
```
pshapiro4broad, line 14: It looks like this is always called (from non-test code) with `e.getMessage()`. You could instead make this accept `Exception` and do the `e.getMessage()` here.

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsClientUtils.java
```diff
@@ -0,0 +1,26 @@
...
+    if (excMsg == null) {
```
pshapiro4broad, line 15: Does this ever happen in practice? I guess it couldn't hurt, and `Throwable.getMessage()` is definitely nullable.

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsClientUtils.java
```diff
@@ -0,0 +1,26 @@
...
+      return "null";
+    }
+    int endPos = excMsg.indexOf("HTTP response code: ");
+    // if the exception message starts with "Message: ",
+    // endPos will be at least 9 characters from the start
+    if (endPos > 9 && excMsg.startsWith("Message: ")) {
+      return excMsg.substring(9, endPos - 1);
+    }
```
pshapiro4broad, line 23: I think a constant here would make the code clearer.
```
  private static final String PREFIX = "Message: ";

...

    if (endPos > PREFIX.length() && excMsg.startsWith(PREFIX)) {
      return excMsg.substring(PREFIX.length(), endPos - 1);
    }
```

pshapiro4broad: Also, if you check for the prefix first, you don't have to check endPos against 9, which I think is slightly clearer still. Clear-ness is subjective of course...
```
    if (excMsg.startsWith(PREFIX)) {
      int endPos = excMsg.indexOf("HTTP response code: ");
      if (endPos != -1) {
        return excMsg.substring(PREFIX.length(), endPos - 1);
      }
    }
```

pshapiro4broad (approved):
I was hoping there was a okhttp configuration for this (maybe via subclassing?) but this gets the job done. It will be hard to make sure new code knows to extract messages in this way rather than calling `getMessage()` directly.

This is totally OK for now, but maybe there's a more heavyweight solution where all `ApiException`s are caught and re-thrown so the logic is localized to `WdsClient`.

PR #197, opened 2023-08-24
[WM-2091][WM-2092] Submit uses Leo cromwell

kpierre13:


kpierre13:
**Note for reviewers**

While reviewing, it would be helpful if you gave extra attention to:
1. The endpoint functions in `CromwellService.java` (for example) and note whether it makes sense that they are a read/write function.
2. If anything seems confusing and could be improved with a comment in the code.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -83,18 +83,27 @@ public WorkflowQueryResult runSummary(String runId) throws ApiException {
 
   public Object getOutputs(String id)
       throws ApiException, DependencyNotAvailableException, AzureAccessTokenException {
-    return cromwellClient.wesAPI().getRunLog(id).getOutputs();
+    ApiClient client = cromwellClient.getWriteApiClient();
```
mspector, line 86: (just to make sure we're on the same page: I'm operating under the assumption that these functions should use the "write client" when an operation is only available to creators/writers, and should use the "read client" when the operation is available to all user types)

should outputs be writer/creator only?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -83,18 +83,27 @@ public WorkflowQueryResult runSummary(String runId) throws ApiException {
...
+
+    return cromwellClient.wesAPI(client).getRunLog(id).getOutputs();
   }
 
-  public WorkflowDescription describeWorkflow(String workflowUrl) throws ApiException {
-    return cromwellConfig.womtoolApi().describe(API_VERSION, null, workflowUrl, null, null, null);
+  public WorkflowDescription describeWorkflow(String workflowUrl)
+      throws ApiException, DependencyNotAvailableException, AzureAccessTokenException {
+    ApiClient client = cromwellClient.getWriteApiClient();
```
mspector, line 93: same here - should this be available to everyone? (I have a feeling I'm missing something, but I figure I ought to ask!)

kpierre13: This one I'm on the fence about. @salonishah11 had a good reason for why this should be a write function over a read function. Saloni, do you remember what your thoughts were?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
mspector: Well, just thinking about this function in a vacuum: it seems to be reading some run information, which feels like it should be doable by readers. 

How `updateOutputAttributes`  uses this information feels like it should be a separate matter. Naively, I'd think that `updateOutputAttributes` should have its own way of enforcing "writers only"

mspector: I see that `updateOutputAttributes` is being used by the `SmartRunsPoller`, and that `SmartRunsPoller` is the object that owns the Cromwell client. Thinking aloud: if a user is a reader, and they're on the run details page (or whichever page shows run status), is their browser being updated by a `SmartRunsPoller` with a reader client or a writer client?

mspector: I hope I'm not leading you down the wrong path, but I'd be curious about how this would work... I'll ask at tech talk tomorrow, unless @cjllanwarne or @salonishah11 see this first

kpierre13: My assumption is that a reader can see what results the poller is reporting, but the work being done by the poller itself is being done on behalf of a cromwell belonging to a user that had permissions to do the actual functions. So the `SmartRunsPoller` doesn't directly update the browser, `SmartRunsPoller` updates the information in the databases that is then pulled down in the UI which continues to refresh the information. That's my understanding, but maybe it's not totally correct.

kpierre13: > Well, just thinking about this function in a vacuum: it seems to be reading some run information, which feels like it should be doable by readers.
> 
> How `updateOutputAttributes` uses this information feels like it should be a separate matter. Naively, I'd think that `updateOutputAttributes` should have its own way of enforcing "writers only"

I see what you mean, I think my logic here was that we are currently only using this api call in a place that writes back to the database, which is an action only writers should be able to do. For that reason, when we get the api client that will call the `getOutputs()` endpoint, we want that action to be taken by someone with write permissions. We may be able to enforce Sam permissions here, but I'd be curious as to what other people's perspectives are.

dillydally414: The SmartRunsPoller is also being modified by the workflow completion callback API, not sure if that has an impact on the relevance of this discussion

service/src/test/java/bio/terra/cbas/dependencies/leonardo/TestAppUtils.java
```diff
@@ -434,5 +465,77 @@ public TestAppUtils() throws IOException {
...
+                    "kubernetesRuntimeConfig": {
+                        "numNodes": 1,
+                        "machineType": "Standard_A2_v2",
+                        "autoscalingEnabled": false
+                    },
+                    "errors": [],
+                    "status": "RUNNING",
+                    "proxyUrls": {
+                        "cbas": "https://lzblahblahblah.servicebus.windows.net/cromwell-runner-app-${workspaceId}/cbas",
+                        "cbas-ui": "https://lzblahblahblah.servicebus.windows.net/cromwell-runner-app-${workspaceId}/",
```
dillydally414, line 488: Cromwell runner app won't have `cbas` or `cbas-ui`, and proxy url prefix will likely be either `cromwell-runner` or `cromwell-runner` (@salonishah11 would know better). Maybe doesn't affect the outcome of this test but probably good to keep our mock data consistent with the actual data

service/src/test/java/bio/terra/cbas/dependencies/leonardo/TestAppUtils.java
```diff
@@ -434,5 +465,77 @@ public TestAppUtils() throws IOException {
...
+                    "kubernetesRuntimeConfig": {
+                        "numNodes": 1,
+                        "machineType": "Standard_A2_v2",
+                        "autoscalingEnabled": false
+                    },
+                    "errors": [],
+                    "status": "RUNNING",
+                    "proxyUrls": {
+                        "cbas": "https://lzblahblahblah.servicebus.windows.net/terra-app-${workspaceId}/cbas",
+                        "cbas-ui": "https://lzblahblahblah.servicebus.windows.net/terra-app-${workspaceId}/",
```
dillydally414, line 524: Similar note to above, workflows app won't have cbas-ui and will have cromwell-reader instead of cromwell. Both will also have listener proxy URLs but I don't think those are necessary to include here. Workflows app will also have `"accessScope": "WORKSPACE_SHARED"`

service/src/test/java/bio/terra/cbas/dependencies/leonardo/TestAppUtils.java
kpierre13: I'll update for accuracy, and as far as outcome, it shouldn't matter too much, but once I know what it actually is, I can change that as well.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
salonishah11: The `describeWorkflow` endpoint is used in 2 places currently, one when a new method is being created and other when submitting a workflow. My initial thoughts were that it would make sense to call the individual Cromwell Runner / Writer especially for when submitting a workflow since the submission will be made to user's Cromwell rather than the Cromwell Reader.

kpierre13: And I'm assuming that we don't want a user with read access to be able to create new methods?

service/src/test/java/bio/terra/cbas/dependencies/leonardo/TestAppUtils.java
kpierre13: I'll change this to cromwell-runner then for accuracy 👍

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
salonishah11: Checking of read/write access will already happen before this endpoint is called. Here we just need to figure out whether we should call the user's Cromwell (which is Cromwell Runner) or the single reader/summarizer Cromwell that is created with Workflows app.

service/src/test/java/bio/terra/cbas/dependencies/leonardo/TestAppUtils.java
salonishah11: Confirming that it will be changed to `cromwell-runner`.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
```diff
@@ -31,15 +33,30 @@ public CromwellClient(
     singletonHttpClient = new ApiClient().getHttpClient();
   }
 
-  private ApiClient getApiClient()
+  public ApiClient getWriteApiClient()
       throws DependencyNotAvailableException, AzureAccessTokenException {
+    String uri =
+        dependencyUrlLoader.loadDependencyUrl(DependencyUrlLoader.DependencyUrlType.CROMWELL_URL);
```
dillydally414, line 39: Maybe we should check `testingBaseUri` here as well so that we can test locally against a single running cromwell?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
cjllanwarne: My $0.02: I would lean towards this being a "read client" action.

1. If it's being done during output updating in WDS it doesn't matter which Cromwell we read from
2. Reading from the "reader" should be significantly faster than the runner (no Leo/Sam/token/cross-network hops/.../etc to worry about)
3. If we reuse this in the future to allow a passthrough "getOutputs" endpoint in CBAS accessible to readers, it makes more sense to go to the reader Cromwell in the WORKFLOWS app.
4. There's a chance that we start shutting down the cromwell_runner app when the workflow is completed, but the reader Cromwell in the WORKFLOWS app should always be there to read outputs from

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
cjllanwarne: I also like this being a read action, for most of the same reasons - 
1. The reader should be much faster
2. The action will probably be called separately from submitting a workflow, so CBAS might not even have started the runner instance yet
3. Imagine a user importing a workflow for someone else to run. It doesn't make sense that the importing user has to start up a Cromwell instance just to run `describe`, then shut it down and have the second user start their runner to do the running

service/src/main/java/bio/terra/cbas/dependencies/common/DependencyUrlLoader.java
```diff
@@ -43,11 +43,9 @@ public String load(@NotNull DependencyUrlType key)
               throws DependencyNotAvailableException {
             if (key == DependencyUrlType.WDS_URL) {
               return fetchWdsUrl();
-            } else if (key == DependencyUrlType.CROMWELL_URL) {
+            } else {
```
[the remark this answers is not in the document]
cjllanwarne: Yes, absolutely, we definitely need to change this cache to be specific to the user. We should also make sure that even if a user _does_ get given the wrong Cromwell client from the cache by accident, they still can't use it because they have to supply their token each time and they'd have the wrong token

service/src/main/java/bio/terra/cbas/dependencies/leonardo/AppUtils.java
```diff
@@ -169,11 +173,18 @@ public String findUrlForWds(List<ListAppResponse> apps) throws DependencyNotAvai
   public String findUrlForCromwell(List<ListAppResponse> apps)
       throws DependencyNotAvailableException {
     ListAppResponse foundApp = findBestAppForAppType(apps, AppType.CROMWELL);
+    Object proxyUrl;
 
+    if (Objects.requireNonNull(foundApp.getProxyUrls()).containsKey("cromwell-writer")) {
+      proxyUrl = "cromwell-writer";
+    } else {
+      proxyUrl = "cromwell";
+    }
```
salonishah11, line 182: TOL: I am wondering if it would be better to figure out the proxy url key based on the app type of `foundApp` like this? Because it is possible that the `foundApp` doesn't have a `cromwell` or `cromwell-runner` key yet because it was not deployed properly 🤔 
```
If (foundApp.appType == CROMWELL_RUNNER_APP) {
  proxyUrlKey = "cromwell-runner"
}
else {
  proxyUrlKey = "cromwell"
}
```

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
```diff
@@ -31,15 +33,30 @@ public CromwellClient(
...
+
+    ApiClient apiClient = new ApiClient().setBasePath(uri);
+    apiClient.setHttpClient(singletonHttpClient);
+    apiClient.addDefaultHeader(
+        "Authorization",
+        "Bearer " + credentialLoader.getCredential(CredentialLoader.CredentialType.AZURE_TOKEN));
```
salonishah11, line 50: Just to clarify my understanding, this will in later ticket(s) be changed to using the token passed in the user's request right? 🤔

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
```diff
@@ -31,15 +33,30 @@ public CromwellClient(
     singletonHttpClient = new ApiClient().getHttpClient();
   }
 
-  private ApiClient getApiClient()
+  public ApiClient getWriteApiClient()
```
cjllanwarne, line 36: Like @dillydally414 did with having thread local beans for the user token, I wonder whether we could have thread-local write api clients? That would mean it would be impossible for us to inappropriately store-and-retrieve the wrong client in a downstream cache (which is a concern that it'd be nice to make impossible rather than having to logic check every time we use a cache)

service/src/main/java/bio/terra/cbas/dependencies/common/DependencyUrlLoader.java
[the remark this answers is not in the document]
kpierre13: If it's okay with both of you, I'd like to do that as part of the next ticket since that adds "as the user" functionality to cbas.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
salonishah11: I agree with what Chris said. I think the way I am thinking this that is if we need to look up any metadata or outputs or do any read from database related to a workflow, we want to use the Cromwell reader / read client so that it is accessible to the users without auth. The things we would go to user's Cromwell Runner is when submitting a workflow or aborting a workflow because user's individual Cromwell would be performing that action.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -83,18 +83,27 @@ public WorkflowQueryResult runSummary(String runId) throws ApiException {
...
+    return cromwellClient
+        .womtoolApi(client)
+        .describe(API_VERSION, null, workflowUrl, null, null, null);
   }
 
-  public String getRunErrors(Run run) throws ApiException {
+  public String getRunErrors(Run run)
+      throws ApiException, DependencyNotAvailableException, AzureAccessTokenException {
+
+    ApiClient client = cromwellClient.getWriteApiClient();
```
salonishah11, line 101: I think here also we might want to use the read client since we are getting updates related to workflow?

service/src/main/resources/application.yml
```diff
@@ -69,7 +71,8 @@ wds:
 
 leonardo:
   baseUri: "https://leonardo.dsde-dev.broadinstitute.org/"
-  wdsAppTypeNames: ['WDS', 'CROMWELL']
+  wdsAppTypeNames: ['WDS', 'CROMWELL'] # keeping CROMWELL type for backwards compatibility
```
salonishah11, line 74: I think it might be helpful to add a comment on what these values are for since the apps are listed in priority order for finding proxy url

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
kpierre13: Yes, that's correct

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
dillydally414: @kpierre13 Is this something you plan to address in the follow up ticket to use the appropriate user's cromwell rather than in this PR?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
```diff
@@ -54,16 +55,37 @@ private ApiClient getApiClient()
...
+    ApiClient apiClient = new ApiClient().setBasePath(cromwellServerConfiguration.baseUri());
+    apiClient.setHttpClient(singletonHttpClient);
+    apiClient.addDefaultHeader(
+        "Authorization",
+        "Bearer " + credentialLoader.getCredential(CredentialLoader.CredentialType.AZURE_TOKEN));
+    // By closing the connection after each request, we avoid the problem of the open connection
+    // being force-closed ungracefully by the Azure Relay/Listener infrastructure:
+    apiClient.addDefaultHeader("Connection", "close");
+    apiClient.setDebugging(cromwellServerConfiguration.debugApiLogging());
+    return apiClient;
```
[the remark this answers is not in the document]
kpierre13: Is it worth doing in this ticket if they're going to be removed in the ticket that follows this one?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
kpierre13: Yes, this work is currently implemented in the next PR

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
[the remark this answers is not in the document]
dillydally414: Probably not 😃 Related question though - are there going to be any issues with the read and write clients using the same `singletonHttpClient`? I'm not sure exactly how that works...

dillydally414 (approved):

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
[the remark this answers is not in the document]
kpierre13: I'm not either... 🤔 @cjllanwarne any thoughts?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
cjllanwarne: I think we could still cache the URLs separately based on user, just not the entire client (which comes pre-loaded with the tokens/authorization)

cjllanwarne (approved):

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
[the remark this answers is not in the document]
cjllanwarne: I wondered about this but I don't think it's a problem. The problem would be sharing a client which has had the default auth tokens attached to it, which is above the level of the http client.

Or if you still feel nervous, you could write a unit test to validate that

PR #199, opened 2023-08-29
[WM-2147] Cromwell lookup as user

kpierre13:


service/src/main/java/bio/terra/cbas/config/BeanConfig.java
```diff
@@ -41,4 +43,13 @@ public ObjectMapper objectMapper() {
   public BearerToken bearerToken(HttpServletRequest request) {
     return new BearerTokenFactory().from(request);
   }
+
+  @Bean("cromwellWriteClient")
+  @RequestScope
+  public cromwell.client.ApiClient cromwellWriteClient(
```
dillydally414, line 49: Do we need the `HttpServletRequest` here?

service/src/main/java/bio/terra/cbas/dependencies/common/ScheduledHealthChecker.java
```diff
@@ -27,15 +26,10 @@ public class ScheduledHealthChecker {
...
   public ScheduledHealthChecker(
-      CromwellService cromwellService,
-      WdsService wdsService,
-      LeonardoService leonardoService,
-      SamService samService) {
+      CromwellService cromwellService, LeonardoService leonardoService, SamService samService) {
     this.healthCheckStatuses = new ConcurrentHashMap<>();
     this.healthCheckSystems =
         Map.of(
-            "cromwell", cromwellService,
```
dillydally414, line 37: Can we continue checking the cromwell reader status?

service/src/main/java/bio/terra/cbas/dependencies/leonardo/LeonardoClient.java
```diff
@@ -26,4 +26,10 @@ public ApiClient getApiClient() throws AzureAccessTokenException {
     apiClient.setDebugging(leonardoServerConfiguration.debugApiLogging());
     return apiClient;
   }
+
+  public ApiClient getApiClient(String accessToken) throws AzureAccessTokenException {
```
dillydally414, line 30: Do we still need the `AzureAccessTokenException` if we're always using the user's bearer token?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
```diff
@@ -19,22 +17,18 @@ public class CromwellClient {
...
+      DependencyUrlLoader dependencyUrlLoader) {
     this.cromwellServerConfiguration = cromwellServerConfiguration;
     this.dependencyUrlLoader = dependencyUrlLoader;
-    this.credentialLoader = credentialLoader;
     singletonHttpClient = new ApiClient().getHttpClient();
   }
 
-  public ApiClient getWriteApiClient()
-      throws DependencyNotAvailableException, AzureAccessTokenException {
+  public ApiClient getWriteApiClient(String accessToken) throws DependencyNotAvailableException {
```
dillydally414, line 31: Is this still needed or is it replaced by the `cromwellWriteClient` in CromwellService?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -81,21 +83,20 @@ public WorkflowQueryResult runSummary(String runId)
     }
   }
 
-  public Object getOutputs(String id) throws ApiException, AzureAccessTokenException {
+  public Object getOutputs(String id) throws ApiException {
```
dillydally414, line 86: Should these methods also be passing in the bearerToken?

service/src/main/resources/application.yml
```diff
@@ -1,5 +1,6 @@
 logging:
   level.bio.terra.cbas: ${BATCHANALYSIS_LOG_LEVEL:info}
+  pattern.level: '%X{requestId} %5p'
```
dillydally414, line 3: Is there a reason this was added back in?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
dillydally414: Nvm, I see that this is used in the bean creation.

service/src/main/resources/application.yml
kpierre13: Accidental from merging with main. I'll remove this again.

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
kpierre13: Do we need to be passing the bearerToken for functions using the read client?

dillydally414: I would assume so? Not sure if the listener will allow requests to the reader cromwell without a token. @cjllanwarne any thoughts?

service/src/main/java/bio/terra/cbas/dependencies/leonardo/LeonardoClient.java
kpierre13: We do if we are still using the above `getAptClient()`as it still utilizes the AzureToken in the default header. As part of this PR, should we remove the default header here and utilize `.setAccessToken()` instead?

service/src/main/java/bio/terra/cbas/config/BeanConfig.java
kpierre13: @cjllanwarne did we decide what to do with the `HttpServletRequest` here?

service/src/main/java/bio/terra/cbas/dependencies/leonardo/LeonardoClient.java
dillydally414: My instinct would be yes since CBAS shouldn't need to be aware of an AzureToken any longer... probably good for someone else to confirm

service/src/main/java/bio/terra/cbas/config/BeanConfig.java
cjllanwarne: I believe it's not needed here because it's already provided to the BearerToken bean constructor which has the same `RequestScope` annotation

service/src/main/java/bio/terra/cbas/dependencies/leonardo/LeonardoClient.java
cjllanwarne: When I tried `setAccessToken` it didn't work or didn't seem to always apply it - hence switching to the default header. If you've double checked it's definitely getting added properly then 👍

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
```diff
@@ -44,31 +38,32 @@ public ApiClient getWriteApiClient()
...
-        "Bearer " + credentialLoader.getCredential(CredentialLoader.CredentialType.AZURE_TOKEN));
     // By closing the connection after each request, we avoid the problem of the open connection
     // being force-closed ungracefully by the Azure Relay/Listener infrastructure:
     apiClient.addDefaultHeader("Connection", "close");
     apiClient.setDebugging(cromwellServerConfiguration.debugApiLogging());
     return apiClient;
   }
 
-  public ApiClient getReadApiClient() throws AzureAccessTokenException {
+  public ApiClient getReadApiClient() {
```
cjllanwarne, line 50: Presumably only one of these (with or without access token) is actually needed?

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
cjllanwarne: It depends whether the traffic from CBAS to the reader needs to leave the app and come back in via the listener or not. Which depends on whether the app is set up with the cromwell URL being the full path or a relative in-app path. I _believe_ it's all in-app but could be wrong. In that case, no tokens would need to be forwarded on

cjllanwarne (approved):

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
dillydally414: Ok, so then we could get rid of the `getReadApiClient` signature with a bearer token and remove it from L59 above?

dillydally414 (approved):
🎉

PR #201, opened 2023-09-12
[WM-2208] Initialize runs with INITIALIZING state

dillydally414:
Small change proposed to address the rocky UI experience immediately following a workflow submission.
New experience:

https://github.com/DataBiosphere/cbas/assets/67511512/b1d9357c-e634-4fcc-9fcc-f102b6d07e2e

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -6,7 +6,7 @@ TODO: Complete me! Should be of similar quality to the ECM documentation.
 
 For compliance reasons, all pull requests **must** be submitted with a Jira ID in the pull request title.
 You should include the Jira ID near the beginning of the title for better readability.
-For example: "[WM-1992] add statement to README.md and DEVELOPMENT.md about including Jira IDs in PR titles" 
+For example: "[WM-1992] add statement to README.md and DEVELOPMENT.md about including Jira IDs in PR titles"
 
 If there is more than one relevant ticket, include all of their Jira IDs.
 For example: "WM-1997, WM-2002, WM-2005: Fix for many bugs with the same root cause"
@@ -185,9 +185,10 @@ Then navigate to the Swagger: `http://localhost:8080/swagger-ui.html`
 
 New images are built and pushed automatically each time a PR is merged. Developers may
 want to do this manually in order to test in-development CBAS versions in non-local
-environments. 
+environments.
 
 First, log into GCP using `gcloud auth login`. You need this to authenticate to push the image.
+You may also need to run `gcloud auth configure-docker` prior to pushing the image.
 
 ```
 TAG="my-custom-image-name"
--- a/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
@@ -8,8 +8,8 @@ import static bio.terra.cbas.common.MetricsUtil.recordRunsSubmittedPerRunSet;
 import static bio.terra.cbas.model.RunSetState.CANCELING;
 import static bio.terra.cbas.model.RunSetState.ERROR;
 import static bio.terra.cbas.model.RunSetState.RUNNING;
+import static bio.terra.cbas.models.CbasRunStatus.INITIALIZING;
 import static bio.terra.cbas.models.CbasRunStatus.SYSTEM_ERROR;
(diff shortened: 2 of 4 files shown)


DEVELOPMENT.md
```diff
@@ -185,13 +185,14 @@ Then navigate to the Swagger: `http://localhost:8080/swagger-ui.html`
 
 New images are built and pushed automatically each time a PR is merged. Developers may
 want to do this manually in order to test in-development CBAS versions in non-local
-environments. 
+environments.
 
 First, log into GCP using `gcloud auth login`. You need this to authenticate to push the image.
+You may also need to run `gcloud auth configure-docker` prior to pushing the image.
```
dillydally414, line 191: Ran into this issue while testing so wanted to add a tip here for future reference.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -484,7 +484,8 @@ private List<RunStateResponse> buildInputsAndSubmitRun(
             cromwellService.submitWorkflow(rawMethodUrl, workflowInputs, workflowOptionsJson);
 
         runStateResponseList.add(
-            storeRun(runId, workflowResponse.getRunId(), runSet, record.getId(), UNKNOWN, null));
+            storeRun(
+                runId, workflowResponse.getRunId(), runSet, record.getId(), INITIALIZING, null));
```
dillydally414, line 488: When we first store the run in our database, I think it should have the INITIALIZING state. If there is an issue submitting the workflow to cromwell above, this code will not be reached as we won't even have a workflowResponse. In this case, the status will be set to SYSTEM_ERROR. By setting it to initializing, we are more accurately reflecting our current knowledge of the state - we have submitted the workflow to cromwell, no issues yet, but it hasn't started running to our knowledge.

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -160,7 +160,7 @@ private Run tryUpdateRun(Run r) {
       recordOutboundApiRequestCompletion("wes/runSummary", getStatusStartNanos, getStatusSuccess);
     }
 
-    CbasRunStatus newStatus = CbasRunStatus.INITIALIZING;
+    CbasRunStatus newStatus = CbasRunStatus.UNKNOWN;
```
dillydally414, line 163: Then, when we want to refresh the status of a particular run, it should default to UNKNOWN if we have no info about said workflow yet, rather than INITIALIZING. This agrees with the expected behavior of UNKNOWN representing missing information from Cromwell, rather than INITIALIZING representing "we think it's been started but it could be an actual error". INITIALIZING should then only correspond to "we sent the workflow to be executed" and last until we receive an actual update about the workflow status (even if that status is UNKNOWN).

salonishah11: TOL: Should we set the `newStatus` here to be same as the current status of Run that we have in CBAS database? This way if CBAS was unable to fetch updated status from Cromwell, the new status would be same as the previous one from CBAS database. This would be a better user experience. Otherwise there is a chance that we set the status of Run from `Initializing` -> `Running` -> `Unknown` -> terminal state no? Which would be a weird user experience.

dillydally414: My thought was that this should represent the status to the best of our current knowledge. I think we should set it to UNKNOWN if we don't know anything about the current status as I don't see any reason Cromwell would go from running -> not having any idea about the specific workflow -> terminal state, it should just go from running -> terminal state or running -> not having any idea about the specific workflow (if something goes horribly wrong).

salonishah11: Based on the code it seems if we don't receive a response from Cromwell or get an error while fetching the status CBAS would set the status to `Unknown` instead of its current state. I also feel we should not go from a `Running` or `Initializing` state to `Unknown`. But I am interested in hearing from other reviewers. @cjllanwarne what do you suggest?

cjllanwarne (approved):
LGTM though I would recommend just quickly syncing with Maria before merging it, to avoid surprise conflicts

kpierre13 (approved):

PR #203, opened 2023-09-21
[WM-1792] Submit workflows as batch to Cromwell

dillydally414:
For timing comparison, ran `fetch_sra_to_bam` with 100 workflows with the current version of CBAS and with a version built from this branch:
- current/individual submissions: ~45 seconds for network response
- new/batched submissions: ~16 seconds for network response


cjllanwarne (approved):

cjllanwarne:
TOL it might be nice - if possible - to limit based on keeping under total request size (in MB) rather than (or as well as) a counter of workflows. That way we can be a bit more resilient to small numbers of giant workflows which might break the intermediate network layers.

kpierre13 (approved):

dillydally414:
Looked into limiting by size of request, doesn't seem like this is something well-supported by Java/Spring.

PR #211, opened 2023-10-04
[WM-2090] Remove workflow completion handling from SmartPoller

zykonda:
1. Removed output processing from SmartRunPoller, added the runCompletionHandler instead of objectMapping, runsDao and WdsService references.
2. Restructured functional and unit tests to reflect the class structure.


service/src/main/java/bio/terra/cbas/runsets/outputs/OutputGenerator.java
```diff
@@ -17,6 +17,7 @@
 
 public class OutputGenerator {
 
+  @SuppressWarnings("unchecked")
```
zykonda, line 20: Added to suppress Unchecked cast: 'java.lang.Object' to 'java.util.Map<java.lang.String,java.lang.Object>'

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -208,83 +165,40 @@ private Run updateDatabaseRunStatus(
       CbasRunStatus status, OffsetDateTime engineStatusChanged, Run updatableRun) {
     long updateDatabaseRunStatusStartNanos = System.nanoTime();
     boolean updateDatabaseRunStatusSuccess = false;
+    ArrayList<String> errors = new ArrayList<>();
+    Object outputs = null;
 
     try {
       var updatedRunState = status;
-      if (updatableRun.status() != updatedRunState) {
```
cjllanwarne, line 214: I'm loving seeing all this simplification in the API handler! 💯 !

service/src/main/java/bio/terra/cbas/runsets/results/RunCompletionHandler.java
```diff
@@ -68,93 +68,123 @@ public void updateOutputAttributes(Run run, Object outputs)
...
     return updateResult;
   }
 
-  /*
-   The method checks if output definitions are associated with run and makes updates.
-   If outputs passed as arguments are null, but the workflow has the output definitions,
-   then call to Cromwell will be made to pull the outputs.
-  */
-  private String saveWorkflowOutputs(Run updatableRun, Object outputs) {
+  private boolean saveWorkflowOutputsToWds(Run updatableRun, Object workflowOutputs)
```
cjllanwarne, line 138: And can we maybe add a good set of unit tests for this function? It feels like a great place to make sure we're doing the right thing for all the wide variety of possible input arguments for `Run` and `Object`?

cjllanwarne: And thinking out loud about making unit testing easier, I wonder whether it would make sense to break this function into two halves - one which is more like a functional mapping that takes these inputs and works out, as a batch, what WDS updates to make (which should be easy to unit test), and then a second half which turns all those WDS updates into actual API calls (which can probably just be a single simple test)? Just a thought!

cjllanwarne (approved):
Looks good, I had a couple of thoughts around whether we can unit test that one specific function a little more comprehensively (and ideally outside of "set up everything / check the final result" style API tests), but not enough to block the PR

service/src/main/java/bio/terra/cbas/runsets/results/RunCompletionHandler.java
```diff
@@ -126,30 +123,50 @@ public RunCompletionResult updateResults(
...
   }
 
-  /*
-   The method checks if output definitions are associated with run and makes updates.
-   If outputs passed as arguments are null, but the workflow has the output definitions,
-   then call to Cromwell will be made to pull the outputs.
-  */
-  private String saveWorkflowOutputs(Run updatableRun, Object outputs) {
+  @SuppressWarnings("unchecked")
+  private boolean saveWorkflowOutputsToWds(Run updatableRun, Object workflowOutputs)
```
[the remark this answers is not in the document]
zykonda: refactored again.
I have found that tests for this class need to swap names: Functional <-> Unit.

mspector (approved):

PR #213, opened 2023-10-11
[WM-2174] runSetsApiController gates post/abort on write (not compute) permission

mspector:


service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -349,10 +349,6 @@ void setupFunctionalChecks() throws Exception {
             });
 
     doReturn(mockUser).when(samService).getSamUser();
-
-    // setup Sam permission check to return true
-    doReturn(true).when(samService).hasReadPermission();
-    doReturn(true).when(samService).hasComputePermission();
```
mspector, line 355: these upfront mocks were causing some issues in subsequent tests that overwrite the return values - I thought it would be clearer to remove them

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -1100,7 +1096,7 @@ void oneFailedOneSucceededRun() throws Exception {
 
   @Test
   void returnErrorForUserWithNoReadAccess() throws Exception {
-    when(samService.hasReadPermission()).thenReturn(false);
+    doReturn(false).when(samService).hasReadPermission();
```
mspector, line 1099: I'm not 100% sure why, but these mocks were not overriding earlier return values when using `when`. `doReturn` resolved the problem.

only these changes seem to be necessary to make the tests pass, but I thought it was worth removing the upfront setup anyway for clarity

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -307,9 +305,8 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
 
   @Override
   public ResponseEntity<AbortRunSetResponse> abortRunSet(UUID runSetId) {
-    // check if current user has compute permissions on the workspace
-    if (!samService.hasComputePermission()) {
-      throw new ForbiddenException(SamService.COMPUTE_ACTION, SamService.RESOURCE_TYPE_WORKSPACE);
+    if (!samService.hasWritePermission()) {
```
cjllanwarne, line 308: TOL abort is interesting... I probably wouldn't want to start a new app just to abort a workflow, but if an app _is_ already running I don't see too much harm in sending the request along

cjllanwarne (approved):

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
dillydally414: I've ran into this funkiness before, from my experience it seems like in some cases `thenReturn` will call the actual method when creating the mock/spy whereas the `doReturn` doesn't have that behavior

dillydally414 (approved):

PR #214, opened 2023-10-11
[WM-2306] Add user ID to GET RunSet response

dillydally414:


zykonda (approved):
LGTM

salonishah11 (approved):
Do we need AppSec approval for this since it involves user ID? Its possible we don't because this is ID and not any actual user data 🤔

dillydally414:
> Do we need AppSec approval for this since it involves user ID? Its possible we don't because this is ID and not any actual user data 🤔

Good point - I will ping them on the ticket to see if they have any qualms.

PR #216, opened 2023-10-16
[WM-2267] Add show_all_users to GET run_sets

dillydally414:
Prerequisite for manual shutdown. Allows optional specification of `show_all_users=false` (default is `true`) to filter run set output to only show runsets that the user making the request submitted.

Also adds UserStatusInfo `userInfo` bean instantiated once per request just like bearerToken, to avoid unnecessary repeat calls to Sam.

### DIFF
--- a/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
@@ -142,7 +142,8 @@ public class RunSetsApiController implements RunSetsApi {
   }
 
   @Override
-  public ResponseEntity<RunSetListResponse> getRunSets(UUID methodId, Integer pageSize) {
+  public ResponseEntity<RunSetListResponse> getRunSets(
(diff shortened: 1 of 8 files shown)


service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -157,6 +161,13 @@ public ResponseEntity<RunSetListResponse> getRunSets(UUID methodId, Integer page
       filteredRunSet = runSetDao.getRunSets(pageSize, false);
     }
 
+    if (Boolean.FALSE.equals(showAllUsers)) {
+      filteredRunSet =
+          filteredRunSet.stream()
+              .filter(runSet -> runSet.userId().equals(userInfo.getUserSubjectId()))
```
salonishah11, line 167: TOL: I wonder if it would be better to filter based on User ID at the database layer rather than here. Apart from it being little faster it would also mean sending less data from database to application layer.

salonishah11 (approved):

PR #217, opened 2023-10-16
[WM-2277] GHA to run Azure E2E test

salonishah11:
This PR adds a Github action to run Azure E2E script being added as part of https://github.com/broadinstitute/dsp-reusable-workflows/pull/6.

Link to successful run- https://github.com/DataBiosphere/cbas/actions/runs/6597974881

Note: This PR should be merged after https://github.com/broadinstitute/dsp-reusable-workflows/pull/6 is merged

JIRA ticket: https://broadworkbench.atlassian.net/browse/WM-2277


.github/workflows/azure_e2e_test.yml
```diff
@@ -1,27 +1,19 @@
...
-on:
+'on':
+  schedule:
+    # run twice a day at 10:00 and 22:00 UTC every day of the week
+    - cron: "0 10/12 * * 1-5"
   workflow_dispatch:
     inputs:
       target-branch:
         description: 'Branch of CBAS to run tests on'
         required: true
```
dillydally414, line 10: Does this still need to be required if there is a default value provided? Also, CBAS has a `main` branch not a `develop` branch

dillydally414 (approved):

.github/workflows/azure_e2e_test.yml
salonishah11: Oops this should definitely be main. I haven't been using this in testing but my though behind this was that if in future we want to run this test against PRs passing a branch would be easier. But on second thought I guess when we get to it there would be other changes needed to make that work so I guess I can remove that for now. 🤔

mspector (approved):
other than the outstanding TODO, LGTM!

.github/workflows/azure_e2e_test.yml
```diff
@@ -92,37 +73,61 @@ jobs:
...
-          access_token_scopes: 'profile, email, openid'
-          access_token_subject: ${{ inputs.owner-subject }}
-          export_environment_variables: false
-          create_credentials_file: false
-      - uses: actions/checkout@v3
-        with:
-          #NOTE: replace with ${{inputs.target-branch}} before merge
-          #TODO: this might change of the python script is moved to dsp-reusable-workflows
-          ref: refs/heads/sps_azure_e2e_test
+          ref: sps_cbas_e2e_test #TODO: switch to main
```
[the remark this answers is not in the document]
salonishah11: This is waiting for the other PR to merge. Then I will update branch to main

.github/workflows/azure_e2e_test.yml
```diff
@@ -1,40 +1,16 @@
...
-      owner-subject:
-        description: 'Owner of billing project'
-        required: true
-        default: 'hermione.owner@quality.firecloud.org'
-        type: string
-      service-account:
-        description: 'Email address or unique identifier of the Google Cloud service account for which to generate credentials'
-        required: true
-        default: 'firecloud-qa@broad-dsde-qa.iam.gserviceaccount.com'
-        type: string
```
cjllanwarne, line 20: Am I reading this right, that manual triggering is still possible, it just always generates all of the parameters itself now, rather than needing us to supply them?

.github/workflows/azure_e2e_test.yml
```diff
@@ -78,51 +60,44 @@ jobs:
...
-        run: |
-          poetry run e2e-test/azure_e2e_test.py
+          workflow: bee-destroy
+          repo: broadinstitute/terra-github-workflows
+          ref: refs/heads/main
+          wait-for-completion: false
+          token: '${{ env.BROADBOT_TOKEN }}'
+          inputs: '{ "bee-name": "${{ env.BEE_NAME }}" }'
+
+  report-workflow:
```
cjllanwarne, line 98: Does this also (like `delete_bee`) need some `needs` and `if` parameters to make sure it always runs, and runs _after_ everything else?

.github/workflows/azure_e2e_test.yml
salonishah11: Yes that is correct. These values are static and had to be setup in tfvars file in terraform-ap-deployments repo. Hence I removed them from here. The manual triggering of workflow is still possible.

.github/workflows/azure_e2e_test.yml
cjllanwarne: Based on [your example](https://github.com/DataBiosphere/cbas/actions/runs/6597974881), it looks like the report happened while the status was still `queued`

salonishah11: Based on my understanding this runs separately from the test run pipeline and runs in parallel to the `params-gen` job. So this step is always run even before the test starts. My understanding is that the step basically tells sherlock that monitor this new workflow run and send Slack notification when the run fails. 
https://github.com/broadinstitute/sherlock/blob/fd3ad8406edd0223713fb59a7124354125a91a1b/.github/workflows/client-report-workflow.yaml#L30-L33

cjllanwarne: oh I see, ok 👍

cjllanwarne (approved):

dillydally414 (approved):

PR #218, opened 2023-10-24
[WM-2191] Contract tests cbas consumer sam provider

zykonda:
Requests from cbas-consumer to sam-provider:
1. A request for the user's status given user status info request with access token.
2. A request for write permission on workspace given user has write permission.
3. A status request given sam is ok.

Interactions
Given user status info request with access token, upon receiving a request for the user's status from cbas-consumer, with
{
  "method": "GET",
  "path": "/register/user/v2/self/info",
  "headers": {
    "Authorization": "Bearer accessToken"
  }
}
sam-provider will respond with:
{
  "status": 200,
  "headers": {
    "Content-Type": "application/json; charset=UTF-8"
  },
  "body": {
    "userSubjectId": "testUser",
    "userEmail": "test@test.com",
    "enabled": true
  }
}
Given user has write permission, upon receiving a request for write permission on workspace from cbas-consumer, with
{
  "method": "GET",
  "path": "/api/resources/v2/workspace/15f36863-30a5-4cab-91f7-52be439f1175/action/write",
  "headers": {
    "Authorization": "Bearer accessToken"
  }
}
sam-provider will respond with:
{
  "status": 200,
  "body": true
}
Given sam is ok, upon receiving a status request from cbas-consumer, with
{
  "method": "GET",
  "path": "/status"
}
sam-provider will respond with:
{
  "status": 200,
  "headers": {
    "Content-Type": "application/json; charset=UTF-8"
  },
  "body": {
    "ok": true
  }
}

### DIFF
--- a/.github/workflows/consumer-contract-tests.yml
+++ b/.github/workflows/consumer-contract-tests.yml
@@ -1,0 +1,103 @@
+name: Consumer contract tests
+# The purpose of this workflow is to validate the service level contract
+# using the Pact framework.
+#
+# More details on Contract Testing can be found in our handbook
+#
+# https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2660368406/Getting+Started+with+Pact+Contract+Testing
+#
+#
+#
+# NOTE: The publish-contracts workflow will use the latest commit of the branch that triggers this workflow to publish the unique consumer contract version to Pact Broker.
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
+  push:
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
+  merge_group:
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
+
+jobs:
+  init-github-context:
+    runs-on: ubuntu-latest
+    outputs:
+      repo-branch: ${{ steps.extract-branch.outputs.repo-branch }}
+      repo-version: ${{ steps.extract-branch.outputs.repo-version }}
+
+    steps:
+      - uses: actions/checkout@v3
+      - id: extract-branch
+        run: |
+          GITHUB_EVENT_NAME=${{ github.event_name }}
+          if [[ "$GITHUB_EVENT_NAME" == "push" ]]; then
+            GITHUB_REF=${{ github.ref }}
+            GITHUB_SHA=${{ github.sha }}
+          elif [[ "$GITHUB_EVENT_NAME" == "pull_request" ]]; then
+            GITHUB_REF=refs/heads/${{ github.head_ref }}
+            GITHUB_SHA=${{ github.event.pull_request.head.sha }}
+          elif [[ "$GITHUB_EVENT_NAME" == "merge_group" ]]; then
+            GITHUB_REF=refs/heads/${{ github.head_ref }}
+          else
+            echo "Failed to extract branch information"
+            exit 1
+          fi
+          echo "repo-branch=${GITHUB_REF/refs\/heads\//""}" >> $GITHUB_OUTPUT
+          echo "repo-version=${GITHUB_SHA}" >> $GITHUB_OUTPUT
+      - name: Echo repo and branch information
+        run: |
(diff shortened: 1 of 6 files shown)


service/src/test/java/bio/terra/cbas/pact/SamPactTest.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  private void initSamService(MockServer mockServer) {
+    SamServerConfiguration samConfig =
+        new SamServerConfiguration(mockServer.getUrl(), false, dummyWorkspaceId, true);
+    SamClient samClient = new SamClient(samConfig);
+    samService = new SamService(samClient, new BearerToken("accessToken"));
+  }
+
+  @Pact(consumer = "cbas-consumer", provider = "sam-provider")
```
mspector, line 50: this naming convention - `${SERVICE}-consumer` or `${SERVICE}-provider` - is flawed, despite it being the most common naming convention. I'm working on getting services to rename their contracts. We should all just use the service name: `cbas` in this case.

You may have to keep the provider name `sam-provider` until they change their code, unfortunately...

mspector: (the reason it is flawed is because there are certain pieces of functionality in the Pact broker that rely on identifying a service regardless of whether it is playing the role of consumer or provider. that functionality is broken for all the services that use the flawed naming convention)

mspector: I'm saying: there's really no reason for us to contribute to this flawed practice by perpetuating it.

zykonda: I can easily change cbas-consumer to cbas, but I cannot change sam-provider to sam as I will hit check_for_potential_duplicate_pacticipant_names rule. And this change needs to be sync'd with everybody on DSP pact.

.github/workflows/consumer-contract-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+# More details on Contract Testing can be found in our handbook
+#
+# https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2660368406/Getting+Started+with+Pact+Contract+Testing
+#
+#
+#
+# NOTE: The publish-contracts workflow will use the latest commit of the branch that triggers this workflow to publish the unique consumer contract version to Pact Broker.
+
+on:
+  pull_request:
```
dillydally414, line 14: Shouldn't we only be publishing consumer contracts after merge, not on PR? @mspector any thoughts here? Comparing to Terra-UI's workflow for reference: https://github.com/DataBiosphere/terra-ui/blob/dev/.github/workflows/publish-pacts.yaml

.github/workflows/consumer-contract-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+    paths-ignore: [ '**.md' ]
+  push:
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
+  merge_group:
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
+
+jobs:
+  init-github-context:
```
dillydally414, line 25: If we only run on push to main, this step would also be unnecessary

service/src/test/java/bio/terra/cbas/pact/SamPactTest.java
mspector: Yeah, agreed. If we can change `cbas-consumer` to `cbas` that would be great.

Sorry for sticking on this minor point. This has been an uphill battle since we started using Pact.

zykonda: np, I have already tested. Closing this conversation then.

mspector (approved):

.github/workflows/consumer-contract-tests.yml
mspector: ah, yes, good catch. @dillydally414 is absolutely right

dillydally414 (approved):

PR #222, opened 2023-11-09
[WM-2291] Callback API contract tests between Cromwell and CBAS

zykonda:
Added provider state for workflow callback


salonishah11 (approved):

cjllanwarne (approved):
This looks fine, but appears to only be adding additional states, I don't see where we are actually using them. Is that right?

PR #224, opened 2023-12-01
[WM-1865] Surface Cromwell-level error in response

dillydally414:
See thread [here](https://broadinstitute.slack.com/archives/GHYJZ2ZE0/p1701379476185339) for more info. Currently the only fix in place is to surface the internal error in addition to the generic message. If users are still seeing this in the new multi-user app layout, the Cromwell client/spec may need to be updated with the proper WorkflowMetadataResponse object structure

### DIFF
--- a/service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
+++ b/service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
@@ -153,10 +153,11 @@ public class SmartRunsPoller {
       }
     } catch (Exception e) {
       String errorMessage =
-          "Error fetching Cromwell-level error from Cromwell for run %s."
-              .formatted(updatableRun.runId());
+          "Error fetching Cromwell-level error from Cromwell for workflow %s."
+              .formatted(updatableRun.engineId());
       logger.error(errorMessage, e);
       errors.add(errorMessage);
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -153,10 +153,11 @@ private List<String> getWorkflowErrors(Run updatableRun) {
       }
     } catch (Exception e) {
       String errorMessage =
-          "Error fetching Cromwell-level error from Cromwell for run %s."
-              .formatted(updatableRun.runId());
+          "Error fetching Cromwell-level error from Cromwell for workflow %s."
```
cjllanwarne, line 156: Since this is already attached to a run record, do we really need the workflow ID in the message itself?

cjllanwarne (approved):

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
salonishah11: TOL: Should we make the error message DRY? It seems we are repeating Cromwell here. Maybe something like this?
```Error fetching Cromwell-level error for workflow %s.```

salonishah11 (approved):

PR #228, opened 2024-01-22
[WM-2457] Add ReadClient that includes user's token

kpierre13:


service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellClient.java
```diff
@@ -19,6 +19,8 @@ public class CromwellClient {
   private final DependencyUrlLoader dependencyUrlLoader;
 
   private final OkHttpClient singletonHttpClient;
+  private static final String CONNECTION = "Connection";
+  private static final String CLOSE = "close";
```
cjllanwarne, line 23: Super minor, but it feels a little weird to me that we're storing these as apparently unconnected constants... I wonder whether a neater way to DRY without these looking quite so "magic" could be with a function to set the client up, eg:
```
public setClientConnectionToClose(ApiClient apiClient) {
  apiClient.addDefaultHeader("connection", "close");
}
```

... what do you think?

cjllanwarne: Or you could even generalize it to something like `setupApiClient` and throw in the  debug api logging at the same time

service/src/main/java/bio/terra/cbas/dependencies/wes/CromwellService.java
```diff
@@ -30,16 +30,19 @@ public class CromwellService implements HealthCheck {
...
+  private final ApiClient cromwellAuthReadClient;
 
   private final CbasNetworkConfiguration cbasNetworkConfiguration;
 
   public CromwellService(
       CromwellClient cromwellClient,
       ApiClient cromwellWriteClient,
-      CbasNetworkConfiguration cbasNetworkConfiguration) {
+      CbasNetworkConfiguration cbasNetworkConfiguration,
+      ApiClient cromwellAuthReadClient) {
```
cjllanwarne, line 41: Something to think about and watch in the future - I think this means that every connection to CBAS through (any API handler which ever uses Cromwell) will always start by creating two connections to Cromwell, one auth'd and one not.

I think that's fine, and anyway we're sandboxing CBAS pretty well so the total scale is going to be low..... just something to watch if we ever run into trouble

cjllanwarne: Perhaps we could use the `@Lazy` [annotation](https://www.baeldung.com/spring-lazy-annotation) on the bean if this ever becomes a problem

cjllanwarne (approved):

mspector (approved):
👍 interesting discussion, nothing more to add from me

service/src/main/java/bio/terra/cbas/config/BeanConfig.java
```diff
@@ -55,6 +56,7 @@ public cromwell.client.ApiClient cromwellWriteClient(
 
   @Bean("cromwellAuthReadClient")
   @RequestScope
+  @Lazy
```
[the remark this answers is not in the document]
kpierre13: Tested the `@Lazy` annotation in a live environment. Found that the `@RequestScope` annotation has the same effect as the lazy annotation, and so for our purposes, we'll stick with the request scope.

PR #229, opened 2024-01-22
WM-2302: Enable actuator API and micrometer custom metrics

cjllanwarne:
### Description
Note: depends on extra env variables being added in https://github.com/broadinstitute/terra-helmfile/pull/4980 for the custom dimensions.

### Examples
Sample actuator API metrics (MRG logs):

image

Sample custom application metrics (application insights logs:
(note: counting a total of 6 workflow completions, all via the callback API)
image

### DIFF
--- a/service/src/main/java/bio/terra/cbas/common/MicrometerMetrics.java
+++ b/service/src/main/java/bio/terra/cbas/common/MicrometerMetrics.java
@@ -1,0 +1,31 @@
+package bio.terra.cbas.common;
+
+import bio.terra.cbas.models.CbasRunStatus;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import org.springframework.stereotype.Component;
+
+@Component
+public class MicrometerMetrics {
+
+  private final MeterRegistry meterRegistry;
+
+  public MicrometerMetrics(MeterRegistry meterRegistry) {
+    this.meterRegistry = meterRegistry;
+  }
+
+  public void logRunCompletion(String completionTrigger, CbasRunStatus resultsStatus) {
+    Counter counter =
+        Counter.builder("run_completion")
+            .tag("completion_trigger", completionTrigger)
+            .tag("status", resultsStatus.toString())
+            .register(meterRegistry);
+    // This null check only triggers when running tests, where it's hard to mock out the
+    // meterRegistry's internal .counter method without springboot 3's observability frameworks.
+    // See https://github.com/DataBiosphere/terra-workspace-data-service/pull/461 (starting at
+    // QuartzJobTest.scala for an example of what we might want to do when we upgrade)
+    if (counter != null) {
+      counter.increment();
+    }
+  }
+}
--- a/service/src/main/java/bio/terra/cbas/common/MicrometerMetrics.java
+++ b/service/src/main/java/bio/terra/cbas/common/MicrometerMetrics.java
@@ -20,12 +20,6 @@ public class MicrometerMetrics {
             .tag("completion_trigger", completionTrigger)
             .tag("status", resultsStatus.toString())
             .register(meterRegistry);
-    // This null check only triggers when running tests, where it's hard to mock out the
-    // meterRegistry's internal .counter method without springboot 3's observability frameworks.
-    // See https://github.com/DataBiosphere/terra-workspace-data-service/pull/461 (starting at
-    // QuartzJobTest.scala for an example of what we might want to do when we upgrade)
-    if (counter != null) {
-      counter.increment();
-    }
+    counter.increment();
   }
 }
--- a/service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
@@ -2,6 +2,7 @@ package bio.terra.cbas.controllers;
 
 import bio.terra.cbas.api.RunsApi;
 import bio.terra.cbas.common.DateUtils;
+import bio.terra.cbas.common.MicrometerMetrics;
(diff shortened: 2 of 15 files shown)


service/src/main/resources/application.yml
```diff
@@ -55,6 +55,19 @@ cbas:
     # How often to check downstream services for health:
     healthCheckIntervalSeconds: 300
 
+management:
+  endpoints:
+    web:
+      exposure:
+        include: info,health
+      base-path: /actuator
```
cjllanwarne, line 63: Note: these APIs come with actuator but I don't (yet) want to replace our existing health monitoring, so these are being sent to `/actuator` paths, for now

mspector: I don't really understand this comment but I'm curious to know more... I'll ask you about it later today

mspector:
Oops, I approved before noticing the unit test failures. But the errors look like some kind of dependencies issue, and/or a liquibase issue, so I'm still approving the core change. Let me know if you want a re-review.

kpierre13 (approved):

service/src/main/java/bio/terra/cbas/controllers/RunsApiController.java
```diff
@@ -85,6 +90,13 @@ public ResponseEntity<Void> postRunResults(RunResultsRequest body) {
     CbasRunStatus resultsStatus = CbasRunStatus.fromCromwellStatus(body.getState().toString());
     List<String> failures = body.getFailures();
 
+    Counter counter =
```
jladieu, line 93: This looks similar to how we did a custom metric, but I want to point out that Spring Boot 3 brings a much more interesting and expressive `Observation` interface, which I think is think will be valuable for getting tracing as well as metrics gathering.  We haven't used it much yet but there's a near-landing PR where we start trying it out under way here: https://github.com/DataBiosphere/terra-workspace-data-service/pull/461/files#r1460108734

jladieu (approved):

service/src/test/java/bio/terra/cbas/controllers/TestRunsApiController.java
```diff
@@ -76,6 +77,8 @@ class TestRunsApiController {
   @Autowired private ObjectMapper objectMapper;
   @MockBean private SamService samService;
 
+  @MockBean private MeterRegistry meterRegistry;
```
jladieu: My hunch is that if you use `@Autowired` here instead, the framework may provide an instance of `SimpleMeterRegistry` which shouldn't return `null` from the `Counter.builder` chain.

If this doesn't work out of the box, then I think it still _should_ be possible to coerce the framework to inject a `SimpleMeterRegistry` here, and this would let you sidestep the null check on counter.

service/src/test/java/bio/terra/cbas/controllers/TestRunsApiController.java
```diff
@@ -77,7 +77,7 @@ class TestRunsApiController {
   @Autowired private ObjectMapper objectMapper;
   @MockBean private SamService samService;
 
-  @MockBean private MeterRegistry meterRegistry;
+  @MockBean private MicrometerMetrics micrometerMetrics;
```
jladieu, line 80: Similarly here to my comment on the prior commit, I think if you use `@Autowired`, the framework should inject a `MicrometerMetrics` instance, which itself will be autowired (to include a `SimpleMetricRegistry`).  Then you wouldn't need to mock out this bean either (although I guess you can, if that's your intent).

service/src/main/java/bio/terra/cbas/common/MicrometerMetrics.java
```diff
@@ -0,0 +1,31 @@
...
+    Counter counter =
+        Counter.builder("run_completion")
+            .tag("completion_trigger", completionTrigger)
+            .tag("status", resultsStatus.toString())
+            .register(meterRegistry);
+    // This null check only triggers when running tests, where it's hard to mock out the
+    // meterRegistry's internal .counter method without springboot 3's observability frameworks.
+    // See https://github.com/DataBiosphere/terra-workspace-data-service/pull/461 (starting at
+    // QuartzJobTest.scala for an example of what we might want to do when we upgrade)
+    if (counter != null) {
```
jladieu, line 27: This null check shouldn't be needed if `MeterRegistry` is autowired by the framework and never passed in as a MockBean.

service/src/test/java/bio/terra/cbas/runsets/results/TestRunCompletionHandlerFunctional.java
```diff
@@ -46,6 +47,8 @@
 @ContextConfiguration(classes = RunCompletionHandler.class)
 class TestRunCompletionHandlerFunctional {
 
+  private MicrometerMetrics micrometerMetrics = mock(MicrometerMetrics.class);
```
jladieu, line 50: `@Autowired` here too...

service/src/main/resources/application.yml
```diff
@@ -55,6 +55,19 @@ cbas:
     # How often to check downstream services for health:
     healthCheckIntervalSeconds: 300
 
+management:
```
okotsopoulos, line 58: Hi @cjllanwarne -- I'd recommend expanding this configuration slightly to also get percentile histogram metrics for endpoint latencies.  More info here: 

https://github.com/DataBiosphere/terra-workspace-manager/pull/1607

cjllanwarne: Cool!!

cjllanwarne: This seems to be prometheus specific? It doesn't look like these show up in the default `AppRequests` section of the azure console?

cjllanwarne:
Merging for now but super interested in finding out more about how to add histograms for those response times in a follow up!

PR #233, opened 2024-01-26
[WM-2445] Add github_method_details table and branch_or_tag column

kpierre13:
image


service/src/main/resources/changelog/changesets/20240126_github_method_details_batch_or_tag.yaml
```diff
@@ -0,0 +1,41 @@
...
+                  constraints:
+                    nullable: false
+              - column:
+                  name: organization
+                  type: varchar(100)
+                  constraints:
+                    nullable: true
+              - column:
+                  name: path
+                  type: varchar(1000)
```
salonishah11, line 21: TOL: Should this and above `varchar` also be `text`? Is there a benefit of `varchar` over `text`?

cjllanwarne:
I thought we were only going to store the `branch_name` column for now, and leave "branch or tag" to a later workstream around method versioning/saving configurations

kpierre13:
> I thought we were only going to store the `branch_name` column for now, and leave "branch or tag" to a later workstream around method versioning/saving configurations

Are these two different columns? I named it branch or tag to more accurately represent what it'll be used for, since when we backfill this data, there's no telling whether or not users have been using branches or tags.

cjllanwarne:
>Are these two different columns? I named it branch or tag to more accurately represent what it'll be used for, since when we backfill this data, there's no telling whether or not users have been using branches or tags.

Maybe `branch_or_tag_name`? To me `branch_or_tag` would be a boolean or an enum with two possible values: `branch` and `tag`

kpierre13:
> Maybe `branch_or_tag_name`? To me `branch_or_tag` would be a boolean or an enum with two possible values: `branch` and `tag`

Currently in this PR, the column name is `branch_or_tag_name`. Is there another place I missed that should be renamed?

cjllanwarne:
Oh, no, that sounds fine then... I think I was just going off the ticket/pr title

service/src/main/resources/changelog/changesets/20231102_set_table_role.yaml
```diff
@@ -20,3 +20,4 @@ databaseChangeLog:
           ALTER TABLE method_version OWNER TO ${dbRole};
           ALTER TABLE run OWNER TO ${dbRole};
           ALTER TABLE run_set OWNER TO ${dbRole};
+          ALTER TABLE github_method_details OWNER TO ${dbRole};
```
cjllanwarne, line 23: Thanks for remembering this!

service/src/main/resources/changelog/changelog.yaml
```diff
@@ -89,6 +89,9 @@ databaseChangeLog:
   - include:
       file: changesets/20231102_set_table_role.yaml
       relativeToChangelogFile: true
```
cjllanwarne, line 91: The set_table_role changeset is the actual changeset that needs to be at the end. It looks like _someone_ (oops!) forgot to honor this the last time the file was updated. So the order should go:
- add_workspace_id
- github_method_details
- set_table_role

And I wonder whether we can make the reminder more easy to notice somehow 🤔

cjllanwarne (changes requested):
Requesting changes but they're not your fault

service/src/main/resources/changelog/changelog.yaml
kpierre13: I added the name of the change set in the reminder to maybe make it easier to notice?

cjllanwarne (approved):

salonishah11 (approved):

PR #236, opened 2024-02-02
WM-2240: Example of using instance initializer pattern to run startup jobs

cjllanwarne:
As proof of the initializer running, this log message shows up in application insights:
image


cjllanwarne:
Subsequently superseded

PR #237, opened 2024-02-02
WM-2488: Record original workspace IDs

cjllanwarne:
Note: depends on https://github.com/broadinstitute/terra-helmfile/pull/5021

### DIFF
--- a/service/src/main/java/bio/terra/cbas/dao/MethodVersionDao.java
+++ b/service/src/main/java/bio/terra/cbas/dao/MethodVersionDao.java
@@ -23,8 +23,8 @@ public class MethodVersionDao {
 
   public int createMethodVersion(MethodVersion methodVersion) {
     return jdbcTemplate.update(
-        "insert into method_version (method_version_id, method_id, method_version_name, method_version_description, method_version_created, method_version_last_run_set_id, method_version_url) "
-            + "values (:methodVersionId, :methodId, :name, :description, :created, :lastRunSetId, :url)",
+        "insert into method_version (method_version_id, method_id, method_version_name, method_version_description, method_version_created, method_version_last_run_set_id, method_version_url, method_version_original_workspace_id) "
+            + "values (:methodVersionId, :methodId, :name, :description, :created, :lastRunSetId, :url, :originalWorkspaceId)",
         new BeanPropertySqlParameterSource(methodVersion));
   }
 
--- a/service/src/main/java/bio/terra/cbas/dao/RunDao.java
+++ b/service/src/main/java/bio/terra/cbas/dao/RunDao.java
@@ -38,8 +38,8 @@ public class RunDao {
 
   public int createRun(Run run) {
     return jdbcTemplate.update(
-        "insert into run (run_id, engine_id, run_set_id, record_id, submission_timestamp, status, last_modified_timestamp, last_polled_timestamp, error_messages, run_original_workspace_id)"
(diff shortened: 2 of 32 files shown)


salonishah11:
@cjllanwarne Since rows in `Run` table would always be tied to a specific run set in `Run Set` table, what would be the benefit of adding `workspace_id` to `Run` table too?

mspector:
+1 to @salonishah11 's question. The only benefit I can think of is that having the original workspace ID included in the runs table means we can use simpler queries when we're trimming the database.

cjllanwarne:
Great point, I don't think we need this for `run`s, let's get rid of it.

service/src/main/java/bio/terra/cbas/dao/RunDao.java
```diff
@@ -128,14 +128,6 @@ public int updateLastPolledTimestamp(UUID runID) {
                 DateUtils.currentTimeInUTC())));
   }
 
-  public int updateErrorMessage(UUID runId, String updatedErrorMessage) {
```
salonishah11, line 131: Was this method not used?

service/src/main/java/bio/terra/cbas/dao/MethodVersionDao.java
```diff
@@ -23,8 +23,8 @@ public MethodVersionDao(NamedParameterJdbcTemplate jdbcTemplate) {
 
   public int createMethodVersion(MethodVersion methodVersion) {
     return jdbcTemplate.update(
-        "insert into method_version (method_version_id, method_id, method_version_name, method_version_description, method_version_created, method_version_last_run_set_id, method_version_url) "
-            + "values (:methodVersionId, :methodId, :name, :description, :created, :lastRunSetId, :url)",
+        "insert into method_version (method_version_id, method_id, method_version_name, method_version_description, method_version_created, method_version_last_run_set_id, method_version_url, method_version_original_workspace_id) "
```
salonishah11, line 26: small nit: Here and in other tables, should this be simply called `workspace_id` or `<table_name>_workspace_id`? Is there a reason to add `original` in the name?

salonishah11 (approved):

service/src/main/java/bio/terra/cbas/dao/RunDao.java
cjllanwarne: Yeah, this is just "while I'm there" tidying up

service/src/main/java/bio/terra/cbas/dao/MethodVersionDao.java
salonishah11: Won't the workspace IDs in this column always belong to the "original" workspace (irrespective of it was from source workspace or current workspace)? I guess maybe I am not understanding the nuance here 🤔

salonishah11: So we won't be storing the workspace id for the current workspace when new methods are imported or when new run sets are launched (in the cloned workspace)?

cjllanwarne: We should get both - when we import a new method or start a new run set, this value gets set to the current workspace. Then when we clone it, it stays set at the original workspace rather than being updated to the new one.

salonishah11: I see. Thank you for the clarification! 👍

PR #239, opened 2024-02-06
WM-2487: Treat failure to update WDS as retryable

cjllanwarne:
Tested by:
- Created a "sabotaged" branch which threw API exceptions 60% of the time
- Submitted a set of workflows
- Watched logs for evidence of the error being thrown in CBAS

image
image

- Watched logs for evidence of Cromwell retrying the callback
image

- Observed that all workflows eventually succeeded and data tables were updated
- Observed that the completion trigger was indeed still callback
image

(Note: there was actually both a smartpoller and callback entry, indicating that the callback happened _as well as_ a successful update on smartpolling. That's a good thing - it means the fallback mechanism of letting the smartpoller pick up if the callback gives up is _also_ working well)

### DIFF
--- a/service/src/main/java/bio/terra/cbas/runsets/results/RunCompletionHandler.java
+++ b/service/src/main/java/bio/terra/cbas/runsets/results/RunCompletionHandler.java
@@ -76,10 +76,8 @@ public class RunCompletionHandler {
     long updateResultsStartNanos = System.nanoTime();
     RunCompletionResult updateResult = RunCompletionResult.ERROR;
     try {
-      var updatedRunState = status;
-
-      if (updatableRun.status() == updatedRunState
-          && updatedRunState != CbasRunStatus.COMPLETE
+      if (updatableRun.status() == status
+          && status != CbasRunStatus.COMPLETE
           && (workflowErrors == null || workflowErrors.isEmpty())) {
         // Status is already up-to-date, not complete (no outputs to process), no errors to save.
         return updateDatabaseRunStatusOnly(updatableRun);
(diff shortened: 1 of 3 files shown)


kpierre13 (approved):

service/src/main/java/bio/terra/cbas/runsets/results/RunCompletionHandler.java
```diff
@@ -112,21 +110,18 @@ public RunCompletionResult updateResults(
         if (recordAttributes != null && !recordAttributes.isEmpty()) {
           String errorMessage = saveOutputsToWDS(updatableRun, recordAttributes);
           if (errorMessage != null && !errorMessage.isEmpty()) {
-            errors.add(errorMessage);
+            // Failure to update WDS. Let this be retried by Cromwell
+            logger.error(errorMessage);
+            return RunCompletionResult.ERROR;
```
salonishah11, line 115: TOL: Should we set an attempt limit on this retry? So after N retries for this error CBAS eventually records an error? Or is that not worth it since in second/third attempt it output saving to WDS should succeed?

cjllanwarne: Cromwell will give up after 10 attempts. After that the user will trigger a smartpoller retry every refresh. 

So, as long as WDS is unavailable, the method will stay in Running but if WDS ever does come back up the workflow will move to Done the next time the user refreshes the page.

Of course, WDS _should_ come back within a couple of retries so I wouldn't expect us to ever actually reach that final state, but there is at least some resolution there.

salonishah11: Good to know. Thank you for the clarification!

salonishah11 (approved):

PR #242, opened 2024-02-13
[WM-2460] Update Method Dao

kpierre13:


service/src/main/java/bio/terra/cbas/common/MethodUtil.java
```diff
@@ -56,4 +56,8 @@ public static MethodSourceEnum convertToMethodSourceEnum(String methodSource)
 
     throw new UnknownMethodSourceException(methodSource);
   }
+
+  public static Boolean verifyIfPrivate(String methodUrl) {
```
kpierre13, line 60: Updated in Michael's [PR](https://github.com/DataBiosphere/cbas/pull/243).

common/openapi.yml
```diff
@@ -66,6 +66,13 @@ paths:
             When specified, filters to only include appropriate method and method version. If
             included, forces "show_versions" to be true.
           required: false
+        - in: query
+          name: method_details
+          schema:
+            type: boolean
+          description: >
+            When specified, toggles on or off the details for each method.
+          required: false
```
cjllanwarne, line 75: Can we document what the default is?

common/openapi.yml
```diff
@@ -653,6 +660,8 @@ components:
           format: date-time
         last_run:
           $ref: '#/components/schemas/MethodLastRunDetails'
+        source_details:
+          $ref: '#/components/schemas/GithubMethodSourceDetailsResponse'
```
cjllanwarne, line 664: Why `source_details` and not `method_details`?

cjllanwarne: What if the method is not a github method?

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -177,20 +185,36 @@ public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMetho
       UUID methodId = UUID.randomUUID();
       UUID runSetId = UUID.randomUUID();
 
+      GithubUrlDetailsManager.GithubUrlComponents githubUrlComponents =
+          githubUrlDetailsManager.extractDetailsFromUrl(rawMethodUrl);
```
cjllanwarne, line 189: What happens if the URL is not a github url?

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -267,7 +298,16 @@ private void createNewMethod(
...
-            cbasContextConfig.getWorkspaceId());
+            cbasContextConfig.getWorkspaceId(),
+            branchOrTagName);
+
+    GithubMethodSource sourceDetails =
+        new GithubMethodSource(
+            githubMethodSourceDetails.repository(),
+            githubMethodSourceDetails.organization(),
+            githubMethodSourceDetails.path(),
+            githubMethodSourceDetails._private(),
```
cjllanwarne, line 309: can we call this `isPrivate`?

service/src/main/java/bio/terra/cbas/util/methods/GithubUrlDetailsManager.java
```diff
@@ -0,0 +1,84 @@
...
+
+    githubUrlComponents.setOrganization(parts[1]);
+    githubUrlComponents.setRepository(parts[2]);
+    githubUrlComponents.setBranchOrTag(parts[3]);
+    githubUrlComponents.setPath(String.join("/", gitHubPathParts));
+
+    return githubUrlComponents;
+  }
+
+  public static class GithubUrlComponents {
```
cjllanwarne, line 26: TOL could this be a `record` instead of a `static class`?

service/src/main/java/bio/terra/cbas/util/methods/GithubUrlDetailsManager.java
```diff
@@ -0,0 +1,84 @@
+package bio.terra.cbas.util.methods;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+import org.springframework.stereotype.Component;
+
+@Component
```
cjllanwarne, line 8: Why does this need to be a component? It looks more like a utility function class?

service/src/test/java/bio/terra/cbas/util/methods/TestGithubUrlDetailsManager.java
```diff
@@ -0,0 +1,40 @@
+package bio.terra.cbas.util.methods;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.net.URISyntaxException;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.test.context.ContextConfiguration;
+
+@WebMvcTest
```
cjllanwarne, line 10: I don't think this is an MVC test. It looks like standard unit tests?

service/src/test/java/bio/terra/cbas/util/methods/TestGithubUrlDetailsManager.java
```diff
@@ -0,0 +1,40 @@
...
+@ContextConfiguration(classes = {GithubUrlDetailsManager.class})
+class TestGithubUrlDetailsManager {
+
+  @Test
+  void returnsExpectedUrlComponents() throws URISyntaxException {
```
cjllanwarne, line 15: This would be a great time to use table based testing. You could define a big table of input and expected org / repo / branch / etc names, and then validate them all with the same test method. And you could also list a bunch of failure cases and check we handle those nicely too

cjllanwarne (commented):
Looks mostly good, with a couple of smaller comments.

Since it will be hard to change later, just a few quick questions to consider about the API changes. Happy to go either way as long as we have thought about them - 
1. Do we want to use a polymorphic type here (like we do elsewhere in the spec for eg type definitions) so that the details could be a github OR a raw url OR a something else in the future?
2. Do we want to include details in the method API at all? Could it be a separate method details APi?
3. Do we have plans to use the other github details elsewhere in CBAS and/or the UI? Will that be wired in separately?
4. Is there any case where we would supply `false` to the option not to return details?

Should we be using a transaction when we add the method and method details? What happens if CBAS gets restarted half way through writing one of the tables but not the other?

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
mspector: I wonder if this should just be a `UrlDetailsManager`, even if its purpose for now is to extract GitHub details. It could also handle the question "is this a GitHub URL?"

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -216,14 +240,19 @@ public ResponseEntity<MethodListResponse> getMethods(
...
       methodDetails =
-          List.of(methodVersionToMethodDetails(methodVersionDao.getMethodVersion(methodVersionId)));
+          List.of(
+              methodVersionToMethodDetails(
+                  methodVersionDao.getMethodVersion(methodVersionId), showMethodDetails));
     } else {
       List<Method> methods =
           methodId == null ? methodDao.getMethods() : List.of(methodDao.getMethod(methodId));
       boolean nullSafeShowVersions = showVersions == null || showVersions;
+      boolean nullSafeShowMethodDetails = showMethodDetails == null || showMethodDetails;
```
mspector, line 250: this is more of a "note to self, look into this" than feedback or a change request, because I see it's a pattern that you didn't introduce -- but it's not obvious how these variables might end up being `null`. Does it happen because it gets its arguments based on the openapi spec? I wonder...

service/src/main/java/bio/terra/cbas/common/MethodUtil.java
```diff
@@ -56,4 +56,8 @@ public static MethodSourceEnum convertToMethodSourceEnum(String methodSource)
...
+    return true;
```
salonishah11, line 61: Just noting it here - as discussed in standup this should be set to false as default.

salonishah11:
@kpierre13 can you paste a sample response for GET methods endpoint containing the new method_details information?

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -479,19 +537,37 @@ private static MethodVersionDetails methodVersionToMethodVersionDetails(
...
+        .branchOrTagName(methodVersion.branchOrTagName());
   }
 
-  private MethodDetails methodVersionToMethodDetails(MethodVersion methodVersion) {
+  private MethodDetails methodVersionToMethodDetails(
+      MethodVersion methodVersion, Boolean includeSourceDetails) {
     Method method = methodVersion.method();
+    GithubMethodSourceDetailsResponse sourceDetails;
+
+    if (includeSourceDetails != null) {
```
salonishah11, line 549: Since `includeSourceDetails` should this essentially be this similar to above?
```
if (includeSourceDetails) ..
```

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -479,19 +537,37 @@ private static MethodVersionDetails methodVersionToMethodVersionDetails(
...
+      GithubMethodSource githubMethodSource = methodDao.getMethodSourceDetails(method.methodId());
+      sourceDetails =
+          new GithubMethodSourceDetailsResponse()
+              .methodId(githubMethodSource.methodId())
+              .path(githubMethodSource.path())
+              ._private(githubMethodSource._private())
+              .organization(githubMethodSource.organization())
+              .repository(githubMethodSource.repository());
```
salonishah11, line 557: Clarifying Java question - this doesn't throw exception when values for these are `null` in database (which it would be until we backfill these values)?

service/src/main/java/bio/terra/cbas/util/methods/GithubUrlDetailsManager.java
salonishah11: +1. I think `extractDetailsFromUrl` can just be a utility function inside a `GithubUtils` class or existing [MethodUtils class](https://github.com/DataBiosphere/cbas/blob/main/service/src/main/java/bio/terra/cbas/common/MethodUtil.java).

service/src/main/java/bio/terra/cbas/util/methods/GithubUrlDetailsManager.java
```diff
@@ -0,0 +1,84 @@
...
+public class GithubUrlDetailsManager {
+
+  public GithubUrlComponents extractDetailsFromUrl(String url) throws URISyntaxException {
```
salonishah11, line 11: small nit: since this is specific to Github maybe can be renamed `extractGithubDetailsFromUrl`?

common/openapi.yml
kpierre13: I named this is as `source_details` keeping in mind that we may have other sources. However, what is returned here will vary, so we'll want to keep that in mind when we support other sources.

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
kpierre13: Since we only support github urls, I didn't find it necessary to name this something more broad. In the future, other urls will most likely follow a different format, so we'll need an additional method in the manager to handle these specific types.

service/src/test/java/bio/terra/cbas/util/methods/TestGithubUrlDetailsManager.java
kpierre13: Good 'ol copy-and-paste!

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
kpierre13: Currently `organization` is the only nullable column in the database. We haven't put any new methods into the db, so it hasn't thrown an exception. It would however if we tried to backfill these values and didn't extract them from the url.

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
kpierre13: Looking more closely it seems like I get different behavior when I set the variable to `boolean` or to `Boolean`. `boolean` throws ```Operator '!=' cannot be applied to 'boolean', 'null'``` while `Boolean` does not. I think I'll change both instances to `Boolean` to have that extra check.

service/src/main/java/bio/terra/cbas/util/methods/GithubUrlDetailsManager.java
```diff
@@ -0,0 +1,6 @@
+package bio.terra.cbas.util.methods;
+
+public class GithubUrlDetailsManager {
```
mspector, line 3: Correct me if I'm wrong, but I don't think we need this GithubUrlDetailsManager class anymore, assuming we can use GithubUrlComponents directly

common/openapi.yml
```diff
@@ -691,6 +695,20 @@ components:
           format: uuid
         method_version_name:
           type: string
+    GithubMethodSourceDetailsResponse:
```
mspector, line 698: do we still want this response here after all, given that we now have `isPrivate` in the top level of the response? (I don't think this is being used anymore after the most recent changes, but correct me if I'm wrong)

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -177,20 +181,36 @@ public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMetho
...
 
+      GithubUrlComponents githubUrlComponents =
+          MethodUtil.extractGithubDetailsFromUrl(rawMethodUrl);
+
+      String path = githubUrlComponents.path();
+      String repository = githubUrlComponents.repo();
+      String organization = githubUrlComponents.org();
+      String branchOrTagName = githubUrlComponents.branchOrTag();
+
+      GithubMethodSource githubMethodSource =
```
mspector, line 192: Similar to another comment I made -- it feels to me like `GithubMethodDetails` is more accurate.

I saw you made a comment elsewhere arguing that using the word "Source" prepares us for adding sources beyond GitHub, but I think we could get that clarity with something like `DockerhubMethodDetails` or `SomeothersourceMethodDetails`

service/src/main/java/bio/terra/cbas/models/GithubMethodSource.java
```diff
@@ -0,0 +1,14 @@
+package bio.terra.cbas.models;
+
+import java.util.UUID;
+
+public record GithubMethodSource(
```
mspector, line 5: Sort of a nit (but I think naming is important so we should discuss if you disagree): this feels like it should be called `GithubMethodDetails`. 

I saw your argument in another comment that using the word "Source" sets us up to support non-Github methods, but I think it'll be just as clear to name some future method details record like `DockerhubMethodDetails`, `SomeothersourceMethodDetails`

mspector (changes requested):
I think we should nail down naming conventions, and one of Chris's comments about _private vs. isPrivate should be addressed, but otherwise this is looking good!

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -177,20 +181,36 @@ public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMetho
...
+      GithubUrlComponents githubUrlComponents =
+          MethodUtil.extractGithubDetailsFromUrl(rawMethodUrl);
+
+      String path = githubUrlComponents.path();
+      String repository = githubUrlComponents.repo();
+      String organization = githubUrlComponents.org();
+      String branchOrTagName = githubUrlComponents.branchOrTag();
+
+      GithubMethodDetails githubMethodDetails =
+          new GithubMethodDetails(repository, organization, path, isPrivate, methodId);
```
salonishah11, line 193: Based on the discussion we had, we only want to store these details when method source is Github since the purpose of `github_details` table is to store information for Github methods. When the need arises for storing more details related to dockstore method, we can create a new table for it.

salonishah11 (changes requested):

service/src/main/java/bio/terra/cbas/dao/mappers/GithubMethodSourceDetailsMapper.java
```diff
@@ -0,0 +1,20 @@
+package bio.terra.cbas.dao.mappers;
+
+import bio.terra.cbas.models.GithubMethodDetails;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.UUID;
+import org.springframework.jdbc.core.RowMapper;
+
+public class GithubMethodSourceDetailsMapper implements RowMapper<GithubMethodDetails> {
```
salonishah11, line 9: Looking at the convention and to keep it consistent, can this also be renamed to ` GithubMethodDetailsMapper`?

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
```diff
@@ -112,4 +114,26 @@ public int countMethods(String methodName, String methodVersion) {
...
+  public GithubMethodDetails getMethodSourceDetails(UUID methodId) {
+    String sql =
+        "SELECT * FROM github_method_details WHERE github_method_details.method_id = :methodId";
+
+    MapSqlParameterSource params = new MapSqlParameterSource(Map.of("methodId", methodId));
+
+    return jdbcTemplate.queryForObject(sql, params, new GithubMethodSourceDetailsMapper());
+  }
+
+  public int deleteMethodSourceDetails(UUID methodId) {
```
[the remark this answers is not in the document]
kpierre13: I was going back an forth about that myself... I think maybe yes? Since we may have many new sources in the future, it may be nice to have a separate Dao associated with each new source.

mspector (approved):

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -236,7 +268,9 @@ private void createNewMethod(
       PostMethodRequest postMethodRequest,
       WorkflowDescription workflowDescription,
       List<MethodInputMapping> methodInputMappings,
-      List<MethodOutputMapping> methodOutputMappings)
+      List<MethodOutputMapping> methodOutputMappings,
+      GithubMethodDetails githubMethodSourceDetails,
```
salonishah11, line 272: small nit: can this variable also be renamed to `githubMethodDetails` to keep it consistent?

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -293,6 +328,21 @@ private void createNewMethod(
...
+    GithubMethodDetails sourceDetails;
+
+    if (githubMethodSourceDetails != null) {
+      sourceDetails =
+          new GithubMethodDetails(
+              githubMethodSourceDetails.repository(),
+              githubMethodSourceDetails.organization(),
+              githubMethodSourceDetails.path(),
+              githubMethodSourceDetails.isPrivate(),
+              githubMethodSourceDetails.methodId());
```
salonishah11, line 341: Maybe I am missing something, but isn't `sourceDetails` same as the parameter `githubMethodSourceDetails` passed into the function? Can `githubMethodSourceDetails` directly be used in L343?

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -453,6 +503,9 @@ private void addLastRunDetails(List<MethodDetails> methodDetails) {
 
   private MethodDetails methodToMethodDetails(Method method, boolean includeVersions) {
 
+    Boolean isPrivate =
+        githubMethodDetailsDao.getMethodSourceDetails(method.methodId()).isPrivate();
```
salonishah11, line 507: Will `.isPrivate()` throw exception if `githubMethodDetailsDao.getMethodSourceDetails(...)` returns null (for dockstore methods)?

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -479,19 +533,24 @@ private static MethodVersionDetails methodVersionToMethodVersionDetails(
...
-        .url(methodVersion.url());
+        .url(methodVersion.url())
+        .branchOrTagName(methodVersion.branchOrTagName());
   }
 
   private MethodDetails methodVersionToMethodDetails(MethodVersion methodVersion) {
     Method method = methodVersion.method();
+    Boolean isMethodPrivate =
+        githubMethodDetailsDao.getMethodSourceDetails(method.methodId()).isPrivate();
+
```
salonishah11, line 544: Similar question regarding null exception being thrown here as well

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
kpierre13: This would produce an exception. To handle this, I added an `if` conditional to first check if there are any method source details for the method. If this does not return null, then we set `isPrivate` to `githubMethodDetailsDao.getMethodSourceDetails(...).isPrivate()`. If the details are null, then we set `isPrivate` to null in the method details. This way, we avoid accessing `isPrivate` for dockstore methods which would throw an exception. Thank you for catching!

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -453,6 +494,16 @@ private void addLastRunDetails(List<MethodDetails> methodDetails) {
 
   private MethodDetails methodToMethodDetails(Method method, boolean includeVersions) {
 
+    Boolean isMethodPrivate = false;
+
+    if (Objects.equals(
+        method.methodSource(), PostMethodRequest.MethodSourceEnum.GITHUB.toString())) {
+      if (githubMethodDetailsDao.getMethodSourceDetails(method.methodId()) != null) {
+        isMethodPrivate =
+            githubMethodDetailsDao.getMethodSourceDetails(method.methodId()).isPrivate();
```
salonishah11, line 503: nit: if GitHub method has data in `github_method_details` table, here it would end up running the same query in   `getMethodSourceDetails()` twice. To avoid this you could store that value in a variable and use it on L503 instead of re-running the query.

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -479,19 +531,31 @@ private static MethodVersionDetails methodVersionToMethodVersionDetails(
...
   private MethodDetails methodVersionToMethodDetails(MethodVersion methodVersion) {
     Method method = methodVersion.method();
+    Boolean isMethodPrivate = false;
+
+    if (Objects.equals(
+        method.methodSource(), PostMethodRequest.MethodSourceEnum.GITHUB.toString())) {
+      if (githubMethodDetailsDao.getMethodSourceDetails(method.methodId()) != null) {
+        isMethodPrivate =
+            githubMethodDetailsDao.getMethodSourceDetails(method.methodId()).isPrivate();
+      }
```
salonishah11, line 547: same nit here too

salonishah11 (approved):

PR #243, opened 2024-02-13
[WM-2453] GitHubService

mspector:


service/src/main/java/bio/terra/cbas/dependencies/github/GitHubService.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.gson.Gson;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.springframework.stereotype.Component;
+
+@Component
+public class GitHubService {
```
mspector, line 12: other dependencies have separate `*Service` and `*Client` classes - why is that? is it relevant here?

service/src/main/java/bio/terra/cbas/dependencies/github/GitHubService.java
```diff
@@ -0,0 +1,61 @@
+package bio.terra.cbas.dependencies.github;
```
mspector, line 1: there isn't an official GitHub API client, but there are some third party ones that I chose not to use. Should I favor those third party libraries over writing something like this class?

service/src/main/java/bio/terra/cbas/dependencies/github/GitHubService.java
```diff
@@ -0,0 +1,61 @@
...
+    WebTarget target = client.target(BASE_URL).path("/repos").path(organization).path(repo);
+    Response response =
+        target
+            .request(MediaType.APPLICATION_JSON_TYPE)
+            .header("Authorization", "token " + token)
+            .get();
+    if (response.getStatus() == 200) {
+      return gson.fromJson(response.readEntity(String.class), GetRepoResponse.class);
+    } else {
+      throw new RuntimeException("Failed to get repo: " + response.getStatusInfo());
```
kpierre13, line 35: Are there different exceptions here? For example, an exception for a bad token vs. a non-existent repo.

mspector: good call - just checked, and the API does behave differently for those examples. I'll make some changes

service/src/main/resources/application.yml
```diff
@@ -99,6 +99,9 @@ leonardo:
 dockstore:
   baseUri: "https://staging.dockstore.org/"
 
+github:
+  personalAccessToken: "<insert personal access token here>"
```
mspector, line 103: this should be replaced by a call to ECM eventually

kpierre13, line 103: Are we saying that we can use this dummy `personalAccessToken` because even with a non-functioning token is attached to the Bearer, we're still only importing public workflows which don't need a real token to check its status?

service/src/main/java/bio/terra/cbas/dependencies/github/GitHubClient.java
```diff
@@ -0,0 +1,71 @@
...
+    WebTarget target = client.target(BASE_URL).path("/repos").path(organization).path(repo);
+    Response response =
+        target
+            .request(MediaType.APPLICATION_JSON_TYPE)
+            .header("Authorization", "Bearer: " + token)
+            .get();
+    if (response.getStatus() == 200) {
+      return gson.fromJson(response.readEntity(String.class), RepoInfo.class);
+    } else {
+      RepoError error = gson.fromJson(response.readEntity(String.class), RepoError.class);
```
kpierre13, line 35: Were you able force different kinds of error messages here? i.e. is "a user has a token, but no access to the repo" different from some other sort of error message we may get, i.e. a 500?

service/src/main/resources/application.yml
mspector: I think you're right, but that wasn't exactly my intent -- I put this here as a way to test while we wait to wire up with ECM. Once we do that, it could potentially stay as a "local only" config, or we could remove it altogether.

service/src/main/java/bio/terra/cbas/dependencies/github/GitHubClient.java
mspector: If the user hits this GitHub endpoint for a private repo with either (1) no token or (2) a token that lacks authorization for the repo, it responds with the same "Not Found" message, so we won't be able to distinguish your two situations from GitHub's response alone. 

If we need to be able to distinguish these, in my opinion that check should be in a level of abstraction above GitHubClient (like GitHubService or the API controller). I think it'll depend a bit on how we choose to wire in ECM, so my vote would be to wait until that happens before we deal with this, but I'm not adamant about it. If we want to get that check into this PR, we should coordinate a little (or catch me up) on where/when the ECM token will be retrieved.

kpierre13: I don't think we need to distinguish the two immediately. I think a general error message touching on both possibilities is enough. I was just curious if there was a difference in error message that you saw.

kpierre13 (approved):

PR #245, opened 2024-02-14
[WM-2460] make tests pass

mspector:


service/src/main/java/bio/terra/cbas/util/methods/GithubUrlDetailsManager.java
```diff
@@ -5,7 +5,7 @@
 @Component
 public class GithubUrlDetailsManager {
 
-  public GithubUrlComponents extractDetailsFromUrl(String url) {
+  public static GithubUrlComponents extractDetailsFromUrl(String url) {
```
mspector, line 8: I made this static because it doesn't actually use any class properties - let's talk if you think this is not a good idea for some reason

PR #254, opened 2024-03-07
[WM-2499] Implement ECM client

kpierre13:


service/src/main/java/bio/terra/cbas/dependencies/ecm/EcmClient.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Component
+public class EcmClient {
+  private final EcmServerConfiguration ecmServerConfiguration;
+
+  public EcmClient(EcmServerConfiguration ecmServerConfiguration) {
+    this.ecmServerConfiguration = ecmServerConfiguration;
+  }
+
+  public ApiClient apiClient() {
```
cjllanwarne, line 17: Would we ever want an unauthorized ecm?

cjllanwarne (approved):
Change looks good, would be good to see how it ends up being used after the next ticket, and whether we can remove any unused functions if (eg) the unauthorized ecm client doesn't end up being needed

service/src/main/java/bio/terra/cbas/dependencies/ecm/EcmClient.java
kpierre13: Yeah, we're currently using it for the scheduled health check

cjllanwarne: Of course!

service/build.gradle
```diff
@@ -34,6 +34,7 @@ dependencies {
     implementation 'bio.terra:datarepo-client:2.13.0-SNAPSHOT'
     implementation 'org.postgresql:postgresql:42.7.2'
     implementation 'jakarta.ws.rs:jakarta.ws.rs-api'
+    implementation 'bio.terra:externalcreds-client-resttemplate:1.12.0-20240229.204210-1'
```
salonishah11, line 37: nit: since you have a method calling `oauthApi.getProviderAccessToken` in EcmService (even though it might not be used until later PR), it might be good to update the client to 1.15 to avoid the `500 error no static resource /api/oidc..` that we were seeing earlier.

salonishah11 (approved):

PR #255, opened 2024-03-08
[WM-2489] backfill original workspace

mspector:


service/src/test/java/bio/terra/cbas/util/methods/TestBackfillOriginalWorkspaceIds.java
```diff
@@ -0,0 +1,280 @@
...
+    verify(methodDao, never())
+        .updateOriginalWorkspaceId(eq(methodClonedBackfilled.methodId()), any());
+    verify(methodDao, never())
+        .updateOriginalWorkspaceId(eq(methodCreatedBackfilled.methodId()), any());
+  }
+
+  @Test
+  void testBackfillRunSets() {
+    BackfillOriginalWorkspaceIds.backfillRunSets(runSetDao, cbasContextConfig, logger);
+    // verify(runSetDao, times(2)).getRunSets(null, false);
```
mspector, line 82: This `verify` call is showing inconsistent behavior for reasons that I don't understand.

When this test is run by itself (in IntelliJ), runSetDao.getRunSets is called twice. My working theory is that `when(runSetDao.getRunSets)...` counts as one call.

When this test runs as part of the test class, runSetDao.getRunSets is called only once! I have no idea why.

service/src/main/java/bio/terra/cbas/initialization/InstanceInitializerBeanConfig.java
```diff
@@ -0,0 +1,23 @@
...
+
+import bio.terra.cbas.config.CbasContextConfiguration;
+import bio.terra.cbas.dao.MethodDao;
+import bio.terra.cbas.dao.MethodVersionDao;
+import bio.terra.cbas.dao.RunDao;
+import bio.terra.cbas.dao.RunSetDao;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
```
mspector, line 11: todo: is this needed?

service/src/main/resources/application.yml
```diff
@@ -41,6 +41,8 @@ cbas:
   context:
     # Should match wds.instanceId and sam.workspaceId:
     workspaceId: "00000000-0000-0000-0000-000000000000"
+    # workspaceCreatedDate is supplied by Leo. When testing locally, manually fill in this value.
+    workspaceCreatedDate: "2024-02-16T14:13:02.500743Z"
```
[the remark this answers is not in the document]
mspector: maybe best to leave this unsupplied, then short circuit if it's not there

service/src/main/java/bio/terra/cbas/initialization/InstanceInitializerBeanConfig.java
```diff
@@ -0,0 +1,23 @@
...
+public class InstanceInitializerBeanConfig {
+
+  @Bean
+  public CloneRecoveryBean cloneRecoveryBean(
+      RunSetDao runSetDao,
+      RunDao runDao,
```
salonishah11, line 17: since we don't update anything in run table do we need this here?

mspector: ah, good catch. you're right!

service/src/main/java/bio/terra/cbas/initialization/InstanceInitializerBeanConfig.java
mspector: update: it turns out *is* needed

service/src/main/resources/application.yml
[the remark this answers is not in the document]
mspector: done

salonishah11 (approved):

service/src/main/java/bio/terra/cbas/initialization/BackfillOriginalWorkspaceIdBean.java
```diff
@@ -0,0 +1,58 @@
...
+    this.methodVersionDao = methodVersionDao;
+    this.cbasContextConfig = cbasContextConfig;
+  }
+
+  public void backfillOriginalWorkspaceIds() {
+    Boolean workspaceCreatedDateIsValid = false;
+
+    try {
+      // call this without assignment to make sure it's not empty,
+      // e.g. during local testing, or due to an upstream error
```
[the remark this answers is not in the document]
cjllanwarne: As a minor side-comment, setting up try/catches and throwing and catching exceptions is significantly harder for the JVM than following if-else paths. So if there's a way to check the validity of this value with a simple logic test, that might be even nicer.

service/src/main/java/bio/terra/cbas/initialization/BackfillOriginalWorkspaceIdBean.java
```diff
@@ -0,0 +1,58 @@
...
+      cbasContextConfig.getWorkspaceCreatedDate();
+      workspaceCreatedDateIsValid = true;
+    } catch (java.time.format.DateTimeParseException e) {
+      logger.error(
+          "Aborting backfillOriginalWorkspaceIds; There was an error parsing workspaceCreatedDate: {}",
+          e.getMessage());
```
cjllanwarne, line 40: Are we likely to ever look for this message and do anything about it?

It seems like an outcome that causes CBAS to fail to start up (and force us to investigate) might be nicer in the long run than not updating the database and potentially losing data if we run straight into the clone recovery despite the backfilling not happening

service/src/main/java/bio/terra/cbas/initialization/InstanceInitializer.java
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.stereotype.Component;
+
+@Component
+public class InstanceInitializer implements ApplicationListener<ContextRefreshedEvent> {
+
+  private final BackfillOriginalWorkspaceIdBean backfillOriginalWorkspaceIdBean;
+
+  public InstanceInitializer(BackfillOriginalWorkspaceIdBean backfillOriginalWorkspaceIdBean) {
+    this.backfillOriginalWorkspaceIdBean = backfillOriginalWorkspaceIdBean;
```
cjllanwarne, line 14: Maybe I was mistaken above (about not calling this a bean).... it certainly seems like it's being autowired here. But it's not marked as something that can be considered a bean 🤔

service/src/main/java/bio/terra/cbas/initialization/InstanceInitializerBeanConfig.java
cjllanwarne: I believe without this, Spring would never find the `@Bean` inside it

service/src/main/java/bio/terra/cbas/initialization/InstanceInitializerBeanConfig.java
```diff
@@ -0,0 +1,22 @@
...
+import bio.terra.cbas.dao.MethodVersionDao;
+import bio.terra.cbas.dao.RunSetDao;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class InstanceInitializerBeanConfig {
+
+  @Bean
+  public BackfillOriginalWorkspaceIdBean backfillOriginalWorkspaceIdBean(
```
cjllanwarne, line 14: Oh (can you tell I'm putting this together as I walk through the code)... this is how the bean gets constructed.

Feel free to double check my terminology but I believe that _this_ is a bean, and the `InstanceInitializer` gets a bean, but the `BackfillOriginalWorkspaceIdBean` class is not necessarily anything bean-y.

So I guess it's up to you, my minor thought would be to rename the `BackfillOriginalWorkspaceIdBean` something more generic (as suggested above), but leave the named references to it as being called "`backfillOriginalWorkspaceIdBean`". What do you think?

service/src/main/java/bio/terra/cbas/util/BackfillOriginalWorkspaceIds.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+
+/*
+NOTE: This class is inherently temporary.  Once all original workspace IDs have been backfilled,
+      this entire class should be deleted, along with its method calls in BackfillOriginalWorkspaceIdBean.java
+ */
+public class BackfillOriginalWorkspaceIds {
```
cjllanwarne, line 19: Now we know how.....

after Doug's presentation this morning I'd be tempted to:
- Make this a `BackfillOriginalWorkspaceIdsService`, with an `@Service` annotation.
- Have the Dao dependencies autowired instead of provided via parameters per method
- Wrap the chains of interactions with `@WriteTransaction` to make sure there are no race conditions possible

I actually don't think it would make _that_ much difference in this particular case, because we will never be writing un-backfilled values so there's no way for them to slip in unnoticed... but it might be good practice

service/src/test/java/bio/terra/cbas/util/methods/TestBackfillOriginalWorkspaceIds.java
```diff
@@ -0,0 +1,289 @@
...
+
+  private Method methodClonedBackfilled =
+      new Method(
+          UUID.fromString("00000000-0000-0000-0000-000000000030"),
+          "MethodClonedBackfilled",
+          "A method that was cloned from a source workspace, with a backfilled originalWorkspaceId",
+          creationDateBefore,
+          UUID.randomUUID(), // this can be random because it isn't relevant to these tests.
+          "GitHub",
+          originalWorkspaceId);
```
cjllanwarne, line 152: If this was cloned and then backfilled, shouldn't this be all-zeros?

cjllanwarne: And for run sets and method versions?

service/src/test/java/bio/terra/cbas/util/methods/TestBackfillOriginalWorkspaceIds.java
```diff
@@ -0,0 +1,289 @@
...
+import bio.terra.cbas.models.RunSet;
+import bio.terra.cbas.util.BackfillOriginalWorkspaceIds;
+import java.time.OffsetDateTime;
+import java.util.Arrays;
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TestBackfillOriginalWorkspaceIds {
```
cjllanwarne, line 28: There's an annotation that makes sure that 
1. All of the mock methods you set up get called
2. And no other mock methods are called unexpectedly

Will try to find this. I'm pretty sure we use it elsewhere in the code

service/src/main/java/bio/terra/cbas/initialization/BackfillOriginalWorkspaceIdBean.java
```diff
@@ -0,0 +1,58 @@
...
+
+import bio.terra.cbas.config.CbasContextConfiguration;
+import bio.terra.cbas.dao.MethodDao;
+import bio.terra.cbas.dao.MethodVersionDao;
+import bio.terra.cbas.dao.RunSetDao;
+import bio.terra.cbas.util.BackfillOriginalWorkspaceIds;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BackfillOriginalWorkspaceIdBean {
```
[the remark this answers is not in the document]
mspector: I'll never be put off by comments about properly naming things. I'll make some change like that...

service/src/main/java/bio/terra/cbas/initialization/BackfillOriginalWorkspaceIdBean.java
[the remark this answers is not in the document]
mspector: > As a minor side-comment, setting up try/catches and throwing and catching exceptions is significantly harder for the JVM than following if-else paths

Interesting, I didn't realize that - I'll give that a try

service/src/main/java/bio/terra/cbas/initialization/InstanceInitializerBeanConfig.java
mspector: and I am responding as I read your comments, without perfect memory of what I wrote!

let me think about it a bit more carefully and get back to you

service/src/test/java/bio/terra/cbas/util/methods/TestBackfillOriginalWorkspaceIds.java
mspector: whoops, you're right. my intention with these methods/run sets/method versions was to show that the backfiller would skip them if they had any originalWorkspaceId value. I'll change it for clarity, but I don't think it invalidates the tests

service/src/main/java/bio/terra/cbas/initialization/BackfillOriginalWorkspaceIdBean.java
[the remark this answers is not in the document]
mspector: in the end, the best solution (IMO) turned out to be: designate a special timestamp (`"0000-01-01T00:00:00.000000Z"`) that short circuits the backfiller.

service/src/main/java/bio/terra/cbas/initialization/InstanceInitializer.java
```diff
@@ -0,0 +1,24 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.stereotype.Component;
+
+@Component
+public class InstanceInitializer implements ApplicationListener<ContextRefreshedEvent> {
+
+  @Autowired BackfillOriginalWorkspaceIdService backfillOriginalWorkspaceIdService;
```
cjllanwarne, line 13: I think it should be generally preferred to make this `private final` provided via the constructor (which makes it auto-auto-wired). Then you shouldn't need the explicit autowired annotation
```suggestion
  private final BackfillOriginalWorkspaceIdService backfillOriginalWorkspaceIdService;
```

service/src/main/java/bio/terra/cbas/util/BackfillOriginalWorkspaceIdService.java
```diff
@@ -0,0 +1,175 @@
...
+NOTE: This class is inherently temporary.  Once all original workspace IDs have been backfilled,
+      this entire class should be deleted, along with its method calls in BackfillOriginalWorkspaceIdBean.java
+ */
+@Service
+public class BackfillOriginalWorkspaceIdService {
+  private final Logger logger = LoggerFactory.getLogger(BackfillOriginalWorkspaceIdService.class);
+  @Autowired CbasContextConfiguration cbasContextConfig;
+  @Autowired RunSetDao runSetDao;
+  @Autowired MethodDao methodDao;
+  @Autowired MethodVersionDao methodVersionDao;
```
cjllanwarne: Same comment about changing autowired to private final

service/src/main/java/bio/terra/cbas/util/BackfillOriginalWorkspaceIdService.java
```diff
@@ -0,0 +1,175 @@
...
+    this.methodVersionDao = methodVersionDao;
+  }
+
+  @WriteTransaction
+  public void backfillAll() {
+    if (cbasContextConfig
+        .getWorkspaceCreatedDate()
+        .isEqual(OffsetDateTime.parse("0000-01-01T00:00:00.000000Z"))) {
+      logger.info(
+          "Skipping BackfillOriginalWorkspaceId operation. This could indicate that workspaceCreatedDate in application.yml has not been configured properly.");
```
cjllanwarne, line 48: > This could indicate that workspaceCreatedDate in application.yml has not been configured properly

What else could it indicate? Config not getting through feels like a error level problem to me (rather than info). Is there any valid situation where this would be left unconfigured?

In the worse case, this feels risky:
- Config is bad
- Backfilling is skipped (instead of failing)
- Cleanup continues and deletes a bunch of content that should have been backfilled

It feels safer (since this is more of a "we always expect it to be configured properly) to error out at this point.

What do you think?

service/src/main/java/bio/terra/cbas/util/BackfillOriginalWorkspaceIdService.java
```diff
@@ -0,0 +1,175 @@
...
+        backfillMethodVersions();
+      } catch (Exception e) {
+        // we want CBAS to fail if this operation fails, just like a liquibase migration
+        throw new RuntimeException(
+            "There was a problem while trying to backfill original workspace IDs: %s".formatted(e));
+      }
+    }
+  }
+
+  @WriteTransaction
```
cjllanwarne, line 62: This is kind of a giant transaction... can we make it any more specific to things that really need to happen atomically?

cjllanwarne: (eg, each runset is updated in its own transaction?)

service/src/main/resources/application.yml
```diff
@@ -41,6 +41,10 @@ cbas:
   context:
     # Should match wds.instanceId and sam.workspaceId:
     workspaceId: "00000000-0000-0000-0000-000000000000"
+    # workspaceCreatedDate is supplied by Leo.
+    # When testing locally, manually fill in this value, e.g. "2024-02-16T14:13:02.500743Z".
+    # BackfillOriginalWorkspaceIdService will be skipped if the default value below is used.
+    workspaceCreatedDate: "0000-01-01T00:00:00.000000Z"
```
cjllanwarne, line 47: I wasn't fully sure why this needed to be a magic value instead of left as null or empty?

service/src/test/java/bio/terra/cbas/util/methods/TestBackfillOriginalWorkspaceIdService.java
```diff
@@ -0,0 +1,298 @@
...
+
+public class TestBackfillOriginalWorkspaceIdService {
+  @MockBean RunSetDao runSetDao;
+  @MockBean MethodDao methodDao;
+  @MockBean MethodVersionDao methodVersionDao;
+  @MockBean CbasContextConfiguration cbasContextConfig;
+
+  @Test
+  void testBackfillMethods() {
+    CbasContextConfiguration cbasContextConfig = mock(CbasContextConfiguration.class);
```
cjllanwarne, line 35: Because you're making these mocks explicitly and then [calling](https://github.com/DataBiosphere/cbas/pull/255/files#diff-7ed4393471905195d5e9217a76c55a3421e7141b80b72b61c22cb4f753da4535R49) the standard constructor, I don't think they need to be marked as mockbeans above. That would be necessary if you were letting spring autowire them through

service/src/main/resources/application.yml
mspector: there may be a better way to do all of this, but the reason I'm not using an empty string is because parsing this to an OffsetDateTime will fail here: https://github.com/DataBiosphere/cbas/pull/255/files#diff-aa4c1c1284253e9e5472f8179c0118a9e3f28f26eb2afdbeacd3c2a00ab3fba3R19

I assume null would cause it to fail too, but didn't confirm that

service/src/main/java/bio/terra/cbas/util/BackfillOriginalWorkspaceIdService.java
mspector: you're right... I arranged things this way to get around some problems related to the integration tests, but I think I should find a different way.

service/src/main/java/bio/terra/cbas/util/BackfillOriginalWorkspaceIdService.java
mspector: yeah, thanks for pointing this out - my thinking was: "we want all this to succeed or fail together, so put it in a transaction". but I see your point completely

service/src/main/resources/application.yml
mspector: I'll leave this null, and override it in the integration test github action

service/src/main/java/bio/terra/cbas/util/BackfillOriginalWorkspaceIdService.java
mspector: looking at this more closely - each of these methods (e.g. for run_set, method, method_version...) only makes one update per object type per iteration (e.g. it loops over a list of run sets, and calls one DAO method)

that makes me think that the WriteTransaction annotation is not really needed here at all.

hopefully I'm not misunderstanding something important

service/src/main/java/bio/terra/cbas/util/BackfillOriginalWorkspaceIdService.java
```diff
@@ -39,27 +37,18 @@ public BackfillOriginalWorkspaceIdService(
...
-        throw new RuntimeException(
-            "There was a problem while trying to backfill original workspace IDs: %s".formatted(e));
-      }
+    try {
+      backfillMethods();
+      backfillRunSets();
+      backfillMethodVersions();
+    } catch (Exception e) {
+      // we want CBAS to fail if this operation fails, just like a liquibase migration
+      throw new RuntimeException(
```
cjllanwarne, line 47: What happens if this is thrown? This function is called by an event hook so it might be outside of the main app processing thread, so throwing an exception might just get lost

service/src/test/java/bio/terra/cbas/util/methods/TestBackfillOriginalWorkspaceIdService.java
```diff
@@ -21,22 +21,28 @@
...
+  MethodDao methodDao;
+  MethodVersionDao methodVersionDao;
+  CbasContextConfiguration cbasContextConfig;
+
+  @BeforeEach
+  void setup() {
+    cbasContextConfig = mock(CbasContextConfiguration.class);
+    runSetDao = mock(RunSetDao.class);
+    methodDao = mock(MethodDao.class);
+    methodVersionDao = mock(MethodVersionDao.class);
```
cjllanwarne, line 38: 💯

service/src/main/java/bio/terra/cbas/util/BackfillOriginalWorkspaceIdService.java
mspector: it fails to boot! that's why this change is needed: https://github.com/DataBiosphere/cbas/pull/255/files/2bd5e96583eb94371f90405171d6fa6852ad709a..a1b8bb7631f36b0b37a5de51be1e1c66657e8847#diff-bc668a2c9f2299cef15b222055b4b4d5311646caec2e7610e540cee18ae9b948R145

PR #258, opened 2024-03-17
[WM-2414] clone recovery

mspector:


service/build.gradle
```diff
@@ -74,6 +74,9 @@ dependencies {
     testImplementation('au.com.dius.pact.provider:junit5spring:4.3.19')
     testImplementation('au.com.dius.pact.consumer:junit5:4.3.19')
     testImplementation("au.com.dius.pact.consumer:junit:4.3.19")
+    testImplementation('org.testcontainers:junit-jupiter')
+    testImplementation('org.testcontainers:postgresql')
+    testImplementation('org.springframework.boot:spring-boot-testcontainers:3.2.3')
```
cjllanwarne, line 79: Could we extract testcontainers into a separate PR. It'd be nice to remove the noise from the clone recovery changes AND make the PRs more targeted and easier to find if we want to reference them in the future

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
```diff
@@ -50,22 +58,40 @@ class TestMethodDao {
...
+          .withDatabaseName("test_db")
+          .withUsername("test_user")
+          .withPassword("test_password");
+
+  @DynamicPropertySource
+  static void postgresProperties(DynamicPropertyRegistry registry) {
+    registry.add("spring.datasource.jdbc-url", postgres::getJdbcUrl);
+    registry.add("spring.datasource.username", postgres::getUsername);
+    registry.add("spring.datasource.password", postgres::getPassword);
+  }
```
cjllanwarne, line 73: This potentially looks like a good candidate for extracting up into an abstract class that all of these DAO tests can inherit? Is that possible?

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
```diff
@@ -50,22 +58,40 @@ class TestMethodDao {
...
+
   @BeforeAll
-  void setUp() {
+  static void setup() {
+    postgres.start();
```
cjllanwarne, line 77: This means that each test reuses the same test container as the previous one... is that deliberate? I guess it's more efficient if the tests never interfere with each other but feels potentially risky?

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
```diff
@@ -50,22 +58,40 @@ class TestMethodDao {
...
-    int recordsDeleted2 = methodDao.deleteMethod(methodId2);
-
-    assertEquals(1, recordsDeleted1);
-    assertEquals(1, recordsDeleted2);
+  @AfterEach
+  void cleanupDb() throws SQLException {
+    DriverManager.getConnection(
+            postgres.getJdbcUrl(), postgres.getUsername(), postgres.getPassword())
+        .createStatement()
+        .execute("DELETE FROM run_set; DELETE FROM method_version; DELETE FROM method;");
```
cjllanwarne, line 94: This might also be good to extract up into an abstract super class?

cjllanwarne, line 94: Should we also have `DELETE FROM run`?

cjllanwarne, line 94: And github method details?

mspector, line 94: make sure to delete github method details and any other tables that I've missed here

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
cjllanwarne: Never mind - the aftereach makes sure they are all independent!

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
mspector: I typed my previous comment before seeing this one -- yes!

mspector:
closing this PR in favor of https://github.com/DataBiosphere/cbas/pull/261

PR #259, opened 2024-03-18
Wm 2501 verify private methods kp

kpierre13:


service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -200,7 +204,9 @@ public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMetho
         String repository = githubUrlComponents.repo();
         String organization = githubUrlComponents.org();
         branchOrTagName = githubUrlComponents.branchOrTag();
-        Boolean isPrivate = gitHubService.isRepoPrivate(organization, repository);
+
+        String token = ecmService.getAccessToken();
```
cjllanwarne, line 208: Can we call this githubToken to avoid any confusion?

service/src/main/java/bio/terra/cbas/dependencies/github/GitHubClient.java
```diff
@@ -1,60 +1,46 @@
...
 
-  public GitHubClient(GitHubConfiguration gitHubConfiguration) {
+  public GitHubClient() {
     this.client = ClientBuilder.newClient();
-    this.token = gitHubConfiguration.personalAccessToken();
     this.gson = new Gson();
   }
 
-  public RepoInfo getRepo(String organization, String repo) throws GitHubClientException {
+  public JSONObject getRepo(String organization, String repo, String token)
```
cjllanwarne, line 23: CBAS imports various libraries that can convert JSONObject into java classes. Our imported libraries for other dependencies all use them behind the scenes. It might be worth a quick investigation into whether we can use one of those to convert the json we get back into a well defined structure instead of a json object which the call just has to "know" how to use.

service/src/test/java/bio/terra/cbas/dependencies/github/TestGithubService.java
```diff
@@ -0,0 +1,52 @@
...
+          "name": "foo",
+          "organization": "broadinstitute",
+          "private": %s
+        }
+        """
+            .formatted("false")
+            .stripIndent()
+            .trim();
+
+    String jsonResponseTrue =
```
cjllanwarne, line 30: Since they're so similar, I would make this a function like `public String jsonResponse(boolean isPrivate)` and just provide true/false as needed in the test cases

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -420,7 +422,7 @@ void validPostRequest() throws Exception {
         objectMapper.readValue(validWorkflowDescriptionJson, WorkflowDescription.class);
     when(cromwellService.describeWorkflow(validRawWorkflow))
         .thenReturn(workflowDescForValidWorkflow);
-    when(gitHubService.isRepoPrivate(any(), any())).thenReturn(true);
+    when(gitHubService.isRepoPrivate(any(), any(), any())).thenReturn(true);
```
cjllanwarne, line 425: Can we check that this is recorded properly in the database call?

cjllanwarne (approved):

service/src/main/java/bio/terra/cbas/config/BeanConfig.java
```diff
@@ -59,4 +60,11 @@ public cromwell.client.ApiClient cromwellAuthReadClient(
       BearerToken bearerToken, CromwellClient cromwellClient) {
     return cromwellClient.getAuthReadApiClient(bearerToken.getToken());
   }
+
+  @Bean("ecmAuthClient")
+  @RequestScope
```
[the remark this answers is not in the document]
kpierre13: Is there an example of this somewhere I can look at?

service/src/main/java/bio/terra/cbas/dependencies/github/GitHubClient.java
kpierre13: Luckily, this was able to be resolved with a `@SerializedName` annotation on the field.

service/src/main/java/bio/terra/cbas/config/BeanConfig.java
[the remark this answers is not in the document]
mspector: he must be talking about the example repo that Doug showed us during our Spring Boot session:

https://github.com/DataBiosphere/terra-java-project-template/blob/main/service/src/main/java/bio/terra/javatemplate/controller/ExampleController.java#L49

mspector (approved):
I'll second Chris's comment about doing things "the right way" according to the java project template, if it isn't too complicated

PR #261, opened 2024-03-20
[WM-2414][WM-2557] clone recovery

mspector:
See previous comments here: https://github.com/DataBiosphere/cbas/pull/258


service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
```diff
@@ -0,0 +1,97 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class CloneRecoveryService {
+
+  private final Logger logger = LoggerFactory.getLogger(CloneRecoveryService.class);
+  @Autowired private final RunDao runDao;
```
mspector, line 21: "@Autowired" probably not needed

service/src/test/java/bio/terra/cbas/initialization/TestCloneRecoveryService.java
```diff
@@ -0,0 +1,291 @@
...
+import bio.terra.cbas.models.*;
+import java.time.OffsetDateTime;
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.*;
+import org.mockito.Mock;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class TestCloneRecoveryService extends ContainerizedDaoTest {
```
cjllanwarne, line 21: Feel free to push back on this, kind of an opinionated statement:

In an ideal world, I think a Service test (which validates business logic) would not need test containers. It would be 75% unit tests checking in-memory data transformations are calculated properly and 25% mocked-dao tests validating that the glue was calling the right functions on the dao. Then the test containers would only be needed for the very final "did our database connection commands get implemented correctly in SQL queries"

mspector: I've developed some thoughts as I actually work on this - maybe we can find a time to talk later today

mspector: I'll write them down to clarify my thoughts, but let's still chat...

- I agree with your philosophy, 100%
- I think the mistake here is calling this a "service"
- I wrote the tests this way because it gave me confidence that it was working as intended. As I try to rewrite, I feel less good about the tests that emerge
- Each of the three helper methods that `pruneCloneSourceWorkspaceHistory` uses feel like they could be DAO methods - I just happened to have written them with Java and existing DAO methods rather than new SQL

So I think what might make the most sense is to move the helpers to their respective DAO classes, and do some basic unit testing (with mocks) of `pruneCloneSourceWorkspaceHistory`.

buildSrc/src/main/groovy/java-common-conventions.gradle
```diff
@@ -14,9 +14,7 @@ if (!isCiServer) {
     tasks.withType(JavaExec).configureEach {
         systemProperty 'spring.profiles.include', 'human-readable-logging'
     }
-    tasks.withType(Test).configureEach {
-        systemProperty 'spring.profiles.include', 'human-readable-logging'
-    }
+    tasks.withType(Test).configureEach {}
```
mspector, line 17: I can take this out if others feel strongly, but I think its presence is a way of saying "this is how you configure tests at the top level" and "there is no configuration right now"

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public Boolean isMethodCloned(Method m) {
+    return !m.originalWorkspaceId().equals(cbasContextConfig.getWorkspaceId());
+  }
+
+  public Boolean isRunSetCloned(RunSet r) {
+    return !r.originalWorkspaceId().equals(cbasContextConfig.getWorkspaceId());
+  }
+
+  @WriteTransaction
```
cjllanwarne, line 62: I believe transactions are re-entrant rather than nested, so this will probably just be part of the same original transaction as you created in the caller... except that it's called from the same class so would be ignored anyway.

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
```diff
@@ -0,0 +1,96 @@
...
+    this.cbasContextConfig = cbasContextConfig;
+  }
+
+  public void cloneRecovery() {
+    logger.info(
+        "Starting clone recovery (workspaceId: {}, workspaceCreatedDate: {})",
+        cbasContextConfig.getWorkspaceId(),
+        cbasContextConfig.getWorkspaceCreatedDate());
+
+    pruneCloneSourceWorkspaceHistory();
```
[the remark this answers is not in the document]
cjllanwarne: I guess we then call other functions from inside the same class anyway, so this only moves the problem one call further down.

Hmm, this might be a use case for the `@Autowired CloneRecoveryService self` pattern that we saw in the spring intro

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
```diff
@@ -0,0 +1,96 @@
...
+  @WriteTransaction
+  public void updateMethodTemplate(Method m) {
+    List<RunSet> methodClonedRunSets =
+        runSetDao.getRunSetsWithMethodId(m.methodId()).stream()
+            .filter(this::isRunSetCloned)
+            .toList();
+
+    if (!methodClonedRunSets.isEmpty()) {
+      RunSet newTemplateRunSet = methodClonedRunSets.get(0);
+      runSetDao.updateIsTemplate(newTemplateRunSet.runSetId(), true);
```
cjllanwarne, line 71: I think there's a minor gotcha here. Imagine if:

- User clones a workspace with a method M and runset RS1
- User runs another instance, as RS2

Desired outcome (I think?):
- RS1 deleted
- RS2 remains in the DB

But actually, both would be kept as templates.

I don't think this is going to be a big deal _today_ because we only show the most recent run sets anyway, but if we ever give users a drop down "select from all your saved config templates" then this extra one might unexpectedly show up.

service/src/main/java/bio/terra/cbas/initialization/InstanceInitializer.java
```diff
@@ -8,16 +9,25 @@
...
     this.backfillOriginalWorkspaceIdService = backfillOriginalWorkspaceIdService;
+    this.cloneRecoveryService = cloneRecoveryService;
   }
 
   @Override
   public void onApplicationEvent(@NotNull ContextRefreshedEvent event) {
-    backfillOriginalWorkspaceIdService.backfillAll();
+    if (cbasInitializationConfiguration.getEnabled()) {
+      backfillOriginalWorkspaceIdService.backfillAll();
+      cloneRecoveryService.cloneRecovery();
```
cjllanwarne, line 30: Is there any way to check that backfilling succeeded before rolling in to recovery? Or validation like "there are no null originalWorkspaceIds any more"?

cjllanwarne (commented):
I didn't look at the TestCloneRecoveryService in detail, I think you're still working on that. Some comments on the other classes though

.github/workflows/build-and-test.yml
```diff
@@ -142,7 +142,7 @@ jobs:
           echo "${{ secrets.PERF_TESTRUNNER_ACCOUNT_B64 }}" | base64 -d > "integration/src/main/resources/rendered/testrunner-perf.json"
 
       - name: Launch the background process for integration tests
-        run: ./gradlew --build-cache bootRun --args='--cbas.context.workspaceCreatedDate=1970-01-00T00:00:00.000000Z' | tee bootrun.log &
+        run: ./gradlew --build-cache bootRun --args='--cbas.context.workspaceCreatedDate=1970-01-01T00:00:00.000000Z' | tee bootrun.log &
```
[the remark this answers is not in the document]
mspector: yes -- I don't know why I didn't catch it the first time, but the original string is actually not a valid OffsetDateTime

service/src/main/java/bio/terra/cbas/config/CbasInitializationConfiguration.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.cbas.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties(prefix = "cbas.initialization")
+public class CbasInitializationConfiguration {
+  private Boolean enabled;
```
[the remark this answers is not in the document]
mspector: I added it so that tests could disable it if necessary. it seemed like the two best options were either to set a dummy workspaceCreatedDate or use this "enabled" flag, and setting a dummy workspaceCreatedDate in a test that didn't touch that value (like TestSamService) seemed strange to me. But let me know if you disagree - I was pretty torn on it.

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
mspector: As I understand it:
- before cloning, there are actually three runSets: RS1, RS2, and the template from the original workspace.
- we want the cloned workspace to start with the configuration of RS1, so it gets turned into a template, and the old template gets deleted
- RS2 remains in the DB as a non-template

That behavior is tested here: https://github.com/DataBiosphere/cbas/pull/261/files#diff-0a982347764cb63bb49dfee632e04a756773c56b45b8591e352748e7dd0c9399R141

We should discuss if you don't think this is the right behavior, but that was my intention here

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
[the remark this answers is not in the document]
mspector: hmm I see - good to know. I remember Doug saying that `@Autowired CloneRecoveryService self` can cause strange outcomes, so I may try to come up with some other way...

service/src/test/java/bio/terra/cbas/initialization/TestCloneRecoveryService.java
mspector: Just merged in a substantial refactor to improve testability

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
```diff
@@ -39,58 +41,77 @@ public void cloneRecovery() {
...
-  }
+    templateUpdateManifests.stream()
+        .map(MethodTemplateUpdateManifest::nonTemplateRunSets)
+        .forEach(rsList -> rsList.forEach(rs -> runSetDao.updateIsTemplate(rs.runSetId(), false)));
 
-  public Boolean isMethodCloned(Method m) {
-    return !m.originalWorkspaceId().equals(cbasContextConfig.getWorkspaceId());
+    clonedMethods.forEach(
+        method -> {
+          List<RunSet> methodRunSets = runSetDao.getRunSetsWithMethodId(method.methodId());
```
cjllanwarne, line 63: This looks like a duplicate call to line 49. Perhaps we can call this once and save the result?

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
```diff
@@ -39,58 +41,77 @@ public void cloneRecovery() {
...
+          getRunSetsToDelete(methodRunSets).forEach(rs -> runSetDao.deleteRunSet(rs.runSetId()));
```
cjllanwarne, line 64: Is there a reason for this two-stage "mark it as non-template", then delete it later on?

Could we just delete them straight away on line 59?

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
```diff
@@ -39,58 +41,77 @@ public void cloneRecovery() {
...
+          getRunsToDelete(methodRunSets).forEach(r -> runDao.deleteRun(r.runId()));
+        });
   }
 
-  public Boolean isRunSetCloned(RunSet r) {
-    return !r.originalWorkspaceId().equals(cbasContextConfig.getWorkspaceId());
-  }
+  public record MethodTemplateUpdateManifest(
+      List<RunSet> templateRunSets, List<RunSet> nonTemplateRunSets) {}
```
cjllanwarne, line 70: Are these really "keepAsTemplate" and "toBeDeleted"?

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
```diff
@@ -39,58 +41,77 @@ public void cloneRecovery() {
...
+        .map(MethodTemplateUpdateManifest::nonTemplateRunSets)
+        .forEach(rsList -> rsList.forEach(rs -> runSetDao.updateIsTemplate(rs.runSetId(), false)));
 
-  public Boolean isMethodCloned(Method m) {
-    return !m.originalWorkspaceId().equals(cbasContextConfig.getWorkspaceId());
+    clonedMethods.forEach(
+        method -> {
+          List<RunSet> methodRunSets = runSetDao.getRunSetsWithMethodId(method.methodId());
+          getRunSetsToDelete(methodRunSets).forEach(rs -> runSetDao.deleteRunSet(rs.runSetId()));
+          getRunsToDelete(methodRunSets).forEach(r -> runDao.deleteRun(r.runId()));
```
cjllanwarne, line 65: Don't you need to delete the runs before the run sets?

cjllanwarne, line 65: Since you already know which run sets you're wanting to delete, why not just `delete from run where run_set_id = ...` right before deleting the run set itself? Could even be in the same Dao method (or at least transaction)?

service/src/test/java/bio/terra/cbas/dao/TestRunSetDao.java
```diff
@@ -98,36 +100,87 @@ void retrievesAllRunSets() {
...
+  void deleteRunSet() {
+    int response = runSetDao.deleteRunSet(runSet.runSetId());
+    assertEquals(1, response);
+
+    try {
+      RunSet deletedRunSet = runSetDao.getRunSet(runSet.runSetId());
+      Assertions.fail("Expected an exception, but none was thrown.");
+    } catch (Exception e) {
+      assertEquals("Index 0 out of bounds for length 0", e.getMessage());
+    }
```
cjllanwarne, line 184: There's an `assertThrows` syntax used elsewhere in CBAS that might cover this case a little more neatly

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
mspector: it is a duplicate, but the run sets change between line 49 and this one. I could maintain a list of run sets and update their fields as I call methods like `updateIsTemplate`, but this seemed simpler

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
mspector: good question... my tests didn't catch any problems with it. would you expect some kind of foreign key error if I try to delete the run set first? or are you expecting a different problem?

mspector: I was trying to reuse existing methods here, but a new Dao method is a good idea

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
mspector: yes, good thinking

service/src/main/java/bio/terra/cbas/initialization/CloneRecoveryService.java
mspector: I think you're right in the bigger context, I wrote it this way to keep the logic of each method self-contained (for testing purposes)

service/src/main/java/bio/terra/cbas/initialization/cloneRecovery/CloneRecoveryTransactionService.java
```diff
@@ -0,0 +1,38 @@
...
+  private final RunDao runDao;
+  private final RunSetDao runSetDao;
+
+  public CloneRecoveryTransactionService(RunSetDao runSetDao, RunDao runDao) {
+    this.runSetDao = runSetDao;
+    this.runDao = runDao;
+  }
+
+  @WriteTransaction
+  public void deleteRuns(UUID runSetId) {
```
cjllanwarne, line 21: I think this can be private, and I think it can be a non-transaction

service/src/main/java/bio/terra/cbas/initialization/cloneRecovery/CloneRecoveryTransactionService.java
```diff
@@ -0,0 +1,38 @@
...
+import bio.terra.common.db.WriteTransaction;
+import java.util.UUID;
+import org.springframework.stereotype.Service;
+
+@Service
+public class CloneRecoveryTransactionService {
+  private final RunDao runDao;
+  private final RunSetDao runSetDao;
+
+  public CloneRecoveryTransactionService(RunSetDao runSetDao, RunDao runDao) {
```
cjllanwarne, line 15: This class looks simple enough that unit tests are probably not strictly required, but if you wanted to wire up a couple of mocks in a unit test to prove that “convert to template” was doing the right thing in a couple of situations, and “deleteRunSetAndRuns” was doing the right thing in a couple of situations, I think it would give us a little bit of confidence now and into the future

service/src/test/java/bio/terra/cbas/initialization/TestCloneRecoveryServiceIntegration.java
```diff
@@ -22,6 +24,7 @@ public class TestCloneRecoveryServiceIntegration extends ContainerizedDatabaseTe
   @Autowired MethodDao methodDao;
   @Autowired RunSetDao runSetDao;
   @Autowired RunDao runDao;
+  @Autowired CloneRecoveryTransactionService transactionService;
```
cjllanwarne, line 27: Now that this is `CloneRecoveryService` is doing all the calculations ahead of time, I think you should be able to unit-test that logic in unit tests, and mock-dao test the transaction level in a `CloneRecoveryTransactionServiceTest` to prove that it's calling down to the right low-level DB functions. Then the only "database integration" tests you need would be super thin "does the DB function do what it's supposed to when building SQL". And this whole class would probably be redundant

service/src/test/java/bio/terra/cbas/initialization/TestCloneRecoveryServiceUnits.java
```diff
@@ -50,67 +49,15 @@ void generateTemplateUpdateManifest() {
...
     // the latest cloned run set should be the new template.
     // all other clone run sets should be listed as non-templates.
-    assertEquals(1, manifest.templateRunSets().size());
-    assertEquals(2, manifest.nonTemplateRunSets().size());
-    assertTrue(manifest.templateRunSets().contains(clonedRunSetLatest));
-    assertTrue(manifest.nonTemplateRunSets().contains(clonedTemplate));
-    assertTrue(manifest.nonTemplateRunSets().contains(clonedRunSet));
+    assertEquals(1, manifest.keepAsTemplate().size());
+    assertEquals(2, manifest.toBeDeleted().size());
+    assertTrue(manifest.keepAsTemplate().contains(clonedRunSetLatest));
```
cjllanwarne, line 54: Could you test actual set membership instead of just contains?
eg `assertEquals(manifest.keepAsTemplate, List.of("clonedRunSetLatest"))`

service/src/main/java/bio/terra/cbas/initialization/cloneRecovery/CloneRecoveryTransactionService.java
mspector: I've added unit tests, but I could not come up with more than one situation to test these methods in -- I think my single test for each method covers everything. But if I'm wrong about that please let me know.

service/src/test/java/bio/terra/cbas/initialization/TestCloneRecoveryService.java
```diff
@@ -51,9 +49,8 @@ void generateTemplateUpdateManifest() {
     // all other clone run sets should be listed as non-templates.
     assertEquals(1, manifest.keepAsTemplate().size());
     assertEquals(2, manifest.toBeDeleted().size());
```
cjllanwarne, line 51: Minor - now that you are testing exact membership, I think these counts are automatically validated too

cjllanwarne (approved):
Looks awesome! Thanks for humoring me through the various refactors!

kpierre13 (approved):
Thank you for all of your hard work on this!

service/src/test/java/bio/terra/cbas/initialization/TestCloneRecoveryService.java
mspector: ah, yes you're right

PR #266, opened 2024-03-27
[WM-2394] clone recovery fix last run

mspector:
this has been tested in a BEE


service/src/main/java/bio/terra/cbas/dao/MethodDao.java
```diff
@@ -97,6 +97,13 @@ public int updateLastRunWithRunSet(RunSet runSet) {
                 "method_id", runSet.methodVersion().method().methodId())));
   }
 
+  public int unsetLastRunSetId(UUID methodId) {
+    String sql =
+        "UPDATE method SET %s = NULL WHERE %s = :method_id"
+            .formatted(Method.LAST_RUN_SET_ID_COL, Method.METHOD_ID_COL);
+    return jdbcTemplate.update(sql, new MapSqlParameterSource(Map.of("method_id", methodId)));
+  }
```
mspector, line 105: I'm adding these new DAO methods because "updateLastRunWithRunSet" is (apparently) not nullable. I could have tried to get around that, but having a separate explicit DAO method for unsetting this seemed reasonable to me.

cjllanwarne: This seems sensible. It kind of makes sense that this is a separate activity from updating it

service/src/main/java/bio/terra/cbas/initialization/cloneRecovery/CloneRecoveryTransactionService.java
```diff
@@ -1,37 +1,45 @@
...
   }
 
-  private void deleteRuns(UUID runSetId) {
+  private void deleteRuns(RunSet runSet) {
     runDao
-        .getRuns(new RunDao.RunsFilters(runSetId, null))
+        .getRuns(new RunDao.RunsFilters(runSet.runSetId(), null))
         .forEach(r -> runDao.deleteRun(r.runId()));
+    methodDao.unsetLastRunSetId(runSet.methodVersion().getMethodId());
+    methodVersionDao.unsetLastRunSetId(runSet.getMethodVersionId());
```
cjllanwarne, line 31: Nit: This is kind of a surprise addition to `deleteRuns` IMO. Today we only use it in two known places but if we re-use this in the future it might surprise us that it's also unsetting last run sets. 

Perhaps this belongs in a separate `unsetLastRunSetIdRecords` function?

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
```diff
@@ -83,4 +83,11 @@ void retrievesAllMethods() {
     assertEquals(method2.methodId(), allMethods.get(0).methodId());
     assertEquals(method1.methodId(), allMethods.get(1).methodId());
   }
+
+  @Test
+  void unsetLastRunSetId() {
+    methodDao.unsetLastRunSetId(method1.methodId());
```
cjllanwarne, line 89: This looks like it's already null in `method1`?

To prove me wrong, perhaps have an explicit "it's there!" check followed by the "unset" action and the "it's not there any more!" check

service/src/test/java/bio/terra/cbas/dao/TestMethodVersionDao.java
```diff
@@ -85,4 +85,12 @@ void retrievesMethodVersionsForMethod() {
     assertEquals(branch, actual.get(0).branchOrTagName());
     assertNull(actual.get(0).lastRunSetId());
   }
+
+  @Test
+  void unsetLastRunSetId() {
+    methodVersionDao.unsetLastRunSetId(methodVersion.methodVersionId());
```
cjllanwarne, line 91: Same question here

cjllanwarne (approved):
General approval with a couple of minor comments

service/src/main/java/bio/terra/cbas/initialization/cloneRecovery/CloneRecoveryTransactionService.java
mspector: ah, I hadn't considered that. I'll move it somewhere else

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
kpierre13: +1 to Chris' comment

mspector: thanks - careless of me

kpierre13 (approved):

service/src/test/java/bio/terra/cbas/initialization/TestCloneRecoveryTransactionService.java
```diff
@@ -46,6 +46,8 @@ void convertToTemplate() {
     verify(runDao).deleteRun(run1.runId());
     verify(runDao).deleteRun(run2.runId());
     verify(runDao).deleteRun(run3.runId());
+    verify(methodDao).unsetLastRunSetId(method.methodId());
```
cjllanwarne, line 49: I would add `@ExtendWith(MockitoExtension.class)` to your class definition, that makes sure everything you set up is called, and nothing you don't set up is called, without having to explicitly call like this to check

cjllanwarne (approved):

service/src/test/java/bio/terra/cbas/initialization/TestCloneRecoveryTransactionService.java
mspector: just talked about this offline with Chris, but for the record: I'm adding this extension, but *not* removing any of these calls to `verify`, because they are on void methods, and it's not clear how the mocks would be set up to specify the particular inputs I'm expecting for those methods

PR #267, opened 2024-03-27
[WM-2563] CBAS handles public workflows with no token

kpierre13:


service/src/main/java/bio/terra/cbas/dependencies/github/GitHubClient.java
```diff
@@ -24,14 +24,23 @@ public RepoInfo getRepo(String organization, String repo, String token)
...
+              .get();
+    } else {
+      response =
+          target
+              .request(MediaType.APPLICATION_JSON_TYPE)
+              .header("Accept", "application/vnd.github+json")
+              .header("Authorization", "Bearer " + token)
+              .header("X-GitHub-Api-Version", "2022-11-28")
+              .get();
+    }
```
salonishah11, line 43: TOL: this can be made less repetitive by extracting the common bits out of the if..else block and the authorization header to be added only in the case token is empty

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -205,8 +206,21 @@ public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMetho
         String organization = githubUrlComponents.org();
         branchOrTagName = githubUrlComponents.branchOrTag();
 
-        String githubToken = ecmService.getAccessToken();
-        Boolean isPrivate = gitHubService.isRepoPrivate(organization, repository, githubToken);
+        String githubToken;
+        Boolean isPrivate;
+
+        try {
```
salonishah11, line 212: nit: since this whole logic is related to determining if the repo is private, I wonder if it can be moved out of Controller and into the Github service layer itself. But if this is more effort and considering this is a time sensitive fix I am happy for it to be punted.

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -844,6 +844,43 @@ void dontStoreMethodDetailsForDockstoreMethod() throws Exception {
     assertNull(githubMethodDetailsDao.getMethodSourceDetails(postMethodResponse.getMethodId()));
   }
 
+  @Test
+  void postMethodForUserWithNoToken() throws Exception {
```
salonishah11, line 848: It might be good to add the error cases also as part of unit testing - especially the ones where GithubService.isPrivateRepo throws an error or EcmService throws an error

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -844,6 +844,43 @@ void dontStoreMethodDetailsForDockstoreMethod() throws Exception {
...
+                "cromwell",
+                "broadinstitute",
+                "this/is/a/path",
+                false,
+                postMethodResponse.getMethodId()));
+
+    assertFalse(
+        githubMethodDetailsDao
+            .getMethodSourceDetails(postMethodResponse.getMethodId())
+            .isPrivate());
```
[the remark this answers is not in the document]
kpierre13: Hm, I see what you mean. This test is actually more of a duplicate of what other tests are doing. I'll remove.

kpierre13:
@cjllanwarne @salonishah11 I'm going to punt the smaller nits like moving things into the service layer and consolidating conditional statements into a follow-up PR for the sake of time. I plan on picking it up right after.

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -205,8 +206,20 @@ public ResponseEntity<PostMethodResponse> postMethod(PostMethodRequest postMetho
...
+          isPrivate = gitHubService.isRepoPrivate(organization, repository, "");
+        } catch (GitHubClient.GitHubClientException e) {
+          githubToken = ecmService.getAccessToken();
+          isPrivate = gitHubService.isRepoPrivate(organization, repository, githubToken);
+        } catch (RestClientException e) {
+          recordMethodCreationCompletion(
+              methodSource, HttpStatus.NOT_FOUND.value(), requestStartNanos);
+          return new ResponseEntity<>(
+              new PostMethodResponse().error(e.getMessage()), HttpStatus.NOT_FOUND);
+        }
```
salonishah11, line 222: It seems we are catching general `RestClientException` here but throwing 404 Not found to the user. We should throw the response code that is returned in the exception

kpierre13: Ah, thank you for catching. I switched to this exception since I see that's what's thrown when `.getAccesToken()` fails. Since this is a general exception, I'm giving it the 500 status, similar to the block of exceptions we have below.

cjllanwarne (approved):

cjllanwarne:
Sonarcloud is saying you have nested try blocks which counts as a major issue... I'm not sure I'd got that far... but it might be worth unpacking that nesting to make it easier to follow

kpierre13:
> Sonarcloud is saying you have nested try blocks which counts as a major issue... I'm not sure I'd got that far... but it might be worth unpacking that nesting to make it easier to follow

I saw that, but I was wondering if it was worth addressing in this ticket, since I plan on doing a follow-up that'll extract this particular code completely. What do you think?

kpierre13:
> Kind of suspicious that sonar things that only 26% of your code is covered despite all the new test cases. Do you know what's going on? Code coverage is far from perfect, but 26% seems super low

I saw that and was a but discouraged as well. Looking through, it seems like it's calling out a lot of code in the controller that I didn't touch, but the bulk of the untested code is in the `GitHubClient`. Since I plan on going back through and making that code less repetitive, I'll also add a test class for that file.

salonishah11 (approved):

PR #268, opened 2024-03-29
[WM-2564] Clean-up ECM verification code

kpierre13:


service/src/main/java/bio/terra/cbas/dependencies/github/GitHubService.java
```diff
@@ -1,18 +1,28 @@
...
+import bio.terra.cbas.dependencies.ecm.EcmService;
 import org.springframework.stereotype.Component;
 
 @Component
 public class GitHubService {
   private final GitHubClient client;
+  private final EcmService ecmService;
 
-  public GitHubService(GitHubClient gitHubClient) {
+  public GitHubService(GitHubClient gitHubClient, EcmService ecmService) {
```
mspector, line 11: my first reaction to this was: is it weird to have a service contain another service?

this might be premature optimization, but I wonder if it makes sense to have a service that is responsible for external workflow repositories. For now that's just github, but maybe having all the logic in the same place will make sense when we have more locations to choose from.

I don't think it's a problem as written, though.

mspector (approved):
one question to think about, but otherwise LGTM

salonishah11 (approved):

cjllanwarne (approved):

PR #271, opened 2024-04-04
[WM-2588] Refactor TestGithubUrlComponents to use DynamicTest

kpierre13:


mspector (approved):
I look forward to hearing more about this!

service/src/test/java/bio/terra/cbas/util/methods/TestGithubUrlComponents.java
```diff
@@ -4,47 +4,41 @@
...
+      assertEquals(extractGithubDetailsFromUrl(url).branchOrTag(), branchOrTag);
+      assertEquals(extractGithubDetailsFromUrl(url).repo(), repo);
     }
+  }
 
-    githubMethods.clear();
+  @TestFactory
+  Stream<DynamicTest> returnExpectedUrlComponents() {
+
+    var testCases =
```
salonishah11, line 26: It would be great to cover a negative test case as well. Do you think you can add it here?

salonishah11 (approved):
One comment about test cases otherwise LGTM

service/src/test/java/bio/terra/cbas/util/methods/TestGithubUrlComponents.java
salonishah11: I suggested the negative test case based on the method's signature that it throws `URISyntaxException` when trying to convert to a URI and since this is a unit test it could cover that scenario as well even though the current way the method is used won't receive any incorrect url.

salonishah11: But if it is non-trivial to add the negative test and considering this was supposed to be a refactor of current test, I am happy with the code as is.

kpierre13: It should be easy enough to do. I'll add in a negative case to at least have.

salonishah11 (approved):

service/src/test/java/bio/terra/cbas/util/methods/TestGithubUrlComponents.java
```diff
@@ -1,50 +1,60 @@
...
-import java.util.Map;
-import org.junit.jupiter.api.Test;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.TestFactory;
 
 class TestGithubUrlComponents {
+  // assertThrows(
+  //            CustomChangeException.class, () ->
+  // backfillGithubDetails.generateStatements(mockDb));
```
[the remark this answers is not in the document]
kpierre13: Yes! forgot to cleanup. Thank you!

PR #272, opened 2024-04-12
[WM-2554] use ProviderStates enum from ECM

mspector:
we can make this minor improvement now that https://github.com/DataBiosphere/terra-external-credentials-manager/pull/189 is merged.


mspector:
this isn't going to be as simple as I expected. closing this for now.

PR #273, opened 2024-04-17
[WM-2554] publish ecm pact, tweak states

mspector:


service/src/test/java/bio/terra/cbas/pact/EcmPactTest.java
```diff
@@ -34,10 +33,9 @@ private void initEcmService(MockServer mockServer) {
     ecmService = new EcmService(ecmClient, new BearerToken("accessToken"));
   }
 
-  @Pact(consumer = "cbas", provider = "ecm")
+  @Pact(consumer = "cbas", provider = "externalcreds")
```
salonishah11, line 36: Out of curiosity, was this changed because Pact broker expects `externalcreds` instead of `ecm`?

salonishah11 (approved):

kpierre13 (approved):

PR #277, opened 2024-04-19
[WM-2597] remove backfillOriginalWorkspaceIdService

mspector:


kpierre13 (approved):

salonishah11 (approved):
@mspector SonarCloud is reporting 2 new issues. It seems 2 fields in `InstanceInitializer` are no longer used. If they are indeed not used can they be removed? Otherwise LGTM.

PR #279, opened 2024-04-24
[WM-2611] use githash during workflow submission

mspector:


service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -99,17 +97,20 @@ private void initSamMocks() {
...
 
-    when(methodDao.getMethods()).thenReturn(List.of(neverRunMethod1, previouslyRunMethod2));
-    when(methodDao.getMethod(neverRunMethod1.methodId())).thenReturn(neverRunMethod1);
-    when(methodDao.getMethod(previouslyRunMethod2.methodId())).thenReturn(previouslyRunMethod2);
-    when(githubMethodDetailsDao.getMethodSourceDetails(previouslyRunMethod2.methodId()))
-        .thenReturn(githubMethodDetails);
-    when(githubMethodDetailsDao.getMethodSourceDetails(neverRunMethod1.methodId()))
-        .thenReturn(githubMethodDetails);
+    Method neverRunMethod1WithGithub = neverRunMethod1.withGithubMethodDetails(githubMethodDetails);
+    Method previouslyRunMethod2WithGithub =
```
cjllanwarne, line 101: Not necessarily something to fix in this PR... this amount of mocking in the API just to check whether a single function gets called in a particular way feels like a code smell.

I don't think it's a new through that perhaps our controller methods would benefit from being divided with much of the logic pushed down into smaller functions in the services layer!

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -859,7 +862,7 @@ void dontStoreMethodDetailsForDockstoreMethod() throws Exception {
         objectMapper.readValue(
             response.getResponse().getContentAsString(), PostMethodResponse.class);
 
-    assertNull(githubMethodDetailsDao.getMethodSourceDetails(postMethodResponse.getMethodId()));
+    // assertNull(githubMethodDetailsDao.getMethodSourceDetails(postMethodResponse.getMethodId()));
```
cjllanwarne, line 865: Commented code

service/src/test/java/bio/terra/cbas/controllers/TestRunSetsApiController.java
```diff
@@ -121,10 +123,24 @@ class TestRunSetsApiController {
...
+          .formatted(gitHubOrganization, gitHubRepository, gitHubBranchOrTag, gitHubPath);
+
+  private final String gitHubWorkflowOriginalUrl =
+      "https://github.com/%s/%s/%s/%s"
+          .formatted(gitHubOrganization, gitHubRepository, gitHubBranchOrTag, gitHubPath);
+  private final GithubMethodDetails githubMethodDetails =
+      new GithubMethodDetails(gitHubRepository, gitHubOrganization, gitHubPath, false, methodId);
   private final Boolean isCallCachingEnabled = false;
+
+  // TODO: this variable says dockstore but is actually github... bad?
```
cjllanwarne, line 141: Still TODO?

cjllanwarne, line 141: FWIW it should be perfectly ok for a dockstore method to resolve to a URL in github. Probably ought to be to a raw github url though 🤔

service/src/test/java/bio/terra/cbas/dao/TestGithubMethodDetailsDao.java
```diff
@@ -1,55 +1,7 @@
...
-  @Test
-  void retrievesGithubMethodDetails() {
-    GithubMethodDetails methodDetails = githubMethodDetailsDao.getMethodSourceDetails(methodId1);
-    assertEquals(methodId1, methodDetails.methodId());
-    assertEquals("cromwell", methodDetails.repository());
-    assertEquals("cbas/dao/test.py", methodDetails.path());
-    assertEquals("broadinstitute", methodDetails.organization());
-    assertEquals(true, methodDetails.isPrivate());
-  }
+  // TODO: Migrate to the TestMethodDao suite
```
cjllanwarne, line 6: Still TODO?

cjllanwarne:
We should make sure to check:

- Newly imported workflows use the new commit URL
- Previously imported workflows continue to use the branch URL
- The UI pages for old and new workflows represent both correctly

We probably want a ticket for:
- Adding the branch information (and actually, all of the github details more generally) to the workflow configuration page
  - This is probably an important prerequisite for allowing users to trigger updates when their branch has moved on

service/src/main/java/bio/terra/cbas/common/MethodUtil.java
```diff
@@ -35,14 +41,48 @@ public static String convertGithubToRawUrl(String originalUrl)
   public static String convertToRawUrl(
```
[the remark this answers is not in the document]
mspector: The reason I separated these is because, in `postMethod`, it seems like we need to pass the method URL to cromwell's `describe` endpoint before we actually create the `methodVersion` object (and the associated `githubMethodDetails` object). 

Right now, `getRawUrl` requires an existing `methodVersion`. I'm not saying the code can't be refactored to accomplish what you're asking for, but at the time, it wasn't obvious to me how to do it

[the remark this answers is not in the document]
mspector: what makes you say so? I see this method as being a thing that takes an original URL, and turns it into ("converts it") into a raw URL

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -221,20 +223,15 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+    String storedRawMethodUrl;
     try {
-      PostMethodRequest.MethodSourceEnum methodSourceEnum =
-          convertToMethodSourceEnum(methodVersion.method().methodSource());
-
-      rawMethodUrl =
-          MethodUtil.convertToRawUrl(
-              methodVersion.url(), methodSourceEnum, methodVersion.name(), dockstoreService);
+      storedRawMethodUrl =
+          MethodUtil.getRawUrl(methodVersion, gitHubService, dockstoreService, userToken);
```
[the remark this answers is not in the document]
mspector: Yeah, I see what you're saying. based on this and your other comments I suspect I need to take a closer look at the flow.

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
```diff
@@ -28,19 +29,37 @@ public MethodDao(NamedParameterJdbcTemplate jdbcTemplate) {
...
 
   public int createMethod(Method method) {
+    method
+        .githubMethodDetails()
+        .ifPresent(
+            details ->
+                jdbcTemplate.update(
+                    "insert into github_method_details (repository, organization, path, private, method_id) "
+                        + "values (:repository, :organization, :path, :isPrivate, :methodId)",
+                    new BeanPropertySqlParameterSource(details)));
```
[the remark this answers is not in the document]
mspector: huh, that's interesting. what you're saying makes sense, but our unit tests don't complain... I'm not sure I can explain it myself theoretically

service/src/main/java/bio/terra/cbas/dependencies/github/GitHubService.java
```diff
@@ -40,4 +40,17 @@ public String getCurrentGithash(
 
     return commitInfo.getSha();
   }
+
+  public String getBaseUrl(String organization, String repo, String branch, BearerToken userToken)
```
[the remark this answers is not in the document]
mspector: it does make sense, but would require some other refactoring. right now the "getRawUrl" logic handles both dockstore and github methods

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
mspector: Yeah, I agree... maybe I can find some opportunity here to chip away at this controller layer division...

service/src/test/java/bio/terra/cbas/dao/TestGithubMethodDetailsDao.java
mspector: yes... sorry, I should be more thorough

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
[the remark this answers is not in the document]
cjllanwarne: In fact, this is exactly what happens. I'll add a test case and update the method.

service/src/main/java/bio/terra/cbas/dependencies/github/GitHubService.java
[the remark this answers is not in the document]
cjllanwarne: In fact this method is unused so can be deleted

service/src/main/java/bio/terra/cbas/models/Method.java
```diff
@@ -24,4 +26,16 @@ public record Method(
...
+  public Method withGithubMethodDetails(GithubMethodDetails githubMethodDetails) {
+    return new Method(
+        methodId,
+        name,
+        description,
+        created,
+        lastRunSetId,
+        methodSource,
+        originalWorkspaceId,
+        Optional.of(githubMethodDetails));
```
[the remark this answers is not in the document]
cjllanwarne: Done

service/src/test/java/bio/terra/cbas/dao/TestGithubMethodDetailsDao.java
cjllanwarne: There is indeed now a test in TestMethodDao validating that GithubMethodDetails are saved and returned by the MethodDao

service/src/test/java/bio/terra/cbas/controllers/TestMethodsApiController.java
```diff
@@ -831,6 +831,9 @@ void returnSpecificMethodWithDetails() throws Exception {
     assertEquals(false, actualResponseForMethod2.isIsPrivate());
   }
 
+  // TODO: A MethodDao test to verify dockstore methods don't get github details entries.
```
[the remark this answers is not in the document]
cjllanwarne: Added to `TestMethodsApiController` because that is where the logic to send or not send github details currently lives.

cjllanwarne:
To check:
- That workflows still submit properly
- That dockstore is still stored/submitted properly

service/src/main/java/bio/terra/cbas/dependencies/github/GitHubService.java
```diff
@@ -41,17 +57,50 @@ public String getCurrentGithash(
...
+          methodDetails.path());
+    }
+  }
+
+  public static String buildRawGithubUrl(String org, String repo, String githash, String path) {
+    return "https://%s/%s/%s/%s/%s".formatted(RAW_GITHUB_URL_HOST, org, repo, githash, path);
+  }
+
+  // helper method to verify that URL is valid and its host is supported
+  public static String validateGithubUrl(String methodUrl) {
```
mspector, line 88: when I look at how this method is used, it makes sense, but its name makes me think it should return a boolean and throw exceptions instead of returning a string or null. at the risk of nit-picking, maybe this should be called something like `generateGithubUrlErrorMessage`?

mspector (commented):
I can't formally approve this because I'm a co-author, but unofficially it LGTM

cjllanwarne (approved):
Michael's "official" review thumb

service/src/main/java/bio/terra/cbas/dao/MethodDao.java
```diff
@@ -28,29 +29,47 @@ public MethodDao(NamedParameterJdbcTemplate jdbcTemplate) {
     this.jdbcTemplate = jdbcTemplate;
   }
 
+  public static final String METHOD_JOIN_GITHUB_METHOD_DETAILS =
```
kpierre13, line 32: Will this cause errors when trying to fetch Dockstore methods?

cjllanwarne: Dockstore methods will not have a github method details so would look like any other pre-backfilling method. We just need to be aware that the values might be null and fill in an Optional.empty value for the details. There are some test cases for this too, which test this out for real in containerized tests.

PR #281, opened 2024-05-06
[WM-2612] Async submissions using Spring annotation

salonishah11:
This PR refactors the logic for submitting workflows by introducing async method using Spring annotation `@EnableAsync`. In this iteration, the timing heavy API calls to WDS and Cromwell is extracted from the main thread and moved into async method. 

The **current** steps that CBAS performs when it receives a POST /run_sets request:
- Checks that user has permission to submit workflows
- Validates the request
- Serially gets the data from WDS for N records
- Gets method data, generates raw URL for workflow, creates submission record in database and updates method and method version tables
- Gets user ID from SAM to store along with submission details
- Builds workflow options JSON to be used when submitting to Cromwell
- Splits the N workflows in batches of 100 to submit to Cromwell
- For each batch, 
   - generates the inputs for each workflow in the batch
   - Submits workflows in batch using Cromwell’s POST /batch endpoint
- Records status and errors if any to database
- Returns response

The **proposed** steps CBAS will perform when it receives a POST /run_sets request to do async submissions:
- Checks that user has permission to submit workflows
- Validates the request
- Gets user ID from SAM to store along with submission details
- Gets method data, generates raw URL for workflow
- Register RunSet in database and updates method and method version tables
- Create mapping between Record ID and Run ID
- Register Runs in the RunSet in database
- Record status and errors in any
- Trigger async submission method  `triggerWorkflowSubmission` for the RunSet request
- Return response

In the async method `triggerWorkflowSubmission` that is defined in `RunSetHelper` class, it will
- Serially gets the data from WDS for N records
- Builds workflow options JSON to be used when submitting to Cromwell
- Splits the N workflows in batches of 100 to submit to Cromwell
- For each batch, 
   - generates the inputs for each workflow in the batch
   - Submits workflows in batch using Cromwell’s POST /batch endpoint
- Update database as needed

Closes https://broadworkbench.atlassian.net/browse/WM-2612

### DIFF
--- a/service/src/main/java/bio/terra/cbas/config/AsyncConfig.java
+++ b/service/src/main/java/bio/terra/cbas/config/AsyncConfig.java
@@ -1,17 +1,21 @@
 package bio.terra.cbas.config;
 
+import bio.terra.cbas.controllers.util.AsyncExceptionHandler;
 import java.util.concurrent.ThreadPoolExecutor;
+import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.AsyncConfigurer;
 import org.springframework.scheduling.annotation.EnableAsync;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 
 @Configuration
 @EnableAsync
-public class AsyncConfig {
+public class AsyncConfig implements AsyncConfigurer {
 
   @Bean
-  public ThreadPoolTaskExecutor runSetExecutor() {
+  @Override
+  public ThreadPoolTaskExecutor getAsyncExecutor() {
     ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
     executor.setCorePoolSize(5);
     executor.setMaxPoolSize(10);
@@ -20,5 +24,10 @@ public class AsyncConfig {
     executor.setRejectedExecutionHandler(new ThreadPoolExecutor.CallerRunsPolicy());
     executor.initialize();
     return executor;
+  }
+
+  @Override
+  public AsyncUncaughtExceptionHandler getAsyncUncaughtExceptionHandler() {
+    return new AsyncExceptionHandler();
   }
 }
--- a/service/src/main/java/bio/terra/cbas/config/AsyncConfig.java
+++ b/service/src/main/java/bio/terra/cbas/config/AsyncConfig.java
@@ -1,0 +1,24 @@
+package bio.terra.cbas.config;
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.EnableAsync;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+import java.util.concurrent.ThreadPoolExecutor;
+
+@Configuration
+@EnableAsync
+public class AsyncConfig {
+
+  @Bean
+  public ThreadPoolTaskExecutor runSetExecutor() {
+    ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+    executor.setCorePoolSize(5);
+    executor.setMaxPoolSize(10);
+    executor.setQueueCapacity(500);
+    executor.setThreadNamePrefix("RunSetExecutor-");
+    executor.setRejectedExecutionHandler(new ThreadPoolExecutor.CallerRunsPolicy());
+    executor.initialize();
+    return executor;
+  }
+}
--- a/service/src/main/java/bio/terra/cbas/config/AsyncConfig.java
+++ b/service/src/main/java/bio/terra/cbas/config/AsyncConfig.java
@@ -3,23 +3,34 @@ package bio.terra.cbas.config;
 import bio.terra.cbas.controllers.util.AsyncExceptionHandler;
 import java.util.concurrent.ThreadPoolExecutor;
 import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.bind.ConstructorBinding;
 import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.Configuration;
 import org.springframework.scheduling.annotation.AsyncConfigurer;
 import org.springframework.scheduling.annotation.EnableAsync;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 
-@Configuration
 @EnableAsync
+@ConfigurationProperties(prefix = "cbas.async.submission")
 public class AsyncConfig implements AsyncConfigurer {
+  private final int coreThreadPoolSize;
+  private final int maxThreadPoolSize;
+  private final int queueCapacity;
+
+  @ConstructorBinding
+  public AsyncConfig(int coreThreadPoolSize, int maxThreadPoolSize, int queueCapacity) {
+    this.coreThreadPoolSize = coreThreadPoolSize;
+    this.maxThreadPoolSize = maxThreadPoolSize;
+    this.queueCapacity = queueCapacity;
+  }
 
   @Bean
   @Override
   public ThreadPoolTaskExecutor getAsyncExecutor() {
     ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
-    executor.setCorePoolSize(5);
-    executor.setMaxPoolSize(10);
-    executor.setQueueCapacity(500);
+    executor.setCorePoolSize(coreThreadPoolSize);
+    executor.setMaxPoolSize(maxThreadPoolSize);
+    executor.setQueueCapacity(queueCapacity);
     executor.setThreadNamePrefix("RunSetExecutor-");
     executor.setRejectedExecutionHandler(new ThreadPoolExecutor.CallerRunsPolicy());
     executor.initialize();
--- a/service/src/main/java/bio/terra/cbas/config/AsyncConfig.java
+++ b/service/src/main/java/bio/terra/cbas/config/AsyncConfig.java
@@ -1,10 +1,10 @@
 package bio.terra.cbas.config;
 
+import java.util.concurrent.ThreadPoolExecutor;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.scheduling.annotation.EnableAsync;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
-import java.util.concurrent.ThreadPoolExecutor;
 
 @Configuration
 @EnableAsync
--- a/service/src/main/java/bio/terra/cbas/config/AsyncConfiguration.java
+++ b/service/src/main/java/bio/terra/cbas/config/AsyncConfiguration.java
@@ -7,7 +7,6 @@ import java.util.concurrent.ThreadPoolExecutor;
 import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.ConfigurationProperties;
-import org.springframework.boot.context.properties.bind.ConstructorBinding;
 import org.springframework.context.annotation.Bean;
 import org.springframework.scheduling.annotation.AsyncConfigurer;
 import org.springframework.scheduling.annotation.EnableAsync;
@@ -20,21 +19,13 @@ public class AsyncConfiguration implements AsyncConfigurer {
   private final int maxThreadPoolSize;
   private final int queueCapacity;
 
-  @Autowired private final RunDao runDao;
-  @Autowired private final RunSetDao runSetDao;
+  @Autowired private RunDao runDao;
+  @Autowired private RunSetDao runSetDao;
 
-  @ConstructorBinding
-  public AsyncConfiguration(
-      int coreThreadPoolSize,
-      int maxThreadPoolSize,
-      int queueCapacity,
-      RunDao runDao,
-      RunSetDao runSetDao) {
+  public AsyncConfiguration(int coreThreadPoolSize, int maxThreadPoolSize, int queueCapacity) {
     this.coreThreadPoolSize = coreThreadPoolSize;
     this.maxThreadPoolSize = maxThreadPoolSize;
     this.queueCapacity = queueCapacity;
-    this.runDao = runDao;
-    this.runSetDao = runSetDao;
   }
 
   @Bean
--- a/service/src/main/java/bio/terra/cbas/config/AsyncConfiguration.java
+++ b/service/src/main/java/bio/terra/cbas/config/AsyncConfiguration.java
@@ -1,8 +1,11 @@
 package bio.terra.cbas.config;
 
 import bio.terra.cbas.controllers.util.AsyncExceptionHandler;
+import bio.terra.cbas.dao.RunDao;
+import bio.terra.cbas.dao.RunSetDao;
 import java.util.concurrent.ThreadPoolExecutor;
 import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.bind.ConstructorBinding;
 import org.springframework.context.annotation.Bean;
@@ -12,16 +15,26 @@ import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 
 @EnableAsync
 @ConfigurationProperties(prefix = "cbas.async.submission")
-public class AsyncConfig implements AsyncConfigurer {
+public class AsyncConfiguration implements AsyncConfigurer {
   private final int coreThreadPoolSize;
   private final int maxThreadPoolSize;
   private final int queueCapacity;
 
+  @Autowired private final RunDao runDao;
+  @Autowired private final RunSetDao runSetDao;
+
   @ConstructorBinding
-  public AsyncConfig(int coreThreadPoolSize, int maxThreadPoolSize, int queueCapacity) {
+  public AsyncConfiguration(
+      int coreThreadPoolSize,
+      int maxThreadPoolSize,
+      int queueCapacity,
+      RunDao runDao,
+      RunSetDao runSetDao) {
     this.coreThreadPoolSize = coreThreadPoolSize;
     this.maxThreadPoolSize = maxThreadPoolSize;
     this.queueCapacity = queueCapacity;
+    this.runDao = runDao;
+    this.runSetDao = runSetDao;
   }
 
   @Bean
@@ -39,6 +52,6 @@ public class AsyncConfig implements AsyncConfigurer {
 
   @Override
   public AsyncUncaughtExceptionHandler getAsyncUncaughtExceptionHandler() {
-    return new AsyncExceptionHandler();
+    return new AsyncExceptionHandler(runDao, runSetDao);
   }
 }
--- a/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
@@ -320,7 +320,8 @@ public class RunSetsApiController implements RunSetsApi {
       String errorMessage =
           "Run Set can't be aborted when it is Queued state as system might still be processing the request.";
       return new ResponseEntity<>(
-          new AbortRunSetResponse().runSetId(runSetId).errors(errorMessage), HttpStatus.OK);
+          new AbortRunSetResponse().runSetId(runSetId).errors(errorMessage),
+          HttpStatus.BAD_REQUEST);
     }
 
     AbortRunSetResponse aborted = new AbortRunSetResponse();
--- a/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
+++ b/service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
@@ -25,8 +25,6 @@ import bio.terra.cbas.dao.RunDao;
 import bio.terra.cbas.dao.RunSetDao;
 import bio.terra.cbas.dependencies.dockstore.DockstoreService;
 import bio.terra.cbas.dependencies.sam.SamService;
-import bio.terra.cbas.dependencies.wds.WdsService;
-import bio.terra.cbas.dependencies.wes.CromwellService;
 import bio.terra.cbas.model.AbortRunSetResponse;
 import bio.terra.cbas.model.OutputDestination;
 import bio.terra.cbas.model.PostMethodRequest;
@@ -45,7 +43,6 @@ import bio.terra.cbas.monitoring.TimeLimitedUpdater;
 import bio.terra.cbas.runsets.monitoring.RunSetAbortManager;
 import bio.terra.cbas.runsets.monitoring.RunSetAbortManager.AbortRequestDetails;
 import bio.terra.cbas.runsets.monitoring.SmartRunSetsPoller;
-import bio.terra.cbas.util.UuidSource;
 import bio.terra.common.iam.BearerToken;
 import bio.terra.common.iam.BearerTokenFactory;
 import com.fasterxml.jackson.core.JsonProcessingException;
@@ -85,8 +82,6 @@ public class RunSetsApiController implements RunSetsApi {
 
   public RunSetsApiController(
       SamService samService,
-      CromwellService cromwellService,
-      WdsService wdsService,
       DockstoreService dockstoreService,
       ObjectMapper objectMapper,
       MethodDao methodDao,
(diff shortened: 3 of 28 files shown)


service/src/test/java/bio/terra/cbas/dependencies/sam/TestSamService.java
```diff
@@ -44,7 +44,6 @@ class TestSamService {
   private final ResourcesApi resourcesApi = mock(ResourcesApi.class);
   @SpyBean private final SamService samService = spy(new SamService(samClient));
 
-  // TODO: maybe change these Strings to BearerToken
```
salonishah11, line 47: This is a cleanup from previous PR.

cjllanwarne:
I see "Verify consumer pacts / verify-consumer-pact (pull_request) Failing after 2m" having errors. Presumably because the status we are returning now is different? Do we need to update the expectations?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -333,11 +294,29 @@ public ResponseEntity<AbortRunSetResponse> abortRunSet(UUID runSetId) {
...
+    // Cromwell got cancelled but some Runs that were still in process of being submitted do in
+    // fact get submitted to Cromwell and are started by Cromwell. And hence the RunSet might have
+    // some aborted Runs and some Runs in Running state.
+    if (runSet.status() == CbasRunSetStatus.QUEUED) {
+      String errorMessage =
+          "Run Set can't be aborted when it is Queued state as system might still be processing the request.";
+      return new ResponseEntity<>(
+          new AbortRunSetResponse().runSetId(runSetId).errors(errorMessage),
+          HttpStatus.BAD_REQUEST);
+    }
```
salonishah11, line 313: This changes when a RunSet can be aborted. Since the submission happens in a different thread it is difficult to know how many Runs have been submitted to Cromwell and how many haven't. Hence the proposal here is to not allow abort requests when RunSet is in Queued state. I think when we move to a better async submissions design where CBAS has queues or scheduler to track the work it might be possible abort RunSet even in Queued state but it would be difficult in this iteration.

service/src/main/java/bio/terra/cbas/controllers/util/RunSetsHelper.java
```diff
@@ -0,0 +1,286 @@
...
+    this.uuidSource = uuidSource;
+  }
+
+  private record WdsRecordResponseDetails(
+      ArrayList<RecordResponse> recordResponseList, Map<String, String> recordIdsWithError) {}
+
+  private record RunAndRecordDetails(UUID runId, RecordResponse recordResponse) {}
+
+  @Async
+  public void triggerWorkflowSubmission(
```
salonishah11, line 80: Async method that will fetch records data from WDS and submit to Cromwell in a different thread as allocated by `getAsyncExecutor` from `AsyncConfig`.

service/src/test/java/bio/terra/cbas/controllers/util/TestRunSetHelper.java
```diff
@@ -0,0 +1,373 @@
...
+    when(wdsService.getRecord(eq(recordType), eq(recordId2), any()))
+        .thenThrow(
+            new WdsServiceApiException(
+                new org.databiosphere.workspacedata.client.ApiException(
+                    400, "ApiException thrown for testing purposes.")));
+
+    when(runDao.getRuns(new RunDao.RunsFilters(runSetId, List.of(QUEUED))))
+        .thenReturn(List.of(run1, run2));
+
+    mockRunSetHelper.triggerWorkflowSubmission(
```
cjllanwarne, line 257: Presumably in the context of this test case, this is _not_ running async?

service/src/test/java/bio/terra/cbas/controllers/util/TestRunSetHelper.java
```diff
@@ -0,0 +1,373 @@
...
+        runSetRequest, runSet, recordIdToRunIdMapping, mockToken, mockWorkflowUrl);
+
+    // verify Runs were set to Error state
+    verify(runDao)
+        .updateRunStatusWithError(
+            eq(runId1),
+            eq(CbasRunStatus.SYSTEM_ERROR),
+            any(),
+            eq(
+                "Input generation failed for record MY_RECORD_ID_1. Coercion error: Coercion from Integer to String failed for parameter myworkflow.mycall.inputname2. Coercion not supported between these types."));
```
cjllanwarne, line 317: Is this text specific to the `RunSetHelper`? Maybe we could set up one of the mock function calls to throw a random error and make sure the trigger function sends it to the DB properly? 

Though, I do know this class needs a refactor to break the "trigger" into many sub-functions, is this specific request too much to unpick right now?

service/src/test/java/bio/terra/cbas/pact/VerifyPactsAllControllers.java
```diff
@@ -204,7 +206,7 @@ public void initializeDAO() throws Exception {
     when(runSetDao.createRunSet(any())).thenReturn(1);
     when(methodDao.updateLastRunWithRunSet(any())).thenReturn(1);
     when(methodVersionDao.updateLastRunWithRunSet(any())).thenReturn(1);
-    when(runSetDao.updateStateAndRunDetails(any(), any(), any(), any(), any())).thenReturn(1);
+    when(runSetDao.updateStateAndRunSetDetails(any(), any(), any(), any(), any())).thenReturn(1);
```
cjllanwarne, line 209: TOL: I wonder whether we should move this into the `RunSetsHelper`/`RunSetsService` layer so that the controller doesn't have to interact with the DAO directly 🤔.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -256,67 +221,64 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+                  run.runId(), CbasRunStatus.SYSTEM_ERROR, DateUtils.currentTimeInUTC(), errorMsg));
+
+      // mark RunSet in Error state
+      int runsCount = runsInRunSet.size();
+      runSetDao.updateStateAndRunSetDetails(
+          runSet.runSetId(), CbasRunSetStatus.ERROR, runsCount, runsCount, OffsetDateTime.now());
+
+      return new ResponseEntity<>(
+          new RunSetStateResponse()
+              .errors("Failed to register submission request. Error(s): " + e.getMessage()),
```
cjllanwarne, line 267: TOL: This might be a good candidate to extract into a `registerRuns` function in the `RunSetsHelper`/`RunSetsService` layer 🤔

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -333,11 +295,29 @@ public ResponseEntity<AbortRunSetResponse> abortRunSet(UUID runSetId) {
...
+    // If RunSet is in Queued state, it means that CBAS is still processing the POST request in the
+    // async method. In this case, if the RunSet is set to Cancelling, it might leave the RunSet in
+    // weird state as it's possible that the background thread could still be processing Runs
+    // when the abort request came in and as a result some Runs that had been submitted to
+    // Cromwell got cancelled but some Runs that were still in process of being submitted do in
+    // fact get submitted to Cromwell and are started by Cromwell. And hence the RunSet might have
+    // Runs both in Aborted and Running state. Instead, don't abort RunSet when in Queued state.
+    if (runSet.status() == CbasRunSetStatus.QUEUED) {
+      String errorMessage =
+          "Run Set can't be aborted when it is Queued state as system might still be processing the request.";
```
cjllanwarne, line 310: I'd leave out the why. I think just "Run Set can't be aborted when it is Queued state" is fine

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -424,187 +404,68 @@ public static List<String> validateRequestInputsAndOutputs(
...
-                runId,
-                externalId,
-                runSet,
-                recordId,
-                DateUtils.currentTimeInUTC(),
-                runState,
-                DateUtils.currentTimeInUTC(),
-                DateUtils.currentTimeInUTC(),
-                errors));
+  private RunSet registerRunSet(
```
cjllanwarne, line 407: TOL this might be a good candidate of business logic to move to the `RunSetService` / `RunSetHandler` layer

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -424,187 +404,68 @@ public static List<String> validateRequestInputsAndOutputs(
...
-                    requestedId ->
-                        storeRun(
-                            requestedIdToRunId.get(requestedId),
-                            null,
-                            runSet,
-                            requestedIdToRecord.get(requestedId).getId(),
-                            SYSTEM_ERROR,
-                            errorMsg + e.getMessage()))
-                .toList());
+  private List<RunStateResponse> registerRunsInRunSet(
```
cjllanwarne, line 444: TOL this might be a good candidate of business logic to move to the RunSetService / RunSetHandler layer

cjllanwarne (commented):
- We need to update the UI to be "Queued" aware on the details pages
- We need to update the UI to be "Queued" aware for the abort button

service/src/main/java/bio/terra/cbas/controllers/util/AsyncExceptionHandler.java
```diff
@@ -0,0 +1,21 @@
...
+public class AsyncExceptionHandler implements AsyncUncaughtExceptionHandler {
+
+  private final Logger logger = LoggerFactory.getLogger(AsyncExceptionHandler.class);
+
+  @Override
+  public void handleUncaughtException(Throwable ex, Method method, Object... params) {
+    // handle exception
+    logger.error(
+        "Exception thrown in Thread '{}' while executing method '{}'. Error message: {}",
+        Thread.currentThread().getName(),
```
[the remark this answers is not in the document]
salonishah11: I think it's the thread that was running the task. This is what I see when I try to simulate an exception being thrown in `triggerWorkflowSubmission`:
```
... b.t.c.c.util.AsyncExceptionHandler       : Exception thrown in Thread 'RunSetExecutor-2' while executing method 'triggerWorkflowSubmission'. Error message ...
```

service/src/main/java/bio/terra/cbas/controllers/util/AsyncExceptionHandler.java
```diff
@@ -0,0 +1,21 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
+
+public class AsyncExceptionHandler implements AsyncUncaughtExceptionHandler {
+
+  private final Logger logger = LoggerFactory.getLogger(AsyncExceptionHandler.class);
+
+  @Override
+  public void handleUncaughtException(Throwable ex, Method method, Object... params) {
```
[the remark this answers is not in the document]
salonishah11: `params` contain the parameters that were passed to the async method, here `triggerWorkflowSubmission`. I had attempted to extract RunSet ID from the `params` but it wasn't trivial. I can try again.

>And should we be updating the database with failure states when a run creation has an error?

I had thought about that, but I wasn't sure how to have `RunSetDao` and `RunDao` references in the `AsyncExceptionHandler` class. One way would be to have RunSetApiController itself pass runSetDao and runDao reference to triggerWorkflowSubmission so that it can be extracted from params but I am not sure if that is the right approach.

service/src/main/java/bio/terra/cbas/config/AsyncConfig.java
```diff
@@ -0,0 +1,33 @@
...
+@EnableAsync
+public class AsyncConfig implements AsyncConfigurer {
+
+  @Bean
+  @Override
+  public ThreadPoolTaskExecutor getAsyncExecutor() {
+    ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+    executor.setCorePoolSize(5);
+    executor.setMaxPoolSize(10);
+    executor.setQueueCapacity(500);
```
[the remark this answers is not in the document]
salonishah11: Yes it should be possible. I can add these to application.yml

service/src/main/java/bio/terra/cbas/runsets/monitoring/RunSetAbortManager.java
```diff
@@ -29,18 +29,15 @@ public RunSetAbortManager(RunSetDao runSetDao, RunDao runDao, CromwellService cr
     this.cromwellService = cromwellService;
   }
 
-  public AbortRequestDetails abortRunSet(UUID runSetId, BearerToken userToken) {
+  public AbortRequestDetails abortRunSet(RunSet runSet, BearerToken userToken) {
```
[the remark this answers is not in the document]
salonishah11: We can refactor this, but I would suggest to do that in a separate PR so as to not increase the scope of this PR.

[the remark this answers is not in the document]
salonishah11: The reason this was updated was because now in the `abortRunset` handler in `RunSetsApiController` it [fetches the run set details from the database](https://github.com/DataBiosphere/cbas/pull/281/files#diff-1bd3bc7318e38eee77c2fce2eff273155ad4195dcb9bed577968795ee1baac6fR298-R299) to check if the run set is in Queued state. As a result, I updated this method since we already have the RunSet object and it doesn't need be fetched again on [L37](https://github.com/DataBiosphere/cbas/pull/281/files#diff-8f3866ed117fc606d74801b93514dd66795f764a6570ed289f7c768e3a66c1ffL37-L39).

service/src/test/java/bio/terra/cbas/controllers/util/TestRunSetHelper.java
salonishah11: Yes that is correct since this is just testing the method directly.

service/src/test/java/bio/terra/cbas/controllers/util/TestRunSetHelper.java
salonishah11: No this error is not specific to `RunSetsHelper`. The test case here is testing that correct DAO methods are called when there is an error during inputs generation while submitting workflows.

service/src/test/java/bio/terra/cbas/pact/VerifyPactsAllControllers.java
salonishah11: This should no longer be needed since it is now moved inside `RunSetsService`.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
salonishah11: Yes it indeed is. I have moved this inside `RunSetsService` in the spirit of making controllers do minimum business logic.

service/src/main/java/bio/terra/cbas/controllers/util/AsyncExceptionHandler.java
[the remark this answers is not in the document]
mspector: > One way would be to have RunSetApiController itself pass runSetDao and runDao reference to triggerWorkflowSubmission so that it can be extracted from params but I am not sure if that is the right approach.

I don't have a better suggestion - do you have a specific concern about doing it this way?

(FWIW, it's not obvious to me why you need the references to extract them from params. I'm sure the devil is in the details...)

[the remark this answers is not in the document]
salonishah11: Another way would be to make AsyncExceptionHandler a component and have RunSetDao and RunDao references be auto-wired 🤔

mspector (approved):
👍 we talked about the Pact verification code and agreed it doesn't make sense to change it in this PR. what you have here looks good for now

service/src/main/java/bio/terra/cbas/controllers/util/AsyncExceptionHandler.java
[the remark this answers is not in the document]
salonishah11: In [this commit](https://github.com/DataBiosphere/cbas/pull/281/commits/6ef01a8d2a42f212184331380022f159b3f067aa) I extracted the Run Set ID that is available in params and added it to the log message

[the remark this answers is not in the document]
salonishah11: > One way would be to have RunSetApiController itself pass runSetDao and runDao reference to triggerWorkflowSubmission so that it can be extracted from params but I am not sure if that is the right approach.

I actually take back this suggestion. Apart from `triggerWorkflowSubmission` there are other methods in the RunSetService which do use RunSetDao and RunDao. Implementing this suggestion would mean each method needs to have these references passed in. This would also mean that RunSetsApiController needs to have RunDao as its member again (this was removed in this PR as some DAO logic was moved into RunSetService)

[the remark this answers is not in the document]
cjllanwarne: It feels weird to me to add DAOs to method signatures just so that an exception handler can extract them again from its `params` object.

It feels must more natural to me for the exception handler to have references to the DAO if part of exception handling is updating database tables.

service/src/main/java/bio/terra/cbas/config/AsyncConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+
+@EnableAsync
+@ConfigurationProperties(prefix = "cbas.async.submission")
+public class AsyncConfiguration implements AsyncConfigurer {
+  private final int coreThreadPoolSize;
+  private final int maxThreadPoolSize;
+  private final int queueCapacity;
+
+  @Autowired private final RunDao runDao;
+  @Autowired private final RunSetDao runSetDao;
```
salonishah11, line 24: Question for reviewers: Is this the right way to auto-wired DAO references in a class that also has annotation `ConfigurationProperties`? I had tried adding `Component` annotation and it gave below error. I tried removing `ConstructorBinding` but it still gave the same error. We already have `ConfigurationPropertiesScan` enabled in [App.java](https://github.com/DataBiosphere/cbas/blob/c328335c187693607a408d63faf791aee16305cf/service/src/main/java/bio/terra/cbas/App.java#L34)
```
Description:
AsyncConfiguration is annotated with @ConstructorBinding but it is defined as a regular bean which caused dependency injection to fail.

Action:
Update your configuration so that AsyncConfiguration is defined via @ConfigurationPropertiesScan or @EnableConfigurationProperties.
```

salonishah11 (commented):
PR is ready for re-review

salonishah11:
The 2 SonarCloud issues are related to replacing `instanceof check and cast with 'instanceof T t'` so here replace
```
if (o instanceof RunSetRequest) {
  runSetRequest = (RunSetRequest) o;
}
``` 
with 
```
if (o instanceof RunSetRequest runSetRequest) {
  // not sure what to put here then
}
```
but in this case I wan't to cast the object and assign it to variable that is already defined above. One way to do this is like this but I am not sure if that's the best way.
```
if (o instanceof RunSetRequest requestObj) {
  runSetRequest = requestObj;
}
```

service/src/main/java/bio/terra/cbas/controllers/util/AsyncExceptionHandler.java
```diff
@@ -1,21 +1,104 @@
...
+            Thread.currentThread().getName(), methodName, ex.getMessage());
+
+    // extract method request parameters
+    for (Object o : params) {
+      if (o instanceof RunSetRequest) {
+        runSetRequest = (RunSetRequest) o;
+      }
+
+      if (o instanceof RunSet) {
+        runSet = (RunSet) o;
```
cjllanwarne, line 62: Since I couldn't reply to your top-level comment in a thread, I'll reply here...

Yes I think you want to do something like 
```java
if (o instanceof RunSet runSet) {
  // ... use 'runSet' as a variable
}
```

It looks like maybe you're running into problems because you're then trying to say something like "if there were no runSets found, throw a different exception" below. You could have a flag variable like `boolean runSetFound` and set that to true if one of the params matched?

But presumably the params are always in the same order, so you should be able to predict which is the right one to key into? Or is there like a `paramName` map that also gets passed in somehow?

service/src/test/java/bio/terra/cbas/service/TestRunSetsService.java
```diff
@@ -43,29 +57,33 @@
...
   private final UuidSource uuidSource = mock(UuidSource.class);
-
-  private final RunSetsHelper mockRunSetHelper =
-      new RunSetsHelper(
-          runDao, runSetDao, cromwellService, wdsService, cbasApiConfiguration, uuidSource);
+  private final ObjectMapper objectMapper = mock(ObjectMapper.class);
+  private final CbasContextConfiguration cbasContextConfiguration =
+      mock(CbasContextConfiguration.class);
+
+  private final RunSetsService mockRunSetsService =
```
cjllanwarne, line 75: It's probably better to have a BeforeEach method to recreate this (and all the mocks) before each test, so there's no chance of leaking between them

service/src/main/java/bio/terra/cbas/config/AsyncConfiguration.java
salonishah11: I removed the `@Autowired` and `@ConstructorBinding` annotation on the constructor and it gave a run time error
```
Exception handler for async method 'public void bio.terra.cbas.service.RunSetsService.triggerWorkflowSubmission(...)' threw unexpected exception itself
java.lang.NullPointerException: Cannot invoke "bio.terra.cbas.dao.RunDao.getRuns(bio.terra.cbas.dao.RunDao$RunsFilters)" because "this.runDao" is null
```
Upon addition of `@Component` it threw the same error as mentioned in first thread `...defined as a regular bean which caused dependency injection to fail`

service/src/main/java/bio/terra/cbas/controllers/util/AsyncExceptionHandler.java
salonishah11: Assuming the params are in order, I updated the code [this way](https://github.com/DataBiosphere/cbas/pull/281/files#diff-a44e5a98fe9164633883e906505b2cb3c7db64d37861c83d9eaeff45162f71a2R53-R83) and SonarCloud seems to be happy.

service/src/main/java/bio/terra/cbas/service/RunSetsService.java
```diff
@@ -0,0 +1,388 @@
...
+import java.util.stream.Collectors;
+import org.broadinstitute.dsde.workbench.client.sam.model.UserStatusInfo;
+import org.databiosphere.workspacedata.model.RecordResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.annotation.Async;
+import org.springframework.stereotype.Component;
+
+@Component
+public class RunSetsService {
```
mspector, line 56: this is the nit-iest pick I've ever picked, but I'm noticing that we sometimes name our files using the singular form `RunSet` and sometimes the plural `RunSets`. My vote is singular `RunSet`!

salonishah11: Yeah I have seen that too but I tried to name the service this based on the controller and its called `RunSetsApiController`. I agree that we should just pick one, but I think that refactoring can be done in a later PR. If there is a strong opinion for singular I can rename `RunSetsService` to `RunSetService` but leave other name refactoring for another time.

mspector: I'm honestly kind of embarrassed to point it out. It can definitely come later

salonishah11: No its a completely valid point and I agree with the idea of being consistent in the naming

cjllanwarne (approved):

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
```diff
@@ -87,6 +87,7 @@ public UpdateResult<Run> updateRuns(
               Run::runId,
               r ->
                   r.status().nonTerminal()
+                      && r.engineId() != null
```
cjllanwarne, line 90: This is fine for now. I wonder if longer term we want something like `r.status().isSubmittedToEngine()`. But fine for now 👍

cjllanwarne: Thinking a little more... because it might be nice to precompute engine IDs and store them in the DB, even before we submit the workflow to the engine. 

That would let us be idempotent with workflow submissions across CBAS restarts, for example.

cjllanwarne: TL;DR: if it's possible to do this with something like a `r.status().isSubmittedToEngine()` then I think that would be better. But feel free to leave this as is if that turns out to be difficult for some reason

cjllanwarne (approved):

service/src/main/java/bio/terra/cbas/runsets/monitoring/SmartRunsPoller.java
salonishah11: Talked offline and this would be included in a follow-up ticket

PR #282, opened 2024-05-14
[ID-1273] Log workflow submissions to Bard

samanehsan:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1273

Purpose:
Identiteam is putting together a metrics dashboard and needs a way to determine when users submit workflows to Azure in Terra.

Changes:
In the `postRunSet` endpoint, call Bard to log the workflow submission event


cjllanwarne (commented):
I'm surprised you didn't need to update the `TestRunSetsApiController` tests to mock the BardService component. Maybe it "just works" because the return type is void and is never checked, so a null from the empty mock doesn't matter 🤔 

It might be nice to do mock it properly anyway, to validate that the right Bard method is called at the end of a successful POST request.

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
```diff
@@ -0,0 +1,63 @@
...
+  private DefaultApi getDefaultApi(BearerToken userToken) {
+    ApiClient client = bardClient.bardAuthClient(userToken);
+    return bardClient.defaultApi(client);
+  }
+
+  public void logRunSetEvent(RunSetRequest request, BearerToken userToken) {
+    String eventName = "workflow-submission";
+    HashMap<String, String> properties = new HashMap<>();
+    properties.put("runSetName", request.getRunSetName());
+    properties.put("methodVersionId", request.getMethodVersionId().toString());
```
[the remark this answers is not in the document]
samanehsan: I was looking at the `RunSetRequest` fields and and I don't see ones for the method name and version. Is there somewhere else I can get that info?

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  @Override
+  public Result checkHealth() {
+    try {
+      ApiClient client = bardClient.apiClient();
+      DefaultApi defaultApi = bardClient.defaultApi(client);
+      EventsEvent200Response status = defaultApi.systemStatus();
+      return new Result(true, "Ok");
```
samanehsan, line 78: If the API call is successful, the response is 200 and `status` is `null` (we should fix this in Bard). If the status is not ok, Bard throws an exception.

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
```diff
@@ -0,0 +1,82 @@
...
+      List<String> workflowIds,
+      BearerToken userToken) {
+    String eventName = "workflow-submission";
+    HashMap<String, String> properties = new HashMap<>();
+    properties.put("runSetName", request.getRunSetName());
+    properties.put("methodName", methodVersion.method().name());
+    properties.put("methodSource", methodVersion.method().methodSource());
+    properties.put("methodVersionName", methodVersion.name());
+    properties.put("methodVersionUrl", methodVersion.url());
+    properties.put("recordCount", String.valueOf(request.getWdsRecords().getRecordIds().size()));
```
samanehsan, line 50: I'm inclined to remove this now that we have the Cromwell workflow ids. Is there a reason to log how many records a user wanted to analyze?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -475,6 +496,7 @@ private RunStateResponse storeRun(
 
     return new RunStateResponse()
         .runId(runId)
+        .engineId(externalId)
```
samanehsan, line 499: This is the Cromwell workflow id.

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
Ghost-in-a-Jar: couldnt hurt to leave there imo

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
```diff
@@ -0,0 +1,82 @@
...
+      properties.put("githubIsPrivate", githubMethodDetails.isPrivate().toString());
+    }
+
+    logEvent(eventName, properties, userToken);
+  }
+
+  public void logEvent(String eventName, Map<String, String> properties, BearerToken userToken) {
+    EventsEventLogRequest request = new EventsEventLogRequest().properties(properties);
+    try {
+      getDefaultApi(userToken).eventsEventLog(eventName, appId, request);
```
Ghost-in-a-Jar, line 65: this is the new endpoint we added right

Ghost-in-a-Jar (approved):

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
samanehsan: Yeah this is the new one

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -323,8 +332,20 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+      List<String> workflowIds =
+          runStateResponseList.stream()
+              .map(RunStateResponse::getEngineId)
+              .filter(Objects::nonNull)
+              .toList();
+      GithubMethodDetails githubMethodDetails = null;
+      if (methodSourceEnum.equals(PostMethodRequest.MethodSourceEnum.GITHUB)) {
+        githubMethodDetails =
+            githubMethodDetailsDao.getMethodSourceDetails(methodVersion.getMethodId());
+      }
```
cjllanwarne, line 345: This mechanism is changing in https://github.com/DataBiosphere/cbas/pull/279 to always just being available already in the `Method`, so no extra "fetch" hops will be required.

Just something to be aware of depending on which order these merge in 😄

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -323,8 +332,20 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
...
+      bardService.logRunSetEvent(
+          request, methodVersion, githubMethodDetails, workflowIds, userToken);
```
cjllanwarne, line 347: FYI this ties bard in to only work in "raw github import" workflows. Is that intended? Or would metrics for a dockstore workflow also be important?

cjllanwarne: As a side note: the changes in #279 would presumably make this interface easier too, since you won't need an extra parameter just for the githubMethodDetails

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
cjllanwarne: In the future we might no longer have a 1:1 mapping between WDS rows and Cromwell workflows, eg:

- Zip together entries from 2 different 10-column tables to produce 10 workflows from 20 WDS records
- Cross product entries from 2 different 10-column tables to produce 100 workflows from 20 WDS records
- Allow non-WDS backed workflows to run, producing 1 (or many) workflows from 0 records

... so I think the more we record, and the more generic the naming can be, the better

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
```diff
@@ -31,20 +34,36 @@ private DefaultApi getDefaultApi(BearerToken userToken) {
...
+    properties.put("methodSource", methodVersion.method().methodSource());
+    properties.put("methodVersionName", methodVersion.name());
+    properties.put("methodVersionUrl", methodVersion.url());
+    properties.put("recordCount", String.valueOf(request.getWdsRecords().getRecordIds().size()));
+    properties.put("workflowIds", workflowIds.toString());
+
+    if (Objects.nonNull(githubMethodDetails)) {
+      properties.put("githubOrganization", githubMethodDetails.organization());
+      properties.put("githubRepository", githubMethodDetails.repository());
+      properties.put("githubIsPrivate", githubMethodDetails.isPrivate().toString());
```
cjllanwarne, line 56: In case either of these sound useful... the method table has a `branchOrTag` field representing the branch name (or tag name!) that the method is coming from. There is also a method version equivalent of this github details with a commit hash field.

common/openapi.yml
```diff
@@ -759,6 +759,9 @@ components:
             string
           format: uuid
           example: 00000000-0000-0000-0000-000000000000
+        engine_id:
+          type:
+            string
```
cjllanwarne, line 764: I'm not sure I followed why this is necessary here?

If it's just to make the field available for internal logic then I would have thought we should be able to pass the value around without changing the interface?

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
```diff
@@ -323,8 +332,20 @@ public ResponseEntity<RunSetStateResponse> postRunSet(RunSetRequest request) {
         new RunSetStateResponse().runSetId(runSetId).runs(runStateResponseList).state(runSetState);
 
     captureResponseMetrics(response);
-    bardService.logRunSetEvent(request, userToken);
-
+    if (bardServerConfiguration.enabled()) {
+      List<String> workflowIds =
+          runStateResponseList.stream()
```
cjllanwarne, line 337: This strikes me as odd - that we have to update the API types just to get a value internally in the system so we can log it.

I believe this becomes a little more sane in https://github.com/DataBiosphere/cbas/pull/281 which splits the Run-creation away from the submitting. So you might have the information you need more easily available after that.

common/openapi.yml
```diff
@@ -759,6 +759,9 @@ components:
             string
           format: uuid
           example: 00000000-0000-0000-0000-000000000000
+        engine_id:
```
cjllanwarne, line 762: Not that I hate this here, but it feels odd that we are relying on external API types for internal data passing.

I think I'd rather change the return types of the internal methods rather than change the external API, if we can?

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
```diff
@@ -0,0 +1,82 @@
...
+  public BardService(BardClient bardClient) {
+    this.bardClient = bardClient;
+  }
+
+  private DefaultApi getDefaultApi(BearerToken userToken) {
+    ApiClient client = bardClient.bardAuthClient(userToken);
+    return bardClient.defaultApi(client);
+  }
+
+  public void logRunSetEvent(
```
tlangs, line 37: I worry about this by synchronous. If Bard, for some reason, is slow in responding, it'll slow down CBAS. [In DRSHub](https://github.com/DataBiosphere/terra-drs-hub/blob/068e2389d3bbd9b1bc888bb574c5ca096d82b5f6/service/src/main/java/bio/terra/drshub/services/TrackingService.java#L34), logging to Bard is async, so that it happens in a different threadpool. We should probably do the same thing here.

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
samanehsan: CBAS will log to Bard in both types of workflows, but if the source is github it will include additional details like the org and repo. I'll add in the dockstore info too!

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
samanehsan: Oh I'll take a look!

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
samanehsan: Saloni has a PR to add an async work queue so we could wait for that to go in first and use that

tlangs: I think that would be a a good idea

service/src/main/java/bio/terra/cbas/controllers/RunSetsApiController.java
samanehsan: I merged in those changes and I still have the same issue. Since `buildInputsAndSubmitRun` returns `List<RunStateResponse>`, I thought it made sense to include the engine id in `RunStateResponse`, but an alternative could be to make it return a new object type with `runStateResponse`` and `workflowIds` fields. Is there a reason why you don't want it to be in the run state response?

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
cjllanwarne: Saloni's PR is now merged. Do you feel like you're willing to have a go at making this async with the same mechanism? We might want a separate queue for this, but it should be easy enough to make that work

service/src/main/java/bio/terra/cbas/service/RunSetsService.java
```diff
@@ -65,6 +68,8 @@ public class RunSetsService {
   private final UuidSource uuidSource;
   private final ObjectMapper objectMapper;
   private final CbasContextConfiguration cbasContextConfiguration;
+  private final BardService bardService;
+  private final BardServerConfiguration bardServerConfiguration;
```
cjllanwarne, line 72: It feels like a little bit of a code smell that the configuration is being brought in and analyzed here instead of in the `BardService`. I see the usage is for an "is it enabled" check, could we push that check down into the BardService itself to simplify the interface here - and not have to remember to do it if we use the bard service somewhere else, too?

service/src/main/java/bio/terra/cbas/service/RunSetsService.java
```diff
@@ -196,7 +206,7 @@ public void triggerWorkflowSubmission(
     }
 
     // For each Record ID, build workflow inputs and submit the workflow to Cromwell
-    List<RunStateResponse> runStateResponseList =
+    ImmutablePair<List<RunStateResponse>, List<String>> runStateResponse =
```
cjllanwarne, line 209: Nit: This return value is getting towards a "wrap it in a `record` type" level of complexity

service/src/main/java/bio/terra/cbas/service/RunSetsService.java
```diff
@@ -348,7 +370,7 @@ private List<RunStateResponse> buildInputsAndSubmitRun(
               .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 
       if (engineIdToWorkflowInput.isEmpty()) {
-        return runStateResponseList;
+        return new ImmutablePair<>(runStateResponseList, List.of());
```
cjllanwarne, line 373: Nit: Is this "build the return type" logic is getting complex enough to warrant a separate function and a small set of unit tests?

service/src/test/java/bio/terra/cbas/dependencies/bard/TestBardService.java
```diff
@@ -0,0 +1,173 @@
...
+
+  @BeforeEach
+  void setup() {
+    BardClient bardClient = mock(BardClient.class);
+    bardService = new BardService(bardClient);
+    ApiClient apiClient = mock(ApiClient.class);
+    defaultApi = mock(DefaultApi.class);
+    userToken = new BearerToken("foo");
+    lenient().when(bardClient.apiClient()).thenReturn(apiClient);
+    lenient().when(bardClient.bardAuthClient(any())).thenReturn(apiClient);
```
cjllanwarne, line 52: Might be nice to have two separate apiClients here so that you can verify that the right methods are being called on the right clients (eg `status` to the non-auth-client and `events` to the auth'd client)

service/src/test/java/bio/terra/cbas/service/TestRunSetsService.java
```diff
@@ -243,6 +247,9 @@ void init() {
     uuidSource = mock(UuidSource.class);
     objectMapper = mock(ObjectMapper.class);
     cbasContextConfiguration = mock(CbasContextConfiguration.class);
+    bardService = mock(BardService.class);
+    bardServerConfiguration = mock(BardServerConfiguration.class);
+    when(bardServerConfiguration.enabled()).thenReturn(true);
```
cjllanwarne, line 252: If we pushed this check down into the bard service itself, we could have a service-level unit test to verify that it calls the API when enabled and doesn't when it's disabled

service/src/main/java/bio/terra/cbas/service/RunSetsService.java
samanehsan: That's a good point! I originally had it here so that I would not have to fetch the github method details unnecessarily but that is no longer an issue

samanehsan:
> Saloni's PR is now merged. Do you feel like you're willing to have a go at making this async with the same mechanism? We might want a separate queue for this, but it should be easy enough to make that work

@cjllanwarne @salonishah11 Yeah! Since `bardService.logRunSetEvent` is being called from within the async `triggerWorkflowSubmission`, is it okay to make that bard method async as well?

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
```diff
@@ -1,84 +1,22 @@
...
 import java.util.Map;
-import java.util.Optional;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Component;
 
 @Component
-public class BardService implements HealthCheck {
+public interface BardService {
```
cjllanwarne, line 12: Should this interface still implement `HealthCheck`?

service/src/main/java/bio/terra/cbas/dependencies/bard/BardServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+      properties.put("githubRepository", githubMethodDetails.repository());
+      properties.put("githubIsPrivate", githubMethodDetails.isPrivate().toString());
+    }
+
+    logEvent(eventName, properties, userToken);
+  }
+
+  @Async("bardAsyncExecutor")
+  @Override
+  public void logEvent(String eventName, Map<String, String> properties, BearerToken userToken) {
```
cjllanwarne, line 69: Note that in Spring, the `Async` annotation only applies to proxied calls. My mental model is something like:

- Class: "I want an autowired BardService"
- Spring: "Ah yes, I can make one of these BardServiceImpls. Oh, and it has async methods so I'm going to create a proxied version which turns this method into one that returns a Future"
- Class: (receives a new, proxying version of the BardServiceImpl)

So in this case, the `logEvent` will only be async if it's called from an autowired reference to BardService. And because this one is coming from _inside_ the same object, it won't be made asynchronously here.

cjllanwarne: Incidentally, I think this is also why the test were having errors referencing "that's a proxy not a BardService instance", and making the tests reference the interface makes the mocking work

service/src/main/java/bio/terra/cbas/dependencies/bard/BardServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  @Override
+  public Result checkHealth() {
+    try {
+      ApiClient client = bardClient.apiClient();
+      DefaultApi defaultApi = bardClient.defaultApi(client);
+      defaultApi.systemStatus();
+      return new Result(true, "Ok");
```
cjllanwarne, line 86: Is there anything in the `systemStatus()` response that could indicate a bad state? Or a message we could forward along instead of `OK`?

service/src/main/java/bio/terra/cbas/dependencies/common/ScheduledHealthChecker.java
```diff
@@ -34,7 +34,7 @@ public ScheduledHealthChecker(
       LeonardoService leonardoService,
       SamService samService,
       EcmService ecmService,
-      BardService bardService,
+      @Qualifier("bardServiceImpl") HealthCheck bardService,
```
cjllanwarne, line 37: If `BardService` implemented HealthCheck too, I think you could avoid doing it like this

service/src/main/resources/application.yml
```diff
@@ -63,7 +63,10 @@ cbas:
       coreThreadPoolSize: 5
       maxThreadPoolSize: 10
       queueCapacity: 500
-
+    bard:
+      coreThreadPoolSize: 5
+      maxThreadPoolSize: 10
+      queueCapacity: 500
```
cjllanwarne, line 69: 500 seems fine for run-set recording. If we ever record workflow-by-workflow we'll want this to be much higher.

Do we do anything special if we overflow this capacity?

service/src/main/java/bio/terra/cbas/dependencies/bard/BardServiceImpl.java
samanehsan: Right now it looks like Bard will only return an error code if it's status is unhealthy, so there should not be a case where it returns a 200 from the status endpoint with a message that it's in a bad state (but I can confirm with the team)!

service/src/main/java/bio/terra/cbas/dependencies/common/ScheduledHealthChecker.java
samanehsan: Oh interesting...I didn't think about making the `BardService` interface implement the `HealthCheck` interface

service/src/main/resources/application.yml
samanehsan: Not at the moment...how is that handled for async workflow submissions? 🤔

service/src/main/java/bio/terra/cbas/dependencies/bard/BardService.java
tlangs: I'm not convinced CBAS should care about Bard's health. CBAS should really be fire-and-forgetting to Bard, since it doesn't actually rely on Bard for anything.

samanehsan: I'm inclined to take this out if that's okay with you @cjllanwarne. It's interesting TDR doesn't include Bard in its [status check](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/status/StatusService.java#L45). I wonder what other services do!

service/src/main/java/bio/terra/cbas/dependencies/bard/BardServiceImpl.java
samanehsan: In that case could I get rid of the `BardService` interface?

samanehsan: I tried getting rid of the interface and got the proxy error again btw

service/src/main/resources/application.yml
mspector: I don't think we do any special handling when the run set queue overflows. This [stackoverflow answer](https://stackoverflow.com/a/44647863) says that items added to a maxed out queue are rejected (which seems like a good fact to keep in mind, but may not imply the same solution for run set, workflow, or bard logging)

I think solving for this when it comes up again is fine

mspector (approved):

PR #283, opened 2024-05-15
WM-2638: Fix role on new method version details table

cjllanwarne:


aednichols (approved):
Nice fix 👍 

Looks like owner updates are a footgun every time we need to add a new table... any way to automate/abstract it? (Not in this PR.)

kpierre13 (approved):

PR #284, opened 2024-05-16
WM-2640: Validate table ownership role is correctly updated

cjllanwarne:
Example test failure message for missing the role reassignment: `Table github_method_version_details is owned by 'test_user' but should be owned by 'test_role'. Do you need to update the 20231102_set_table_role.yaml changeset?`


THWiseman (approved):

aednichols (approved):

salonishah11 (approved):
💯

PR #285, opened 2024-05-22
[WM-2335] delete method

mspector:
this is a proof-of-concept for the simplest possible approach to method cleanup


service/src/main/java/bio/terra/cbas/dao/MethodVersionDao.java
```diff
@@ -134,8 +134,16 @@ public int updateOriginalWorkspaceId(UUID methodVersionId, UUID originalWorkspac
   }
 
   public int deleteMethodVersion(UUID methodVersionId) {
+    MapSqlParameterSource methodVersionIdParamSource =
+        new MapSqlParameterSource("methodVersionId", methodVersionId);
+
+    String deleteMethodVersionSql = "DELETE FROM %s WHERE method_version_id = :methodVersionId";
```
cjllanwarne, line 140: Not a super strong opinion but this does feel a little funky to me and it took a while to work out what the intention was... 

I wonder how offended you'd be if I suggested just spelling out the two (almost) identical sql strings here - 
```java
String deleteGithubMethodVersionDetailsSql = "DELETE FROM github_method_version_details WHERE method_version_id = :methodVersionId";
String deleteMethodVersionSql = "DELETE FROM method_version WHERE method_version_id = :methodVersionId";
```

service/src/main/java/bio/terra/cbas/service/MethodService.java
```diff
@@ -0,0 +1,62 @@
...
+      RunDao runDao,
+      GithubMethodDetailsDao githubMethodDetailsDao) {
+    this.methodDao = methodDao;
+    this.methodVersionDao = methodVersionDao;
+    this.runSetDao = runSetDao;
+    this.runDao = runDao;
+    this.githubMethodDetailsDao = githubMethodDetailsDao;
+  }
+
+  public void deleteMethod(UUID methodId) {
```
cjllanwarne, line 36: This operation would probably benefit from a transactional annotation

service/src/main/java/bio/terra/cbas/service/MethodService.java
```diff
@@ -0,0 +1,62 @@
...
+    Method methodToDelete = methodDao.getMethod(methodId);
+    methodDao.unsetLastRunSetId(methodId);
+
+    List<MethodVersion> methodVersions =
+        methodVersionDao.getMethodVersionsForMethod(methodToDelete);
+    methodVersions.forEach(
+        methodVersion -> methodVersionDao.unsetLastRunSetId(methodVersion.methodVersionId()));
+
+    // TODO: put this in RunSetService
```
cjllanwarne, line 45: 👍

service/src/main/java/bio/terra/cbas/service/MethodService.java
```diff
@@ -0,0 +1,62 @@
...
+    runSetDao
+        .getRunSetsWithMethodId(methodId)
+        .forEach(
+            runSet -> {
+              List<Run> runSetRuns =
+                  runDao.getRuns(new RunDao.RunsFilters(runSet.runSetId(), null));
```
cjllanwarne, line 51: ... and this in a `RunsService`, perhaps?

cjllanwarne: Neither of these is a blocker to merging the PR though, IMO

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
```diff
@@ -134,4 +135,22 @@ void unsetLastRunSetId() {
     Method updatedMethod = methodDao.getMethod(retrievedMethod.methodId());
     assertNull(updatedMethod.lastRunSetId());
   }
+
+  @Test
+  void deleteMethod() {
+    List<Method> allMethods = methodDao.getMethods();
+    assertEquals(2, allMethods.size());
```
cjllanwarne, line 142: It's probably better to assume that each test starts with an empty database and sets up its own state, instead of relying on either previous tests (because order isn't guaranteed) or some initializer (because that locks tests to using the same initial state).

Otherwise when I look at this test case my main thought is... wait, why should I expect there to be 2 methods already in there?

If you're worried about code copying, you could always just make the "initialize state" method called explicitly instead of implicitly

service/src/test/java/bio/terra/cbas/dao/TestMethodDao.java
```diff
@@ -134,4 +135,22 @@ void unsetLastRunSetId() {
...
+    assertEquals(1, remainingMethods.size());
+
+    methodDao.deleteMethod(method2.methodId());
+    List<Method> remainingMethods2 = methodDao.getMethods();
+    assertEquals(0, remainingMethods2.size());
+
+    // TODO: assert a "method doesn't exist" exception
+    //    Method deletedMethod = methodDao.getMethod(method1.methodId());
+    //    assert deletedMethod.isDeleted();
+  }
```
cjllanwarne, line 155: Sounds good!

cjllanwarne (commented):
Would also like to see a containerized test of the composition of the different calls in the `MethodsService`, if possible

PR #289, opened 2024-06-06
WM-2646: Use WDS query to fetch rows in batches

cjllanwarne:


service/src/main/java/bio/terra/cbas/dependencies/wds/WdsService.java
```diff
@@ -57,6 +78,82 @@ public RecordResponse updateRecord(
...
+  }
+
+  private WdsRecordResponseDetails getRecordsBatch(
+      String recordType, List<String> recordIds, BearerToken userToken) {
+
+    try {
+      RecordQueryResponse queryResponse =
+          queryRecords(
+              new SearchRequest()
+                  .filter(new SearchFilter().ids(recordIds))
```
salonishah11, line 108: When this new filter was released, AJ team had mentioned that it is possible some of WDS apps in prod might not have been updated to have the new search filter for querying records as all the apps might not be updated. But it has been a while since this new version was released. Do you think it would be worth checking with AJ team to see if all existing WDS apps might have the new search filter?

salonishah11: They had also mentioned that there is a `capabilities` API that can be used to figure out which features are available in a particular WDS version.

mspector: +1... if we can determine that the WDS apps have all been updated in prod, this LGTM, otherwise I wonder if we should build in a call to this `capabilities` API to decide which call to make

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsService.java
```diff
@@ -57,6 +83,131 @@ public RecordResponse updateRecord(
...
+        listenerResetRetryTemplate, () -> wdsClient.capabilitiesApi(userToken).capabilities());
+  }
+
+  public boolean useSearchByIdsFilter(BearerToken userToken) {
+    try {
+      Capabilities wdsAppInstanceCapabilities = getCapabilities(userToken);
+      Map<String, Object> additionalProperties =
+          wdsAppInstanceCapabilities.getAdditionalProperties();
+
+      return additionalProperties != null && additionalProperties.containsKey("search.filter.ids");
```
davidangb, line 97: Ideally you'd also check here that the value of `search.filter.ids` is true, instead of just checking that the key exists.

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsService.java
```diff
@@ -57,6 +83,131 @@ public RecordResponse updateRecord(
...
+      return false;
+    }
+  }
+
+  public WdsRecordResponseDetails getRecords(
+      String recordType, List<String> recordIds, BearerToken userToken) {
+    if (useSearchByIdsFilter(userToken))
+      return getRecordsUsingSearchFilter(recordType, recordIds, userToken);
+    else return getRecordsSerially(recordType, recordIds, userToken);
+  }
```
davidangb, line 110: the capabilities api is a very lightweight implementation inside WDS, so it's no big deal … but conceivably you could check and cache the capabilities response for some amount of time. WDS's response will only change infrequently: when we perform app updates, and then only if something significant has changed in the app. I wouldn't block this PR on any kind of caching though.

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsClient.java
```diff
@@ -51,4 +52,8 @@ protected ApiClient getApiClient(BearerToken userToken) throws DependencyNotAvai
   RecordsApi recordsApi(BearerToken userToken) throws DependencyNotAvailableException {
     return new RecordsApi(getApiClient(userToken));
   }
+
+  CapabilitiesApi capabilitiesApi(BearerToken userToken) {
+    return new CapabilitiesApi(getApiClient(userToken));
```
mspector, line 57: 🎉

mspector (approved):
I would want to see David's comment addressed about `search.filter.ids`, but I don't have any other comments - LGTM

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsService.java
salonishah11: Yes we had discussed this briefly and I think we would want to cache this if its start affecting the performance or as part of future slice whichever happens first. But it is good to know that is a very lightweight implementation.

service/src/main/java/bio/terra/cbas/dependencies/wds/WdsService.java
salonishah11: Added this check in [this commit](https://github.com/DataBiosphere/cbas/pull/289/commits/09962fe41d91cfbcd1bc5a66467f1acb251d6ab6)

kpierre13 (approved):

PR #292, opened 2024-06-18
[WM-2670] Runs submitted to Cromwell are marked in Initializing status in CBAS

salonishah11:
Jira ticket: https://broadworkbench.atlassian.net/browse/WM-2670

Current status transitions that was causing the workflow to revert back to Queued state mid-run:
- When a workflow is being processed in CBAS upon submission, it is marked in `Queued` state
- When CBAS submits that workflow to Cromwell, it is marked in `Initializing` state
- Now when Cromwell receives a workflow it stores the workflow in DB in `Submitted` state 
- When CBAS polls for status update it would mark the same run back to `Queued` state because of the mapping in `fromCromwellStatus()` and as a result the status transitions were Queued -> Initializing -> Queued -> Running -> other states

Proposed changes:
When Cromwell returns a workflow in `Submitted` state, the corresponding CBAS status is now `Initializing` and hence the status transitions would be Queued -> Initializing -> Running -> other states

### DIFF
--- a/service/src/main/java/bio/terra/cbas/models/CbasRunStatus.java
+++ b/service/src/main/java/bio/terra/cbas/models/CbasRunStatus.java
@@ -75,7 +75,7 @@ public enum CbasRunStatus {
 
   public static CbasRunStatus fromCromwellStatus(String status) {
     return switch (status) {
-      case "Submitted" -> CbasRunStatus.QUEUED;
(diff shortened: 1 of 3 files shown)


mspector (approved):
nice. 1 line of code change, 50 lines of test

kpierre13 (approved):

PR #296, opened 2024-06-27
[WM-2613] construct fixed url during get methods call

mspector:


service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -301,8 +303,16 @@ public ResponseEntity<MethodListResponse> getMethods(
...
-          List.of(methodVersionToMethodDetails(methodVersionDao.getMethodVersion(methodVersionId)));
+      try {
+        methodDetails =
+            List.of(
+                methodVersionToMethodDetails(methodVersionDao.getMethodVersion(methodVersionId)));
+      } catch (bio.terra.cbas.common.exceptions.MethodProcessingException
+          | java.net.MalformedURLException
+          | java.net.URISyntaxException
+          | bio.terra.dockstore.client.ApiException e) {
+        return ResponseEntity.badRequest().build();
```
mspector, line 314: TODO: fix this

mspector:
reviewers, I'm not sure why these 7 sonarcloud issues are popping up only now... that rule seems inconsistently applied. I feel ok ignoring them

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -301,8 +304,16 @@ public ResponseEntity<MethodListResponse> getMethods(
...
-          List.of(methodVersionToMethodDetails(methodVersionDao.getMethodVersion(methodVersionId)));
+      try {
+        methodDetails =
+            List.of(
+                methodVersionToMethodDetails(methodVersionDao.getMethodVersion(methodVersionId)));
+      } catch (bio.terra.cbas.common.exceptions.MethodProcessingException
+          | java.net.MalformedURLException
+          | java.net.URISyntaxException
+          | bio.terra.dockstore.client.ApiException e) {
+        return ResponseEntity.badRequest().body(new MethodListResponse().error(e.toString()));
```
mspector, line 315: reviewers, is this the right thing to do when the method version URL can't be resolved?

kpierre13 (approved):

kpierre13:
> reviewers, I'm not sure why these 7 sonarcloud issues are popping up only now... that rule seems inconsistently applied. I feel ok ignoring them

I agree this seems odd now? Unless this has always been an issue and ignored. I'm indifferent about it personally. I'm good with whatever you feel comfortable with.

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
salonishah11: Should this be Internal Server Error? If it was a user issue / bad request then the method wouldn't have been imported only no? 🤔

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
```diff
@@ -301,8 +304,16 @@ public ResponseEntity<MethodListResponse> getMethods(
...
+      }
     } else {
       List<Method> methods =
           methodId == null ? methodDao.getMethods() : List.of(methodDao.getMethod(methodId));
```
salonishah11, line 319: Should the similar construct hash-ed method url also be added here in this else case?

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
mspector: you're right, bad request seemed wrong - internal server error sounds good to me

service/src/main/java/bio/terra/cbas/controllers/MethodsApiController.java
mspector: it should be getting the constructed url by way of `methodToMethodDetails` [on line 323](https://github.com/DataBiosphere/cbas/pull/296/files/ebc22f5e2d03d6a1dda034e1ce514659975ca3e2#diff-905f97fe9aa55195961e43b0f23f4b48e35fe77006e1536aaecf034fcdf361fcR323)

mspector:
@salonishah11 , it's not what you asked for exactly, but your comment about internal server errors made me realize that I was handling errors in this code path in the wrong way. I've simplified it a lot - I think you'll agree it's better. I can walk you through it later today if it helps

salonishah11 (approved):

PR #297, opened 2024-07-09
[WM-2683] Adds /capabilities API

salonishah11:
Jira ticket: https://broadworkbench.atlassian.net/browse/WM-2683

This PR adds new `capabilities` API. The response is a JSON object that is defined in `capabilities.json` file. The response structure is [similar to WDS' capabilities API](https://github.com/DataBiosphere/terra-workspace-data-service/wiki/Capabilities-API#response-structure) where response is:
- flat list of key-value pairs where keys are strings and values should be primitive type
- keys use dot notation to group related keys into a namespace, e.g. `submission.limits`

Wiki documentation: https://github.com/DataBiosphere/cbas/wiki/Capabilities-API

Response:
```
{
  "submission.limits.maxOutputs": 300,
  "submission.limits.maxWorkflows": 100,
  "submission.limits.maxInputs": 200
}
```

TODO:

- [x] add documentation in Wiki page

### DIFF
--- a/service/src/main/resources/capabilities.json
+++ b/service/src/main/resources/capabilities.json
@@ -2,5 +2,6 @@
   "submission.limits.maxWorkflows" : 100,
   "submission.limits.maxInputs" : 200,
   "submission.limits.maxOutputs" : 300,
-  "workflow.archiving" : true
+  "workflow.archiving" : true,
+  "workflow.private_workflows.github" : true
 }
--- a/service/src/main/resources/capabilities.json
+++ b/service/src/main/resources/capabilities.json
@@ -1,0 +1,6 @@
+{
+  "submission.limits.maxWorkflows" : 100,
+  "submission.limits.maxInputs" : 200,
+  "submission.limits.maxOutputs" : 300,
+  "workflow.archiving" : true
+}
--- a/service/src/main/resources/capabilities.json
+++ b/service/src/main/resources/capabilities.json
@@ -1,7 +1,5 @@
 {
   "submission.limits.maxWorkflows" : 100,
   "submission.limits.maxInputs" : 200,
-  "submission.limits.maxOutputs" : 300,
-  "workflow.archiving" : true,
-  "workflow.private-workflows.github" : true
+  "submission.limits.maxOutputs" : 300
 }
--- a/service/src/main/resources/capabilities.json
+++ b/service/src/main/resources/capabilities.json
@@ -3,5 +3,5 @@
   "submission.limits.maxInputs" : 200,
   "submission.limits.maxOutputs" : 300,
   "workflow.archiving" : true,
-  "workflow.private_workflows.github" : true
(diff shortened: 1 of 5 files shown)


service/src/main/resources/capabilities.json
```diff
@@ -0,0 +1,7 @@
+{
+  "submission.limits.maxWorkflows" : 100,
+  "submission.limits.maxInputs" : 200,
+  "submission.limits.maxOutputs" : 300,
+  "workflow.archiving" : true,
+  "workflow.private-workflows.github" : true
```
salonishah11, line 6: Question for reviewers - should these 2 capabilities be included here considering they have already been released few months back?

kpierre13: Hm, that's a good question. Since these are more like features and not configurations, once this is merged in to apps then so too should those features. Any app that may not have these features also won't have this API, so then I think it would be safe enough to omit these two capabilities.

kpierre13 (approved):

service/src/main/resources/capabilities.json
salonishah11: Yeah that's what I was thinking too - that since those 2 features have already been merged it doesn't make sense to include in the response currently. I will remove these 2 capabilities from response. In future when new features are added to CBAS maybe then they can be added to capabilities response.

service/src/test/java/bio/terra/cbas/controllers/TestPublicApiController.java
```diff
@@ -0,0 +1,56 @@
...
+    MvcResult mvcResult =
+        mockMvc.perform(get(CAPABILITIES_API)).andExpect(status().isOk()).andReturn();
+
+    CapabilitiesResponse parsedResponse =
+        objectMapper.readValue(
+            mvcResult.getResponse().getContentAsString(), CapabilitiesResponse.class);
+
+    assertThat(parsedResponse).isNotNull();
+
+    assertEquals(100, parsedResponse.get("submission.limits.maxWorkflows"));
```
cjllanwarne, line 45: Something to consider - do we want to be using exact values from the actual `capabilities.json` in these tests? Could we have something maybe split in two halves, like 

- Test that the real file is readable by the service
- Test that we can interpret a mock file correctly and get the right values from it

salonishah11: That is a good point. I do have a test case for this but I see your point about the downsides for directly testing the actual json file for exact values. I will make that change.
> Test that the real file is readable by the service

salonishah11: I actually simply updated this test case to check that the response returned from API is parseable into CapabilitiesResponse model and verify it is not empty. I think this should cover most of the cases. If there is a need to have a test case read from mock file then we can add that later.

